PR #2, opened 2021-09-15
Adds custom chakra theme and storybook

georgeweiler:


src/App.tsx
```diff
@@ -1,13 +1,98 @@
...
     <ChakraProvider theme={theme}>
-      <div>
-        <Heading>Threshold Token Dashboard</Heading>
-      </div>
+      <Container>
+        <Heading>Threshold Token Dashboard?</Heading>
+        <VStack spacing={12} mt={6}>
+          <VStack>
+            <HStack spacing={4}>
+              <Badge size="sm">Badge</Badge>
```
r-czajkowski, line 20: Looks like the `size` prop doesn't work. I've played with this prop a bit- regardless of the prop value(`xs, sm, md, lg`), the component is always the same size. I can change the size in the storybook but it doesn't work on the dapp side.

From the [Threshold Design System](https://www.figma.com/file/zZi2fYDUjWEMPQJWAt8VWv/Threshold-DS?node-id=834%3A18160):
> The badge comes in 2 sizes: large and small. Use the larger option if you want to draw attention to this badge.`

So we should probably only support 2 sizes. Thoughts? 🤔

src/App.tsx
```diff
@@ -1,13 +1,98 @@
 import React, { FC } from "react"
```
r-czajkowski, line 1: Unnecessary `React` import.

.storybook/badge.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react"
+import { Badge as ChakraBadge, Container } from "@chakra-ui/react"
+import {
+  badgeVariants,
+  buttonVariants,
```
r-czajkowski, line 5: Unused `buttonVariants` import.

.storybook/badge.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react"
```
r-czajkowski, line 1: We use `React v17` so there is no need to import `React` to be in scope. Ofc, we can still import, but it is not required- I think it would be nice to have the same opinion about it. What do you think guys? @georgeweiler @michalsmiarowski

.storybook/button.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react"
```
r-czajkowski, line 1: Same as above.

src/theme/Button.ts
```diff
@@ -0,0 +1,57 @@
...
+    link: {
+      backgroundColor: "transparent",
+      _hover: {
+        backgroundColor: "transparent",
+      },
+      _active: {
+        backgroundColor: "transparent",
+      },
+    },
+    solid: (props: any) => {
```
r-czajkowski, line 18: According to the [Threshold Design System](https://www.figma.com/file/zZi2fYDUjWEMPQJWAt8VWv/Threshold-DS?node-id=817%3A9473)- we should have 3 different variants: `primary`, `secondary`, `tertiary`.

.storybook/button.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const Button = Template.bind({})
+
+// @ts-ignore
+Button.args = {
+  colorScheme: colorSchemes[0],
+  variant: buttonVariants[0],
+  size: sizes[1],
+  icon: "none",
+}
```
r-czajkowski, line 28: Can we add `isDisabled` prop? I think it would be useful.

r-czajkowski (changes requested):
Chakra looks so powerful, good job George! Left a few comments. Regarding unnecessary imports, I think we can provide a rule for eslint config to avoid unused imports in the feature.

src/theme/Button.ts
georgeweiler: Yeah Liz and I talked about this and for the sake of not having to create new variant names and apply all the appropriate styles I would suggest we stick to the chakra names: `solid`, `outline` & `ghost`. Liz put the chakra name conversions in the Design System, so it's just a naming change. wdyt?

.storybook/badge.stories.tsx
georgeweiler: It is such a habit for me and my IDE complains about having react in scope, so I would prefer to import it.  Not super strong opinion though

src/App.tsx
georgeweiler: I think it's okay keep the size options from Chakra, and just not use them instead of going through the effort to rebuild the component but strip away the extra size options.

georgeweiler: I will double check the size prop. thanks for doing QA! 👍

.storybook/badge.stories.tsx
georgeweiler: I did some research into this and it's easy to remove the required check for my IDE, so that's a non-issue. I've pushed up a change to our eslint config based on the react docs: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint

Only issue is that this is that eslint won't enforce that you *don't* import it. It just won't complain if you don't, and won't complain if you do.

so here are the options I can think of. I'm happy with any, although I do think the last one is a bit overkill.

- try to remember not to import React, but don't enforce via linting
- forget about it completely and either import, or not, but we don't care
- enforce importing React, even though it's not necessary
- write a custom eslint rule to strictly enforce

lmk what you think

michalsmiarowski (changes requested):
Left some suggestions, but overall great job!!

.storybook/badge.stories.tsx
michalsmiarowski: I'm on the "do not import React when it's not necessary" side. Who's with me? Anyone? 😅

src/App.tsx
michalsmiarowski: I like the Options 2 better. Let's just change the storybook option to `fontSize`

.storybook/badge.stories.tsx
georgeweiler: I'm happy to omit the React import. Do you guys think it's worth writing a rule to enforce that it's not imported?

.storybook/chakraVars.ts
```diff
@@ -0,0 +1,17 @@
...
+  "red",
+  "orange",
+  "yellow",
+  "green",
+  "teal",
+  "blue",
+  "cyan",
+  "purple",
+  "pink",
+]
```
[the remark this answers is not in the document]
georgeweiler: Sure that's a good idea

.storybook/button.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+      >
+        Button
+      </ChakraButton>
+    </Container>
+  )
+}
+
+export const Button = Template.bind({})
+
+// @ts-ignore
```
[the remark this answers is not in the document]
georgeweiler: I was able to fix by passing the args directly to the binding.

.storybook/button.stories.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react"
+import { Button as ChakraButton, Container } from "@chakra-ui/react"
+import { MdBuild } from "react-icons/md"
+import { buttonVariants, colorSchemes, sizes } from "./chakraVars"
+
+const Template = ({ icon, ...args }: any) => {
```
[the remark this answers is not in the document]
georgeweiler: this is a great suggestion.

.storybook/badge.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Badge as ChakraBadge, Container } from "@chakra-ui/react"
```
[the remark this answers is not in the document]
georgeweiler: I like the idea of having story files directly next to their component. For instance:

`src/components/Button/index.tsx`
`src/components/Button/button.stories.tsx`

georgeweiler:
@michalsmiarowski , @r-czajkowski 

I pushed up some more changes. Going to resolve some of the comments so that it's a bit clearer to see what discussions are still relevant in the review

.storybook/badge.stories.tsx
r-czajkowski: I'm not sure how hard it is to implement a rule but anyway we should address it in a follow-up PR. For now, let's keep in mind not to import React.

src/theme/Button.ts
r-czajkowski: For me works both name conversions but I would prefer to be consistent. I mean when I build a UI on the dapp side based on the Figma views I would like to use the same convention name as is on the Figma. Eg. When I select a button in Figma I see `variant: primary` but on the dapp side I have to use `solid`. I believe we should avoid a mismatch like this.

obraz

Also, I think it would be confusing on the storybook side. eg I would like to see If the component looks the same as on the Fimga but we have 2 different convention names of variants.

src/stories/button.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface ButtonStoryProps extends ButtonProps {
+  icon: string
+}
+
+const Template: Story<ButtonStoryProps> = ({
+  icon,
+  ...args
+}: ButtonStoryProps) => {
+  return (
+    <Container>
```
r-czajkowski, line 23: I'm not sure if we need `Container` component here. If you take a look at the screenshot above the code generated by storybook includes `Container` component. I think it's unnecessary.

src/stories/button.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <ChakraButton
+        {...args}
+        leftIcon={icon === "left" ? <MdBuild /> : undefined}
+        rightIcon={icon === "right" ? <MdBuild /> : undefined}
```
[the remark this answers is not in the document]
r-czajkowski: Also, I think we should document components. An [audi storybook](https://react.ui.audi/?path=/docs/components-buttons--properties) is a great example.

r-czajkowski (changes requested):
Left a few thoughts.

src/stories/badge.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+})
+
+export default {
+  title: "Badge",
+  component: Badge,
+  argTypes: {
+    colorScheme: {
+      options: colorSchemes,
+      control: { type: "radio" },
+    },
```
michalsmiarowski, line 28: As Rafał already mentioned we should document more our stories. Here i would add some description and default values which are specified in `Badge.ts`:

```suggestion
  argTypes: {
    colorScheme: {
      options: colorSchemes,
      description: "Changes the color of the badge"
      control: { type: "radio" },
      table: {
        defaultValue: {summary: 'brand'} //this is for setting default value
      },
```

We should add similar properties to all argTypes both in `Badge` and `Button`

michalsmiarowski (changes requested):

src/theme/Button.ts
georgeweiler: Yeah I hear you. I'm not 100% happy with two different variable names either, but I'm also not convinced it's worth the extra code, chances of bugs, etc with renaming the variants. This is a good call out though, and it if it's not too difficult I'm happy to update. I will play around today and investigate further.

I'm afraid we have to rebuild all the chakra CSS that the buttons receive, and for all color schemes. I have not seen any documentation on how to easily rename a variant.

I see this from the [docs:](https://chakra-ui.com/docs/theming/customize-theme)
```
const theme = extendTheme({
  components: {
    Button: {
      // We can add a new variant
      variants: {
          "with-shadow": {
                bg: "red.400",
                boxShadow: "0 0 2px 2px #efdfde",
           }
        },
        // Or we can override existing variants
        solid: {
             bg: "blue"
        },
      },
    },
  },
})
```

But I think our case is a bit more complicated than just adding box shadow.

I asked Liz about this and it took her a while to rename things in Figma so she's not willing to make adjustments on the design side.

.storybook/badge.stories.tsx
georgeweiler: FWIW I tried for about 45 min today to build a custom rule and didn't get very far. lots of annoying build errors. I'll stash the changes and try again if this becomes an issue but yeah.. let's just keep it in mind for now 👍

src/stories/button.stories.tsx
georgeweiler: nice thanks for the examples! Those were super helpful. I was able to get it working, but with a `//@ts-ignore` comment on the icon mapping. The icon prop defined by `ButtonProps` is supposed to be an element, so typescript doesn't like when you pass as a string initially and then map to an element.

I could change the interface for ButtonProps to include left & right icons as strings, but that feels weird. Happy do that that if you have a strong opinion but I'm also happy to leave a ts-ignore comment here

src/stories/button.stories.tsx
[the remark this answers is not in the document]
georgeweiler: nice thanks for the examples! Those were super helpful. I was able to get it working, but with a `//@ts-ignore` comment on the icon mapping. The icon prop defined by `ButtonProps` is supposed to be an element, so typescript doesn't like when you pass as a string initially and then map to an element.

I could change the interface for ButtonProps to include left & right icons as strings, but that feels weird. Happy do that that if you have a strong opinion but I'm also happy to leave a ts-ignore comment here

georgeweiler:
Great review @michalsmiarowski  and @r-czajkowski!  This is looking much better than my first iteration. I've added documentation and updated a few props.

I think we have one more outstanding discussion, which is how to match our Design System directly to our storybook. The button component is specifically is challenging because I don't think the design system matches up that well with chakra.


- Variant names are not consistent and renaming the variants in our codebase is not trivial (or at least I haven't found a sleek way to do it)
- There's only 1 real "colorScheme" in the design system (the brand color), and I think it's a placeholder
- consider these states from the design system that includes an "error" state:
Screen Shot 2021-09-29 at 4 06 01 PM

The problem here is that chakra has no concept of an error state, but rather a red color scheme that comes along with variants, the same as any other colorScheme would.  We could add another "error" variant that is only dark red, but I'm not sure that makes sense considering Chakra's colorScheme functionality.

Not sure what the best approach is here: we can take the time to get all the details right and make sure the design system syncs up perfectly with our storybook, or move on knowing that we just won't use some colors, props, sizes from chakra, and that the storybook might not be perfectly aligned to the TDS.

src/theme/Button.ts
r-czajkowski: Hey @georgeweiler! I think we got it!

```
import { theme } from "@chakra-ui/react"

const Button = {
 ...
 variants: {
  primary: (props: any) => {
      return {
        ...theme.components.Button.variants.solid(props),
        _disabled: {
          backgroundColor: gcfp(props, 500, "gray.400"),
        },
      }
    },
    secondary: (props: any) => {
      return {
        ...theme.components.Button.variants.outline(props),
        borderColor: gcfp(props, 600, "gray.200"),
        backgroundColor: "transparent",
        color: gcfp(props, 600, "gray.700"),
        _active: {
          backgroundColor: gcfp(props, 600, "gray.300"),
        },
        _hover: {
          borderColor: gcfp(props, 600, "gray.200"),
          backgroundColor: gcfp(props, 50, "transparent"),
          color: gcfp(props, 600, "gray.700"),
        },
        _disabled: {
          backgroundColor: "white",
        },
      }
    },
 }
}
```

Code:
```
<Button variant="primary">Primary TDS</Button>
<Button variant="solid">solid Chakra</Button>

<Button variant="secondary">secondary TDS</Button>
<Button variant="outline">outline Chakra</Button>
```

Screenshot:
obraz

r-czajkowski: So we can create our variants of buttons. Also, the `outline`/`secondary` variant doesn't fit to the TDS.

`outline`/`secondary` variant details:
- default state:
-- border color `A0AEC0` -> `grey.400`
- hover state:
-- border color `A0AEC0` -> `grey.400`
-- bg color `EDF2F7` -> `grey.100`
- active state:
-- border and bg color `CBD5E0` -> `grey.400`
- disabled state:
-- bg `white`
-- border color `E2E8F0` -> `grey.200`
-- opacity `50%`

georgeweiler: yeah this should work! Just curious - where did you find documentation on this? I was searching for a bit yesterday on how to copy variants but didn't find much help

r-czajkowski: I took a look at the `extendTheme` function and then found [this](https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/button.ts#L158-L164). There is also an issue in the official chakra repo that may interest us. Please see https://github.com/chakra-ui/chakra-ui/issues/3931

r-czajkowski: Going to resolve this convo because George addressed this in d3b7615.

r-czajkowski (approved):
I think we are good to go! Great job George!! 🚀

From my side, we should consider 2 things- non-blockers and we should address them in a follow-up PRs:
1. We are going to use custom names of the button variants but the IDE can't auto-complete them.
obraz
So I believe we should take a look a the [`chakra-cli`](https://chakra-ui.com/docs/theming/advanced#theme-typings).
2. How to document our stories? Should the button story include other props such as `onClick`, `type` and others provided by Chakra?

michalsmiarowski (approved):
Great job Chris!

I think we are ready to merge it.

One thing that also pops into my mind is that we can use [actions](https://storybook.js.org/docs/react/essentials/actions) for stories like buttons in the future, but we will address this in future PRs.

For now, I approve! Looks wonderful! 🏆

PR #4, opened 2021-09-20
Create Divider component with Icon prop

georgeweiler:


src/components/Divider.tsx
```diff
@@ -0,0 +1,37 @@
...
+  Icon,
+  Box,
+  Divider as ChakraDivider,
+  DividerProps,
+} from "@chakra-ui/react"
+
+const baseIconSize = 24
+const mdIconSize = 40
+
+export const Divider = ({ icon, ...props }: DividerProps & { icon?: any }) => (
```
r-czajkowski, line 12: Shouldn't we build a custom component in the "Chakra way"? Please see docs https://chakra-ui.com/docs/theming/advanced#multipart-or-composite-component ? Then we can define styles eg for different sizes and we will be consistent with theming API provided by Chakra. eg. of custom Button from the [Chakra docs](https://chakra-ui.com/docs/theming/component-style#styling-single-part-components):
```
  sizes: {
    sm: {
      fontSize: "sm",
      px: 4, // <-- px is short for paddingLeft and paddingRight
      py: 3, // <-- py is short for paddingTop and paddingBottom
    },
    md: {
      fontSize: "md",
      px: 6, // <-- these values are tokens from the design system
      py: 4, // <-- these values are tokens from the design system
    },
  },
```

src/components/Divider.tsx
```diff
@@ -0,0 +1,37 @@
...
+        margin="auto"
+        top={{
+          base: `${(-1 * baseIconSize) / 2}px`,
+          md: `${(-1 * mdIconSize) / 2}px`,
+        }}
+        zIndex={1}
+        w={{ base: `${baseIconSize}px`, md: `${mdIconSize}px` }}
+        h={{ base: `${baseIconSize}px`, md: `${mdIconSize}px` }}
+      />
+    )}
```
r-czajkowski, line 34: I would consider extract it to a separate component eg. `DividerIcon` so we can easily change default props. Also if we need a divider only we will use `<Divider />`. If we need a divider w/  an icon we will use:
```
<Divider>
 <DividerIcon {...ourProps} />
</Divider>
```

But maybe it's too complicated for this simple component. What do you think?

src/components/Divider.tsx
r-czajkowski, line 12: I believe we should omit the `orientation` prop from the `DividerProps` because this component only makes sense with `orientation="horizontal"` prop. Or we should support this prop in our custom component.

r-czajkowski (changes requested):

r-czajkowski (commented):
There is a warning: `Line 3: 3: 'Box' is defined but never used` in react console. In `files changed` in GH I can't see any changes in `App.tsx` file. Also the `App.tsx` doesn't contain `Box` import in `chakra-theming` branch. So I don't know how it is possible 😄

src/components/Divider.tsx
georgeweiler: I like this idea!

georgeweiler: On second thought, the Divider component from  Charka [renders an <hr> tag](https://chakra-ui.com/docs/data-display/divider#usage) which does not accept children. I think i'd prefer to extend with an `Icon` prop like we're doing now instead of trying to rebuild the divider. Thoughts?

src/components/Divider.tsx
georgeweiler: tried for a bit to see if we can support the vertical orientation but it's pretty ugly and unlikely that we'd need a vertical divider with an icon anyway. Decided just to omit the prop

georgeweiler: normally I agree this is the way to adjust styling for a component in Chakra, but we aren't actually changing the styles of the Divider, just wrapping in a relative div & adding an Icon. The Divider itself doesn't have many props besides orientation, which I've omitted. Did you have an idea of how we could do it with the theme?

georgeweiler:
I've updated to pull in the changes from the storybook config branch and added more strict types.

src/components/Divider.tsx
r-czajkowski: Yes, a good point! I'm just afraid we can't manipulate icon props. I don't think that we will always use e.g. `bg ="white"`.

src/components/Divider.tsx
r-czajkowski: Oh sorry, I provided the wrong link. I meant a `Menu` example from this [link](https://www.youtube.com/watch?v=rwjmyZUA2O8). My understanding of building components in "Chakra way":

`components/Divider.tsx`
```
import {
  Icon,
  Box,
  Divider as ChakraDivider,
  DividerProps,
  useMultiStyleConfig,
} from "@chakra-ui/react"

export const Divider = ({
  icon,
  ...props
}: Omit<DividerProps, "orientation"> & { icon?: any }) => {
  const styles = useMultiStyleConfig("Divider", {})

  return (
    <Box __css={styles.dividerWrapper}>
      {icon && <Icon as={icon} __css={styles.icon} />}
      <ChakraDivider {...props} />
    </Box>
  )
}

```

`theme/Divider.ts`

```
const baseIconSize = 24
const mdIconSize = 40

export const Divider = {
  parts: ["dividerWrapper", "divider", "icon"],
  baseStyle: (props: any) => ({
    dividerWrapper: {
      position: "relative",
      my: props.icon ? 3 : undefined,
    },
    icon: {
      bg: "white",
      color: "yellow.50",
      position: "absolute",
      left: 0,
      right: 0,
      margin: "auto",
      top: `${(-1 * baseIconSize) / 2}px`,
      zIndex: 1,
      w: `${baseIconSize}px`,
      h: `${baseIconSize}px`,
    },
  }),
  sizes: {
    md: {
      icon: {
        top: `${(-1 * mdIconSize) / 2}px`,
        w: `${mdIconSize}px`,
        h: `${mdIconSize}px`,
      },
    },
  },
  defaultProps: {
    size: "md",
  },
}

```

So in this way, we can add easily `variants`, `sizes` and `colors` of this component if needed and we are consistent with Chakra. I'm not sure how we can change the `baseStyles` via props instead of a creating another variant in the `theme/Divider.ts` file - I mean something like this `<Divider baseStyle={{ icon: {bg: "red" }  }} />`.

r-czajkowski: An example of the flexibility of this solution would be a divider with icon on the `left` or `right` side(it's just an example I know we don't need it based on the TDS). So to support that we will add variants: `left` and `right` and dedicated styles for these variants.

r-czajkowski: Oh sorry, I provided the wrong link. I meant a `Menu` example from this [link](https://chakra-ui.com/docs/theming/component-style). My understanding of building components in "Chakra way":

`components/Divider.tsx`
```
import {
  Icon,
  Box,
  Divider as ChakraDivider,
  DividerProps,
  useMultiStyleConfig,
} from "@chakra-ui/react"

export const Divider = ({
  icon,
  ...props
}: Omit<DividerProps, "orientation"> & { icon?: any }) => {
  const styles = useMultiStyleConfig("Divider", {})

  return (
    <Box __css={styles.dividerWrapper}>
      {icon && <Icon as={icon} __css={styles.icon} />}
      <ChakraDivider {...props} />
    </Box>
  )
}

```

`theme/Divider.ts`

```
const baseIconSize = 24
const mdIconSize = 40

export const Divider = {
  parts: ["dividerWrapper", "divider", "icon"],
  baseStyle: (props: any) => ({
    dividerWrapper: {
      position: "relative",
      my: props.icon ? 3 : undefined,
    },
    icon: {
      bg: "white",
      color: "yellow.50",
      position: "absolute",
      left: 0,
      right: 0,
      margin: "auto",
      top: `${(-1 * baseIconSize) / 2}px`,
      zIndex: 1,
      w: `${baseIconSize}px`,
      h: `${baseIconSize}px`,
    },
  }),
  sizes: {
    md: {
      icon: {
        top: `${(-1 * mdIconSize) / 2}px`,
        w: `${mdIconSize}px`,
        h: `${mdIconSize}px`,
      },
    },
  },
  defaultProps: {
    size: "md",
  },
}

```

So in this way, we can add easily `variants`, `sizes` and `colors` of this component if needed and we are consistent with Chakra. I'm not sure how we can change the `baseStyles` via props instead of a creating another variant in the `theme/Divider.ts` file - I mean something like this `<Divider baseStyle={{ icon: {bg: "red" }  }} />`.

r-czajkowski: An example of the flexibility of this solution would be a divider with icon on the `left` or `right` side(it's just an example I know we don't need it based on the TDS). So to support that we will add variants: `left` and `right` and dedicated styles for these variants.

georgeweiler:
Hey @r-czajkowski check the latest updates - Based on your suggestions I decided to go with an `<IconDivider />` component and chakra theme implementation.

src/components/Divider.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const Divider = ({
+  children,
+  ...props
+}: Omit<DividerProps, "orientation">) => {
+  const styles = useMultiStyleConfig("Divider", {})
+
+  return (
+    <Box __css={styles.dividerWrapper}>
+      {children}
```
r-czajkowski, line 28: I think we should use `StylesProvider` and in the `DividerIcon` component use `useStyles` hook instead of  `useMultiStyleConfig`. An advantage of this is a `useStyle` hook reads **computed** (in the parent component by  `useMultiStyleConfig("Divider", {})`) `icon` styles from styles provider. Please see an [example](https://chakra-ui.com/docs/theming/component-style#return-values).

`DividerIcon`
```
const styles = useStyles()
return <Icon as={as} __css={{ ...styles.icon, ...propStyles }} />}
```
`Divider`
```
<Box ...>
 <StylesProvider value={styles}>
  {children}
  <ChakraDivider {...props} />
 </StylesProvider>
</Box>
```

georgeweiler: 👀  this is super sleek, just updated.

r-czajkowski (changes requested):
Left a tiny comment. Once we fix that, we can merge! Looks very good! 🚀

r-czajkowski (dismissed):
LGTM! 👍 @michalsmiarowski's 👀 would be appreciated.

src/theme/Button.ts
```diff
@@ -15,28 +16,65 @@ export const Button = {
...
+      }
+    },
+    tertiary: (props: any) => {
+      const ghostStyles = theme.components.Button.variants.ghost(props)
+
+      return {
+        ...ghostStyles,
+        _hover: {
+          // @ts-ignore
+          ...ghostStyles._hover,
```
michalsmiarowski, line 66: I think you already noticed that I'm not a fan of using @ts-ignore 😅 

But here I'm not sure why it displays the error 🤔 But when i `console.log` the `ghostStyles` it seems that the `_hover` property has only one property which is `bg`, and we are already defining that. So i think we can safely remove .`..ghostStyles._hover`, `...ghostStyles._active` and .`..ghostStyles._disabled`.

michalsmiarowski (changes requested):
Some comments to look at below.

michalsmiarowski (approved):
LGTM!

r-czajkowski (approved):
:shipit:

PR #1, opened 2021-09-13
initial CRA setup

georgeweiler:
major libraries installed with this PR:

- create-react-app
- eslint / prettier
- husky / lint-staged
- storybook
- chakra-ui

PR #5, opened 2021-09-21
CI workflows configuration

nkuba:
WIP

Depends on https://github.com/threshold-network/token-dashboard/pull/1


.github/workflows/format.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  dashboard-code-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "16"
```
nkuba, line 18: Are there any expectations regarding the node version the dashboard should be built with?

nkuba:
The current solution assumes that mainnet deployments are made by `releases/mainnet/**` branch. We can optimize it and invoke mainnet deployment on the `main` merge, with a requirement that each deployment will have to be approved manually. In such a case, we would download the latest contracts versions that are tagged in the npm registry with `mainnet` tag.

.github/workflows/dashboard.yaml
```diff
@@ -0,0 +1,107 @@
+name: Dashboard
```
lukasz-zimnoch, line 1: Can we align workflows and files naming to what we have in other repos (e.g. `keep-core`). We could name this file as `dashboard-testnet.yml` and the workflow as `Token Dashboard / Testnet`. WDYT?

.github/workflows/release-mainnet.yml
```diff
@@ -0,0 +1,95 @@
+name: Dashboard Build and Deploy for Mainnet
```
lukasz-zimnoch, line 1: And same here regarding naming. What do you think about naming the file `dashboard-mainnet.yml` and the workflow as `Token Dashboard / Mainnet`?

.github/workflows/release-mainnet.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Build
+        run: yarn build
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: build
+          path: build
+
+  deploy-preview-mainnet:
```
lukasz-zimnoch, line 46: Maybe `deploy-preview`?

.github/workflows/release-mainnet.yml
```diff
@@ -0,0 +1,95 @@
...
+  pull_request:
+    branches:
+      - releases/mainnet/**
+  release:
+    types:
+      - "published"
+  workflow_dispatch:
+
+jobs:
+  build-mainnet:
```
lukasz-zimnoch, line 16: Maybe just `build` as we are in mainnet workflow?

.github/workflows/release-mainnet.yml
```diff
@@ -0,0 +1,95 @@
...
+        with:
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Preview uploaded to https://preview.dashboard.threshold.network/${{ github.head_ref }}/.'
+            })
+
+  deploy-mainnet:
```
lukasz-zimnoch, line 76: `deploy`?

.github/workflows/dashboard.yaml
```diff
@@ -0,0 +1,107 @@
...
+        run: yarn install --frozen-lockfile
+
+      - name: Build
+        run: yarn build
+
+      # FIXME: Tests are missing, will this step be needed?
+      # - name: Test
+      #   run: yarn test
+
+  build-and-deploy-testnet:
```
lukasz-zimnoch, line 49: Just `build-and-deploy` since we are already in testnet workflow context?

.github/workflows/dashboard.yaml
georgeweiler: I'm cool with these changes & the related change requests so we're keeping things consistent with Keep.

.github/workflows/dashboard.yaml
nkuba: We try to keep the jobs naming unique across the workflows as branch protection rules use jobs names and we don't want them to overlap and be confused there.

.github/workflows/release-mainnet.yml
nkuba: We try to keep the jobs naming unique across the workflows as branch protection rules use jobs names and we don't want them to overlap and be confused there.

.github/workflows/release-mainnet.yml
nkuba: We try to keep the jobs naming unique across the workflows as branch protection rules use jobs names and we don't want them to overlap and be confused there.

.github/workflows/release-mainnet.yml
nkuba: We try to keep the jobs naming unique across the workflows as branch protection rules use jobs names and we don't want them to overlap and be confused there.

.github/workflows/release-mainnet.yml
nkuba: After reconsideration, I decided to rename the jobs as you suggested, we won't use them for branch rules.

.github/workflows/dashboard-ci.yaml
```diff
@@ -0,0 +1,107 @@
...
+          bucket-name: preview.dashboard.test.threshold.network
+          bucket-path: ${{ github.head_ref }}
+          build-folder: build
+
+      - name: Post preview URL to PR
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v3
+        with:
+          script: |
+            github.issues.createComment({
```
michalinacienciala, line 89: We need to call the `issues.createComment` command using `github.rest.` context (["the Octokit context available via github no longer has REST methods directly"](https://github.com/actions/github-script#breaking-changes-in-v5)). Also, we need to use `v5` of the `github-script` action, which supports the new changes.

```suggestion
        uses: actions/github-script@v5
        with:
          script: |
            github.rest.issues.createComment({
```

FYI, I'm going to apply same change to the https://github.com/keep-network/keep-core/blob/main/.github/workflows/dashboard-mainnet.yml workflow, which also uses the `github-script` action.

.github/workflows/dashboard-mainnet.yml
```diff
@@ -0,0 +1,96 @@
...
+          bucket-name: preview.dashboard.threshold.network
+          bucket-path: ${{ github.head_ref }}
+          build-folder: build
+
+      - name: Post mainnet preview URL to PR
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v3
+        with:
+          script: |
+            github.issues.createComment({
```
michalinacienciala, line 70: We need to call the `issues.createComment` command using `github.rest.` context (["the Octokit context available via github no longer has REST methods directly"](https://github.com/actions/github-script#breaking-changes-in-v5)). Also, we need to use `v5` of the `github-script` action, which supports the new changes.

```suggestion
        uses: actions/github-script@v5
        with:
          script: |
            github.rest.issues.createComment({
```

FYI, I'm going to apply same change to the https://github.com/keep-network/keep-core/blob/main/.github/workflows/dashboard-mainnet.yml workflow, which also uses the `github-script` action.

michalinacienciala (commented):
Left one comment - it would be good to fix this at some point, but I don't think it should block the merge.

georgeweiler (approved):

.github/workflows/dashboard-mainnet.yml
```diff
@@ -0,0 +1,96 @@
...
+  deploy-preview:
+    name: Deploy mainnet preview
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/download-artifact@v2
+
+      - name: Deploy PR mainnet preview to GCP
```
michalinacienciala, line 55:
```suggestion
      - name: Deploy mainnet preview to GCP
```

michalinacienciala (changes requested):

georgeweiler (approved):

PR #6, opened 2021-09-22
Configure commitlint

georgeweiler:
depends on #1


pdyraga:
Errr... somethins is off here. Should we see one more commit in this branch with commitlint configuration?

I have not merged this PR. I merged #1 and then it was automatically marked as merged by GitHub.

georgeweiler:
darn.. I was trying to be fancy by rebasing the changes out of the other PR into this one but I think I just cloned the same commit history, so github was confused. I'll create a new branch and add these changes in.

PR #10, opened 2021-09-30
cypress environment setup

georgeweiler:


georgeweiler:
I think we can configure the tests to run before merging into main after #5 Is merged

cypress/integration/App.spec.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="../support" />
+
+describe("My First Test", () => {
+  it("Opens the app", () => {
+    cy.visit("/")
+    cy.dataCy("app-container")
+  })
+})
+
+export {}
```
r-czajkowski, line 10: Do we need this?

cypress/support/index.ts
```diff
@@ -0,0 +1,7 @@
+function dataCy(identifier: string) {
+  return cy.get(`[data-cy=${identifier}]`)
+}
+
+Cypress.Commands.add("dataCy", dataCy)
+
+export {}
```
r-czajkowski, line 7: Same as above

cypress/integration/App.spec.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="../support" />
```
r-czajkowski, line 1: Do we need this? Looks like it works w/o this in VSC.

cypress/support/index.d.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable no-unused-vars */
+/// <reference types="cypress" />
+
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    dataCy(value: string): Chainable<Element>
```
r-czajkowski, line 6: Can we add docs for this method? eg. https://docs.cypress.io/guides/tooling/typescript-support#Types-for-custom-commands

cypress/support/index.ts
```diff
@@ -0,0 +1,7 @@
+function dataCy(identifier: string) {
```
r-czajkowski, line 1: Can we mention that it is a good practice based on the cypress docs? I was confused a little bit about why we need this function. We can add here or in the interface- I mentioned adding docs in a comment above.

package.json
```diff
@@ -35,7 +35,8 @@
     "prepare": "husky install",
     "storybook": "start-storybook -p 6006 -s public",
     "build-storybook": "build-storybook -s public",
-    "chromatic": "npx chromatic --exit-zero-on-changes"
+    "chromatic": "npx chromatic --exit-zero-on-changes",
+    "cypress:open": "cypress open"
```
r-czajkowski, line 39: I've run the cypress locally and it auto-generated some files. Do we need them? If not, we should ignore them in the `.gitigone`, `.eslintignore` `.prettierignore` files.

obraz

r-czajkowski (changes requested):
Cypress looks so good! I can't wait to write the integration tests. Left a few comments.

I'll ping @michalinacienciala here- could you please take a look at the `cypress` workflow?

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,15 @@
+name: Cypress E2E Tests
+
+on: [push]
```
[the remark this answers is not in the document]
georgeweiler: ooh yeah I like the idea of a manual trigger. The e2e tests might take a while, which is why I'm wondering if it makes sense to run on every commit. Maybe it doesn't matter?

cypress/support/index.ts
georgeweiler: replied above

cypress/support/index.d.ts
georgeweiler: yep good idea - i just copied the documentation comment from their docs

.github/workflows/cypress.yml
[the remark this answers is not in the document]
r-czajkowski: As far as I know `on: [push]` runs workflow on every commit on every branch. Suggested changes run a workflow:
* on push to `main` branch,
* when opening PR or pushing new commit on the feature branch when PR is already open.

cypress/integration/App.spec.ts
r-czajkowski: I see... But it looks like we are using the different ts config. Based on the [cypress docs](https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfig-json) they recommend adding tsconfig.json file to the cypress folder. After adding config file my IDE shows me the same ts error when I removed the `/// <reference types="../support" />` but I can't see error because the file has no import or exports. So maybe we should add another `tsconfig` inside `cypress` dir?

cypress/integration/App.spec.ts
r-czajkowski: omg I see :( I replied above please see https://github.com/threshold-network/token-dashboard/pull/10#discussion_r722961521

.github/workflows/cypress.yml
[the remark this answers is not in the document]
michalinacienciala: Yes, `workflow_dispatch` triggering event can be useful. With this trigger present, workflow can be triggered by going to Actions -> Cypress E2E Tests -> Run workflow (the option is not present when `workflow_dispatch` trigger is absent in the workflow config file on `main`).
And in regards to other triggering conditions - we have some other workflows which run different checks for all changes in the pull requests, so we should catch some problems thanks to them. If we expect the E2E Cypress tests to last long and don't want to wait for their execution before merging the changes to main, we could set the workflow to run on workflow_dispatch and on ever push to `main` (i.e. merging of PR to `main`). We could also/alternatively configure the workflow to be triggered by cron, for example every day at midnight.

```suggestion
on:
  push:
    branches:
     - main
  schedule:
    - cron: '0 0 * * *'
  workflow_dispatch:
```

It would be good to not rely just on the cron trigger, as GH poorly handles the notifications about failures for workflows triggered by `schedule` event.

michalinacienciala (changes requested):
I took a look at the workflow, left one suggestion

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,21 @@
+name: Cypress E2E Tests
+
+on:
+  push:
+    branches:
+      - main
+    schedule:
+      - cron: "0 0 * * *"
+    workflow_dispatch:
```
michalinacienciala, line 9:
```suggestion
  schedule:
    - cron: "0 0 * * *"
  workflow_dispatch:
```
The `schedule` and `workflow_dispatch` events should be at the same level as `push`.

michalinacienciala (changes requested):

michalinacienciala (approved):

PR #12, opened 2021-10-04
Create redux store with an example modal

georgeweiler:
- Set up redux toolkit
- Build the `<RootModal/>` component that reads information from the store

### DIFF
--- a/src/components/Modal/ExampleModal.tsx
+++ b/src/components/Modal/ExampleModal.tsx
@@ -1,0 +1,33 @@
+import {
+  Button,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+} from "@chakra-ui/react"
+import { useModal } from "../../store/modal"
+
+const ExampleModal = () => {
+  const { closeModal, modalProps } = useModal()
+
+  return (
+    <Modal isOpen onClose={closeModal}>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Modal Title</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>Hello {modalProps.name}!</ModalBody>
+        <ModalFooter>
+          <Button colorScheme="blue" mr={3} onClick={closeModal}>
+            Close
+          </Button>
+          <Button variant="ghost">Secondary Action</Button>
+        </ModalFooter>
+      </ModalContent>
+    </Modal>
+  )
+}
+export default ExampleModal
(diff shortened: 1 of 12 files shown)


r-czajkowski (changes requested):
I really like the idea! Looks super flexible!. Left few thoughts.

src/components/Modal/ExampleModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+  Button,
+  ModalBody,
+  ModalCloseButton,
+  ModalFooter,
+  ModalHeader,
+} from "@chakra-ui/react"
+import { FC } from "react"
+import withBaseModal from "./withBaseModal"
+
+const ExampleModal: FC<{ name: string; closeModal: () => void }> = ({
```
r-czajkowski, line 11: I think we should use it here `BaseModalProps`. `FC<{name: string } & BaseModalProps>`

src/components/Modal/withBaseModal.tsx
```diff
@@ -0,0 +1,23 @@
+import { ComponentType } from "react"
+import { Modal, ModalContent, ModalOverlay } from "@chakra-ui/react"
+
+interface BaseModalProps {
+  closeModal: () => void
+}
```
r-czajkowski, line 6: regarding the suggestion above- maybe we should move the `BaseModalProps` to the `components/Modal/index.ts` or to `components/Modal/types.ts` and then import and export them in `components/Modal/index.ts`.

`components/Modal/types.ts`
```
interface BaseModalProps {
  closeModal: () => void
}
<other types here>
```
`components/Modal/index.ts`
```
import { BaseModalProps } from "./types"

export { BaseModalProps } 
```

src/types.ts
```diff
@@ -0,0 +1,21 @@
+import { ModalType } from "./enums"
```
r-czajkowski, line 1: Because in 6e38e18a683d9c94be27ce9b1339404d08202a03 we decided to keep the types in `types/index.ts` let's move these types to `types/index.ts`.

src/enums.ts
```diff
@@ -0,0 +1,3 @@
+export enum ModalType {
+  Example = "EXAMPLE",
```
r-czajkowski, line 2: Same as above. To be consistent- let's move to `enums/index.ts`

r-czajkowski (changes requested):
Left really minor comments. Let's address them and I'm ok with merging! 🚀

michalsmiarowski (approved):
Looking good. Didn't find any issues! Also tested locally and the modal looks and works awesome. Let's go 🚀

r-czajkowski (approved):
:shipit:

PR #14, opened 2021-10-05
Alert component styling

georgeweiler:
- updates the chakra alert theme to match the TDS

### DIFF
--- a/src/stories/alert.stories.tsx
+++ b/src/stories/alert.stories.tsx
@@ -1,0 +1,48 @@
+import {
+  Alert as ChakraAlert,
+  AlertDescription,
+  AlertIcon,
+  AlertProps,
+  AlertTitle,
+  Stack,
+} from "@chakra-ui/react"
+import { Meta, Story } from "@storybook/react"
+
+const Template: Story<AlertProps> = ({ ...args }) => {
+  return (
+    <Stack spacing={8}>
+      <ChakraAlert {...args}>
+        <AlertIcon />
+        This is an alert
+      </ChakraAlert>
+      <ChakraAlert {...args}>
+        <AlertIcon />
+        <Stack>
+          <AlertTitle>Alert Title Goes Here</AlertTitle>
+          <AlertDescription>
+            Here's a description about something the user should be aware of
+          </AlertDescription>
+        </Stack>
+      </ChakraAlert>
+    </Stack>
+  )
+}
+
+export const Alert = Template.bind({
+  status: "info",
+})
+
+export default {
+  title: "Alert",
+  component: Alert,
+  argTypes: {
+    status: {
+      description: "Adjusts the status of the alert",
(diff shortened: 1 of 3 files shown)


src/theme/Alert.ts
```diff
@@ -0,0 +1,106 @@
...
+  return {}
+}
+
+export const Alert = {
+  baseStyle: (props: any) => ({
+    container: {
+      borderRadius: "md",
+      border: "1px solid",
+      borderColor: getColorFromProps(props, 500),
+    },
```
r-czajkowski, line 98: Shouldn't we add a base style for `icon`? [Based on the TDS](https://www.figma.com/file/zZi2fYDUjWEMPQJWAt8VWv/Threshold-DS?node-id=877%3A18956), the icon should always be at the top instead of a center of the `Alert` container. Eg:

```
icon: {
 alignSelf: "flex-start",
},
```

src/theme/Alert.ts
```diff
@@ -0,0 +1,106 @@
...
+    return {
+      container: {
+        backgroundColor: mode("white", transparentize("white", 0.16))(props),
+        borderColor: "gray.200",
+      },
+      icon: {
+        color: mode("gray.500", "white")(props),
+      },
+    }
+  }
```
r-czajkowski, line 28: Nested ifs everywhere 😄. I took a look at the Chakra code and it looks like they build variants based on the `colorSchema`  prop and create a separate function for each variant. Can we take the same approach? Please see:
* [component code](https://github.com/chakra-ui/chakra-ui/blob/main/packages/alert/src/alert.tsx#L17-L22)
* [theme code](https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/alert.ts#L30-L43)
* variant functions:
** https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/alert.ts#L45
** https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/alert.ts#L75

georgeweiler:
@r-czajkowski I've pushed up some refactor changes to move away from the "if" statements. lmk what you think

src/theme/Alert.ts
georgeweiler: Check my latest update on this. Was able to do something a bit more concise.

src/stories/alert.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+            Here's a description about something the user should be aware of
+          </AlertDescription>
+        </Stack>
+      </ChakraAlert>
+    </Stack>
+  )
+}
+
+export const Alert = Template.bind({
+  status: "info",
```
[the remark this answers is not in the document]
r-czajkowski: Great! Thanks for the heads up! 🚀

r-czajkowski (approved):
LGTM! 👍

PR #25, opened 2021-11-09
Trezor connector with multi address selection

georgeweiler:


Battenfield (approved):

Battenfield:
merging, will circle back when we bring back native hw

PR #28, opened 2021-11-17
Remove hardware options from wallet dropdown list

georgeweiler:


Battenfield (approved):

Battenfield:
descoping hw for now, circle back when we have more bandwidth

PR #29, opened 2021-11-23
Adds connect wallet CTA to upgrade card

georgeweiler:


src/components/UpgradeCard/UpgradeCardTemplate.tsx
```diff
@@ -77,9 +82,19 @@ const UpgradeCardTemplate: FC<UpgradeCardTemplateProps> = ({
...
+          </Button>
+        ) : (
+          <Button
+            mt={6}
+            isFullWidth
+            onClick={() => openModal(ModalType.SelectWallet)}
+          >
+            Connect Wallet
+          </Button>
+        )}
```
r-czajkowski, line 97: Can we extract it to a separate component? My point is we will probably need the same solution for other transactions so I believe we should create a reusable component. Eg `SubmitTxButton`? Maybe you will have a better idea of the name of this component 🙂.

r-czajkowski (changes requested):

src/components/UpgradeCard/UpgradeCardTemplate.tsx
georgeweiler: good call, i've updated.

src/components/SubmitTxButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useModal } from "../hooks/useModal"
+
+const SubmitTxButton: FC<{ onSubmit: () => void }> = ({ onSubmit }) => {
+  const { active } = useWeb3React()
+  const { openModal } = useModal()
+
+  if (active) {
+    return (
+      <Button mt={6} isFullWidth onClick={onSubmit}>
+        Upgrade
```
r-czajkowski, line 14: Can we add eg. `text` prop and render here? Or maybe pass rest props to the `Button` so we will be able to pass any button props(w/o `onClick` prop) and children. eg

`SubmitTxButton.tsx`
```
if(active){
 return (
  <Button ...restProps onClick={onSubmit} />
 )
}
```

`AnyComponent.tsx`
```
<SubmitTxButton isFullWidth mt={6} onSubmit={} <other Chakra Button prosp> >
 Upgrade
</SubmitTxButton>
```

What do you think?

r-czajkowski (changes requested):

r-czajkowski (approved):
LGTM!

PR #30, opened 2021-11-23
Add a supported chain id to the env

georgeweiler:


.gitignore
```diff
@@ -16,3 +16,5 @@ cypress/plugins
 # Yarn
 node_modules/
 yarn-error.log
+
+.env
```
r-czajkowski, line 20: Can we add the `.env` file with default values for local development?

```
REACT_APP_SUPPORTED_CHAIN_ID=1337
REACT_APP_RPC_URL=http://localhost:8545
REACT_APP_WS_URL=ws://localhost:8545
```

r-czajkowski, line 20: I think we can add `.env.production` file as well which will be used during the build process and the file might look like:

`.env.production`
```
REACT_APP_SUPPORTED_CHAIN_ID=$chain_id
REACT_APP_RPC_URL=$rpc_url
REACT_APP_WS_URL=$ws_url
```

where environment variables(`chain_id`, `rpc_url`, `ws_url`) will be set in CI side.

src/utils/isSupportedNetwork.ts
```diff
@@ -1,9 +1,19 @@
-import { ChainID } from "../enums"
+import { getEnvVariable } from "./getEnvVariable"
+import { EnvVariable } from "../enums"
+import * as net from "net"
```
r-czajkowski, line 3: I think it is unnecessary import.

r-czajkowski (changes requested):

src/web3/connectors/walletConnect.ts
```diff
@@ -1,7 +1,18 @@
...
-import { RPC_URL } from "../../config"
+import { getEnvVariable } from "../../utils/getEnvVariable"
+import { EnvVariable } from "../../enums"
+
+const rpcUrl = getEnvVariable(EnvVariable.RpcUrl)
+const supportedChainId = getEnvVariable(EnvVariable.SupportedChainId)
+
+console.log({
+  [Number(supportedChainId)]: rpcUrl as string,
+})
```
r-czajkowski, line 10: Do we need this?

r-czajkowski (approved):
LGTM!

nkuba:
@georgeweiler having a PR description provided would be super helpful.

src/web3/connectors/walletConnect.ts
r-czajkowski: Addressed in https://github.com/threshold-network/token-dashboard/pull/30/commits/2dcbcea3fabbb2e8818d9b6ada43dca2c1b83ad7

r-czajkowski (approved):
LGTM! Ropsten preview looks good and I was able to test the migration flow so let's :shipit:

PR #34, opened 2021-12-10
Overview page

georgeweiler:


src/pages/Overview/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    {
+      text: "Network",
+      href: "/network",
+      isActive: pathname.includes("/network"),
+    },
+    { text: "tBTC", href: "/tBTC", isActive: pathname.includes("/tBTC") },
+    { text: "PRE", href: "/pre", isActive: pathname.includes("/pre") },
+  ]
+
+  const { path } = useRouteMatch()
```
r-czajkowski, line 28: I think we can use a similar pattern as we used in KEEP token dashboard for pages components. In KEEP token dashboard the page component has an additional field `route` where you can define:
- title- title of the page,
- path- path to the page,
- pages- a sub pages, an array of page components.

And the main page component should be wrapped with `PageWrapper` component which render the sub pages. Please see how it works in the KEEP dapp and let me know what do you think. Main routing components in KEEP dapp:
- https://github.com/keep-network/keep-core/blob/main/solidity-v1/dashboard/src/components/Routing.jsx
- https://github.com/keep-network/keep-core/blob/main/solidity-v1/dashboard/src/components/PageWrapper.jsx
- https://github.com/keep-network/keep-core/blob/main/solidity-v1/dashboard/src/components/Header.jsx
- example of main page https://github.com/keep-network/keep-core/blob/main/solidity-v1/dashboard/src/pages/threshold/index.js
- example of sub page https://github.com/keep-network/keep-core/blob/main/solidity-v1/dashboard/src/pages/threshold/ThresholdUpgradePage.jsx#L320-L324

In T dapp we could add an additional field indicating which sub navigation component should be rendered in `PageWrapper`. Based on the Figma views we need 2 types of navigation:

<details>
  <summary>First type</summary>
  
  image

</details>

<details>
  <summary>Second type</summary>
  
image

</details>

eg. `Overview` page in T dapp:
```
OverviewPage.route = {
 path: "/overview",
 pages: [Network, tBTC, Pre],
 subNavigationStrategy: FirstTypeOfNavigation or SecondTypeOfNavigation
}
```

r-czajkowski (changes requested):
Did a first pass and left some thoughts. I would like to flag that almost the entire HTML tree contains only `divs`- I believe we should pay more attention to semantic tags. Eg side bar component doesn't contain `nav` tag, we don't use `ul` tag to list item(eg. to list navigation items), we don't use the `header` tag in NavBar component etc.

Battenfield (commented):
Screen Shot 2021-12-15 at 9 57 12 AM

subnav needs a dark mode text

src/pages/Overview/index.tsx
georgeweiler: I like this idea! here are a few thoughts:

- 100% agree that we need to create a smarter Route component. Listing the Routes in App.jsx is fine for now, but will only get us so far. The pattern of including some routing metadata on the component is really cool.

- I think we could drop the subNavigation strategy because the designs have changed a bit. Hopefully we only need one design.

- I'm not sure if we need the page wrapper component that defines styling for the header and subnavs. Worried that designs for other pages will change for other pages and we'll be have trouble extending the component to fit the new page designs. 


 Would you be ok with taking on this sub-navigation in a separate PR?

r-czajkowski: > I'm not sure if we need the page wrapper component that defines styling for the header and subnavs. Worried that designs for other pages will change for other pages and we'll be have trouble extending the component to fit the new page designs.

Yeah, that's one of the cons...

> Would you be ok with taking on this sub-navigation in a separate PR?
sure!

src/components/Sidebar/index.tsx
```diff
@@ -16,25 +16,29 @@ import { IoChartOutlineSharp } from "../../static/icons/IoChartOutlineSharp"
 const Sidebar = () => {
   const { pathname } = useLocation()
 
+  const isOverview = useRouteMatch("/overview")
```
georgeweiler, line 19: @r-czajkowski what do you think about keeping useRouteMatch here? The issue is that the icon may depend on the route match. We could pass `passiveIcon` and `activeIcon`?

r-czajkowski: I think we should use the `useRouteMatch` per `NavItem`, because the element should be responsible for recognizing the current active path. I believe it will better work with the routing metadata approach. Also if we add a new page we will have to remember to use `useRouteMatch`.

src/components/SubNavigationPills/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Stack,
+  useColorModeValue,
+} from "@chakra-ui/react"
+import { Link as RouterLink, useRouteMatch } from "react-router-dom"
+
+interface Props {
+  links: {
+    text: string
+    href: string
+    isActive: boolean
```
r-czajkowski, line 16: Let's use `useRouteMatch` instead of `isActive` prop.

src/components/SubNavigationPills/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }[]
+}
+
+const SubNavigationPills: FC<Props> = ({ links }) => {
+  const { url } = useRouteMatch()
+
+  return (
+    <Box w="100%" borderBottom="1px solid" borderColor="gray.100" padding={6}>
+      <HStack
+        spacing={4}
```
r-czajkowski, line 26: Can we `as="ul"` and then for children `as="li"`?

src/pages/Overview/Network/Interest.tsx
```diff
@@ -0,0 +1,25 @@
+import { FC } from "react"
+import CardTemplate from "./CardTemplate"
+import { HStack } from "@chakra-ui/react"
+import { Label3 } from "../../../components/Typography"
```
r-czajkowski, line 4: Unnecessary import.

src/pages/Overview/Network/Nodes.tsx
```diff
@@ -0,0 +1,24 @@
+import { FC } from "react"
+import CardTemplate from "./CardTemplate"
+import { Stack, useMediaQuery, HStack } from "@chakra-ui/react"
```
r-czajkowski, line 3: Unnecessary imports `Stack, useMediaQuery`.

src/pages/Overview/Network/UpgradeBanner.tsx
```diff
@@ -0,0 +1,77 @@
+import { FC } from "react"
+import Card from "../../../components/Card"
+import {
+  Button,
+  CloseButton,
+  HStack,
```
r-czajkowski, line 6: Unnecessary import.

src/pages/Overview/Network/UpgradeBanner.tsx
```diff
@@ -0,0 +1,77 @@
...
+            mt={{ base: 12, xl: "auto !important" }}
+          >
+            <Button
+              size={isMobile ? "sm" : "lg"}
+              px={{ base: 4, md: 12 }}
+              isFullWidth
+            >
+              {isMobile ? "Upgrade Now" : "Upgrade Tokens"}
+            </Button>
+          </Link>
```
r-czajkowski, line 70: same as above about `button` inside `a` tag.

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,126 @@
...
+  balance: number
+  icon: IconEnum
+  text: string
+  conversionRate?: number
+}> = ({ balance, icon, text, conversionRate }) => {
+  return (
+    <HStack justify="space-between">
+      <HStack>
+        <Icon as={icon} boxSize="24px" />
+        <H5>{numeral(balance).format("0,00.00")}</H5>
```
r-czajkowski, line 31: Could we use `TokenBalance` component?

src/pages/Overview/index.tsx
```diff
@@ -0,0 +1,52 @@
+import { Container } from "@chakra-ui/react"
+import {
+  Redirect,
+  Route,
+  Switch,
```
r-czajkowski, line 5: Unnecessary import.

src/pages/Overview/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+const Overview = ({}) => {
+  const { pathname } = useLocation()
+
+  const subNavLinks = [
+    {
+      text: "Network",
+      href: "/network",
+      isActive: pathname.includes("/network"),
+    },
+    { text: "tBTC", href: "/tBTC", isActive: pathname.includes("/tBTC") },
```
r-czajkowski, line 25: let's remove `isActive`. The `SubNavigationPills` should be responsible for this.

src/pages/Overview/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { path } = useRouteMatch()
+
+  const isMobile = useChakraBreakpoint("md")
+
+  return (
+    <Container maxW={{ base: "2xl", xl: "6xl" }} mt={16}>
+      <Label1>Threshold</Label1>
+      {isMobile ? <H3>Network Overview</H3> : <H1>Network Overview</H1>}
+      <SubNavigationPills links={subNavLinks} />
+      <Redirect from="/overview" to="/overview/network" />
```
r-czajkowski, line 38: Should't we wrap routes with `Switch` component from `react-router-dom`?

r-czajkowski (changes requested):

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,126 @@
...
+      <Body2>Possible T amount</Body2>
+      <HStack
+        bg={useColorModeValue("gray.50", "gray.700")}
+        mt={4}
+        px={6}
+        py={2}
+        borderRadius="md"
+      >
+        <Icon as={t.icon} boxSize="32px" />
+        <H3>{numeral(conversionToTAmount).format("0,00.00")}</H3>
```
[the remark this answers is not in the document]
r-czajkowski: Let's create a util function that format amount eg:

`utils/formatAmount.ts`
```
import { formatUnits } from "@ethersproject/units"

export formatTokenAmount = (rawAmount, format = "0,00.00", decimals= 18) => {
 return numeral(formatUnits(rawAmount, decimals).format("0,00.00")
}

export formatFiatCurrencyAmount = (amount, format = "0,00" currencySymbol="$") => {
 return numeral(amount).format(`${currencySymbol}${format}`)
}
```

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,126 @@
...
+    </HStack>
+  )
+}
+
+const WalletBalances: FC = () => {
+  const { account } = useWeb3React()
+  const { keep, nu, t } = useReduxToken()
+
+  // do to: figure out how to pass in the token decimals
+  const formattedKeep = Number(formatUnits(keep.balance, 6))
```
r-czajkowski, line 48: Let's remove 6.

r-czajkowski (changes requested):

src/pages/Overview/Network/WalletBalances.tsx
georgeweiler: Don't we need the decimal identifier for KEEP here?

src/pages/Overview/index.tsx
georgeweiler: I think switch is not needed here because they are sub routes, and we want the path to match inclusively. https://v5.reactrouter.com/web/api/Switch

The sub routing doesn't work if you add a switch wrapper.

src/pages/Overview/index.tsx
georgeweiler: 👍  Ok I will add the activeIcon and passiveIcon props.

src/pages/Overview/Network/WalletBalances.tsx
r-czajkowski: KEEP token has 18 decimals and `formatUnits` fn uses `18` by default. So I think we can get rid of `6` 🤔

src/utils/formatAmount.ts
```diff
@@ -0,0 +1,22 @@
...
+export const formatNumeral = (number: string | number, format = "0,00.00") => {
+  return numeral(number).format(format)
+}
+
+export const formatTokenAmount = (
+  rawAmount: number | string,
+  format = "0,00.00",
+  decimals = 18
+) => {
+  return numeral(formatUnits(rawAmount, decimals)).format(format)
```
r-czajkowski, line 13: Let's use `formatNumeral` fn here.
```
export const formatTokenAmount = (
  rawAmount: number | string,
  format = "0,00.00",
  decimals = 18
) => {
 return formatNumeral(formatUnits(rawAmount,decimals), format)
}
```

src/utils/formatAmount.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export const formatFiatCurrencyAmount = (
+  amount: number | string,
+  format = "0,00",
+  currencySymbol = "$"
+) => {
+  return numeral(amount).format(`${currencySymbol}${format}`)
```
r-czajkowski, line 21: Same as above.

src/store/tokens/tokenSlice.ts
```diff
@@ -35,6 +35,7 @@ export const tokenSlice = createSlice({
       icon: Icon.KeepCircleBrand,
       usdConversion: 0,
       usdBalance: "0",
+      decimals: 6,
```
r-czajkowski, line 38: KEEP/NU/T token has 18 decimals.

r-czajkowski (changes requested):

src/pages/Overview/index.tsx
r-czajkowski: Based on the [docs](https://v5.reactrouter.com/web/example/nesting ) we should probably use the `Switch` statement for sub routes.

Regarding `The sub routing doesn't work if you add a switch wrapper`- because the `<Redirect from="/overview" to="/overview/network" />` is first. From docs:
>Renders the first child <Route> or <Redirect> that matches the location.

So If we use `<Redirect from="/overview" to="/overview/network" />` or pass `exact` prop it works.

I still don't understand why we use `const subroute = pathname.slice(url.length); const isActive = link.href === subroute` in `SubNavigationPills` component 😆. Here is my suggestion:

`src/pages/Overview/index.tsx`
```
import { Container } from "@chakra-ui/react"
import { Redirect, Route, useRouteMatch, Switch } from "react-router-dom"
import { H1, H3, Label1 } from "../../components/Typography"
import SubNavigationPills from "../../components/SubNavigationPills"
import Network from "./Network"
import TBTC from "./tBTC"
import Pre from "./Pre"
import useChakraBreakpoint from "../../hooks/useChakraBreakpoint"

const subNavLinks = [
  {
    text: "Network",
    path: "/network",
  },
  { text: "tBTC", path: "/tBTC" },
  { text: "PRE", path: "/pre" },
]

const Overview = ({}) => {
  const { path } = useRouteMatch()

  const isMobile = useChakraBreakpoint("md")

  return (
    <Container maxW={{ base: "2xl", xl: "6xl" }} mt={16}>
      <Label1>Threshold</Label1>
      {isMobile ? <H3>Network Overview</H3> : <H1>Network Overview</H1>}
      <SubNavigationPills parentPath={path} links={subNavLinks} />
      <Switch>
        <Route path={`${path}/network`}>
          <Network />
        </Route>
        <Route path={`${path}/tBTC`}>
          <TBTC />
        </Route>
        <Route path={`${path}/pre`}>
          <Pre />
        </Route>
        <Redirect from="/overview" to="/overview/network" />
      </Switch>
    </Container>
  )
}

export default Overview

```

`src/components/SubNavigationPills/index.tsx`
```
import { FC, createContext, useContext } from "react"
import {
  Box,
  Divider,
  HStack,
  Link,
  Stack,
  useColorModeValue,
} from "@chakra-ui/react"
import { Link as RouterLink, useRouteMatch } from "react-router-dom"

interface NavPillProps {
  path: string
  text: string
}

interface Props {
  links: NavPillProps[]
  parentPath: string
}

const SubNavigationPillContext = createContext({ parentPath: "" })

const useSubNavigationPillContext = () => {
  return useContext(SubNavigationPillContext)
}

const SubNavigationPills: FC<Props> = ({ parentPath, links }) => {
  return (
    <SubNavigationPillContext.Provider value={{ parentPath }}>
      <Box
        w="100%"
        borderBottom="1px solid"
        borderColor="gray.100"
        padding={6}
        as="nav"
      >
        <HStack
          spacing={4}
          divider={<Divider orientation="vertical" borderColor="gray.300" />}
          height="28px"
          as="ul"
        >
          {links.map(renderPill)}
        </HStack>
      </Box>
    </SubNavigationPillContext.Provider>
  )
}

const NavPill: FC<NavPillProps> = ({ path, text }) => {
  const { parentPath } = useSubNavigationPillContext()
  const _path = `${parentPath}${path}`
  const isActive = useRouteMatch({ path: _path, exact: true })

  return (
    <Stack position="relative" padding={2} as="li">
      <Link
        fontWeight={isActive ? "700" : undefined}
        color={isActive ? useColorModeValue("brand.500", "white") : undefined}
        as={RouterLink}
        to={_path}
        _hover={{
          textDecoration: "none",
        }}
      >
        {text}
      </Link>
      {isActive && (
        <Divider
          borderColor="brand.500"
          border="2px solid"
          top="47px"
          position="absolute"
          left={0}
        />
      )}
    </Stack>
  )
}

const renderPill = (pill: NavPillProps) => <NavPill key={pill.path} {...pill} />

export default SubNavigationPills

```

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const formattedNu = Number(formatUnits(nu.balance))
+  const formattedT = Number(formatUnits(t.balance))
+
+  const totalAssetBalance = formattedKeep + formattedNu + formattedT
+
+  const progressBarValues = useMemo(() => {
+    return {
+      ["#7D00FF"]: {
```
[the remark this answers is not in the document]
georgeweiler: great point. The [{}, {}] interface is much nicer. I've updated

src/pages/Overview/index.tsx
georgeweiler: thanks @r-czajkowski you were right. Good catch.

src/utils/formatAmount.ts
```diff
@@ -0,0 +1,22 @@
...
+) => {
+  return formatNumeral(formatUnits(rawAmount, decimals), format)
+}
+
+export const formatFiatCurrencyAmount = (
+  amount: number | string,
+  format = "0,00",
+  currencySymbol = "$"
+) => {
+  return formatNumeral(formatUnits(amount), `${currencySymbol}${format}`)
```
r-czajkowski, line 21: Shouldn't be just an `amount` instead of `formatUnits(amount)`?

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <Body2>Possible T amount</Body2>
+      <HStack
+        bg={useColorModeValue("gray.50", "gray.700")}
+        mt={4}
+        px={6}
+        py={2}
+        borderRadius="md"
+      >
+        <Icon as={t.icon} boxSize="32px" />
+        <H3>{formatNumeral(conversionToTAmount)}</H3>
```
r-czajkowski, line 127: Here we should use `formatTokenAmount` or `TokenBalance` component because `conversionToTAmount ` is in  "raw" format.

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <CardTemplate title="WALLET">
+      <Body2 mb={2}>Liquid Tokens</Body2>
+      <MultiSegmentProgress values={progressBarValues} />
+      <Stack spacing={2} mt={4}>
+        {formattedT !== 0 && (
+          <BalanceStat balance={t.balance} icon={t.icon} text={t.text} />
+        )}
+        <BalanceStat
+          conversionRate={keep.conversionRate}
```
r-czajkowski, line 104: Let's use `useVendingMachineRatio` hook.

```
const keepRatio = useVendingMachineRatio(Token.Keep)
```

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,154 @@
...
+          balance={keep.balance}
+          icon={keep.icon}
+          text={keep.text}
+          tokenDecimals={keep.decimals}
+        />
+        <BalanceStat
+          conversionRate={8.26}
```
r-czajkowski, line 111: Same as above

```
const nuRatio = useVendingMachineRatio(Token.NU)
```

src/store/tokens/tokenSlice.ts
```diff
@@ -28,18 +30,27 @@ export const tokenSlice = createSlice({
     [Token.Keep]: {
       loading: false,
       balance: 0,
+      conversionRate: 4.87,
```
r-czajkowski, line 33: We can use `useVendingMachineRatio` hook in component so let's remove it.

src/store/tokens/tokenSlice.ts
```diff
@@ -28,18 +30,27 @@ export const tokenSlice = createSlice({
...
+      text: Token.Keep,
+      icon: Icon.KeepCircleBrand,
       usdConversion: 0,
       usdBalance: "0",
     },
     [Token.Nu]: {
       loading: false,
       balance: 0,
+      conversionRate: 2.66,
```
r-czajkowski, line 42: Same as above.

src/store/tokens/tokenSlice.ts
```diff
@@ -28,18 +30,27 @@ export const tokenSlice = createSlice({
...
+      text: Token.Nu,
+      icon: Icon.NuCircleBrand,
       usdConversion: 0,
       usdBalance: "0",
     },
     [Token.T]: {
       loading: false,
       balance: 0,
+      conversionRate: 1,
```
r-czajkowski, line 51: It's probably unnecessary 🤔

r-czajkowski (changes requested):
Overview page looks awesome. Great job George 🚀. Left some notes.

src/pages/Overview/Network/WalletBalances.tsx
```diff
@@ -0,0 +1,154 @@
...
+        to="/upgrade"
+        _hover={{ textDecoration: "none" }}
+      >
+        Upgrade Tokens
+      </Button>
+
+      <HStack justify="center" mt={4}>
+        <Link
+          color={useColorModeValue("brand.500", "white")}
+          textDecoration="underline"
```
r-czajkowski, line 144: Let's add href to the link and `rel="noopener noreferrer" target="_blank"` to open link in the new tab. Not sure where we should redirect users but this link https://github.com/threshold-network/solidity-contracts/blob/main/docs/rfc-2-vending-machine.adoc seems reasonable.

src/store/tokens/tokenSlice.ts
georgeweiler: what do you think about using the `useVendingmachineRatio` hook in the tokenContext to set these values in the store? That way we can keep the store as our source of truth for token info.

src/pages/Overview/Network/UpgradeBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+import vortex from "../../../static/images/vortex.png"
+import { Body2, H4 } from "../../../components/Typography"
+import useChakraBreakpoint from "../../../hooks/useChakraBreakpoint"
+import { Link as RouterLink } from "react-router-dom"
+
+const UpgradeBanner: FC = () => {
+  const { isOpen, onClose } = useDisclosure({ defaultIsOpen: true })
+
+  const heroText =
+    "Have KEEP or NuCircleBrand tokens? Upgrade them to T and harness the power of ThresholdCircleBrand."
```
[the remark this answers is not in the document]
georgeweiler: Whoops, this must have been a findAndReplace bug

src/pages/Overview/Network/WalletBalances.tsx
georgeweiler: see my comment about using that hook to set values in the store.

src/pages/Overview/Network/WalletBalances.tsx
georgeweiler: see my comment about using that hook to set values in the store.

src/store/tokens/tokenSlice.ts
georgeweiler: see my comment about using that hook to set values in the store.

src/store/tokens/tokenSlice.ts
georgeweiler: see my comment about using that hook to set values in the store.

src/store/tokens/tokenSlice.ts
r-czajkowski: hmm... tbh conversion rate makes only sense for KEEP and NU token and it's not something common for other tokens- eg. I've added TBTC token to context in #36 and this field it doesn't make sense for this token. Also, the vending machine defines ratio not token- actually token doesn't care about the exchange rate. So I'm not sure if we should do it this way. But up to you :) just left my thoughts.

r-czajkowski (approved):
LGTM! :shipit: Just small note. Could you please provide PR description and some screenshots how the overview page looks like?

Battenfield:
Merging - we can adjust the conversion rate after we ship, I'll ticket it out 🚀

PR #35, opened 2021-12-10
Erc20 approvals

georgeweiler:


michalinacienciala:
For the `Token Dashboard / CI / Build and test ` check to pass, the dependency to `threshold-network/solidity-contracts` should be updated (currently yarn.lock points to `0.0.2-dev.0` package, whereas we already have `0.0.2-dev.4` published).
I think the best, to avoid similar problems in the future, would be to modify the workflow config file and replace this step
```
      - name: Install dependencies
        run: yarn install --frozen-lockfile
```
with this:
```
      - name: Resolve latest contracts
        run: |
            yarn upgrade \
              @keep-network/keep-core \
              @threshold-network/solidity-contracts
```
This way the workflow will always use latest package matching the semver in the `package.json`.

r-czajkowski (approved):
The PR looks like a good basic concept so let's merge this PR!

PR #36, opened 2021-12-23
Display TVL on overview page

r-czajkowski:
~Depends on: #34~

This PR:
* adds support for [`Multicall`](https://github.com/makerdao/multicall) contract- thanks to this we can fetch eg. multiple token balances in one request. To test the integration with this contract you need to deploy contrarct to your local network- please see `multicall/README.md` file.
* displays the TVL on the `Overview` page (we need to figure out how to fetch tvl for PRE).
* updates the ropsten CI process- fetch the packages with `ropsten` tag.
* adds T dapp readme file.

<details>
  
  **TVL on Overview page**
  obraz
  
**Displaying T token USD balance**
obraz

</details>


.github/workflows/dashboard-ci.yaml
```diff
@@ -39,8 +39,11 @@ jobs:
       - name: Resolve latest contracts
         run: |
           yarn upgrade \
+            @threshold-network/solidity-contracts \
```
nkuba, line 42:
```suggestion
            @threshold-network/solidity-contracts \
```

.github/workflows/dashboard-ci.yaml
```diff
@@ -66,17 +69,18 @@ jobs:
...
-      #       @keep-network/keep-core@ropsten \
-      #       @threshold-network/solidity-contracts@ropsten
+      # TODO: Use `ropsten` tag for `threshold-network/solidity-contracts`
+      # pacakge to fetch latest ropsten contracts.
+      # `@threshold-network/solidity-contracts@ropsten`
 
       - name: Resolve latest ropsten contracts
         run: |
           yarn upgrade \
+            @threshold-network/solidity-contracts@0.0.2-ropsten.1 \
```
georgeweiler, line 79: is it okay to hardcode the version on this package?

README.md
```diff
@@ -1,70 +1,23 @@
-# Getting Started with Create React App
+# Threshold Token Dashboard
```
georgeweiler, line 1: thanks, been meaning to update this readme. @Battenfield Is there any additional info we want to add? Not in this PR necessarily, but just to put on your radar.

georgeweiler:
This is awesome. Great job, I like how we're pulling in a multi-call contract to fetch these values. One thought: should we create a new slice for the store to hold onto the TVL so we don't have to fetch in multiple places? We'll need it on the `<Portfolio />` and `<Overview />` pages, for example.

georgeweiler (approved):
I left a few questions and a thought about using redux, but this code looks great. Approving, but curious to hear your thoughts on my comments

.github/workflows/dashboard-ci.yaml
michalinacienciala: Yes and no, but mostly no ;) It was only introduced temporarily because we did not have the `ropsten`-tagged package yet published. I was going to change this once I published the package. But now the plan is to run the deployment flow soon (after https://github.com/keep-network/keep-ecdsa/pull/941 and https://github.com/keep-network/tbtc/pull/842 get merged) to get the `ropsten`-tagged packages also in other repositories (we need them for the current PR). So we now should switch to `@threshold-network/solidity-contracts@ropsten` and remove the comment.

r-czajkowski: Thanks Michalina for the explanation! I'll update.

r-czajkowski:
> This is awesome. Great job, I like how we're pulling in a multi-call contract to fetch these values. One thought: should we create a new slice for the store to hold onto the TVL so we don't have to fetch in multiple places? We'll need it on the `<Portfolio />` and `<Overview />` pages, for example.

I thought about it but still you have to fetch the data on the `<Portfolio />` and `<Overview />` even if we store this value in redux-store. IMO it makes sense if we'll create something like `TokenContext` to fetch data on the top of the component hierarchy and only once- but not sure if we should do that in this case 🤔. What do you think?

Battenfield:
> > This is awesome. Great job, I like how we're pulling in a multi-call contract to fetch these values. One thought: should we create a new slice for the store to hold onto the TVL so we don't have to fetch in multiple places? We'll need it on the `<Portfolio />` and `<Overview />` pages, for example.
> 
> I thought about it but still we have to fetch the data on the `<Portfolio />` and `<Overview />` even if we store this value in redux-store. IMO it makes sense if we'll create something like `TokenContext` to fetch data on the top of the component hierarchy and only once- but not sure if we should do that in this case 🤔. What do you think?

I'll create a ticket to revisit this, we can take on this after we ship 🚀

michalinacienciala:
@r-czajkowski, I think we should do in `dashboard-mainnet.yaml` the same changes as in `dashboard-ci.yaml` (unless there's something I'm not aware of)...

r-czajkowski:
> @r-czajkowski, I think we should do in `dashboard-mainnet.yaml` the same changes as in `dashboard-ci.yaml` (unless there's something I'm not aware of)...

Do you mean using the `mainnet` tag like [here](https://github.com/threshold-network/token-dashboard/blob/main/.github/workflows/dashboard-mainnet.yml#L30-L36)?

michalinacienciala:
> > @r-czajkowski, I think we should do in `dashboard-mainnet.yaml` the same changes as in `dashboard-ci.yaml` (unless there's something I'm not aware of)...
> 
> @michalinacienciala do you mean using the `mainnet` tag like [here](https://github.com/threshold-network/token-dashboard/blob/main/.github/workflows/dashboard-mainnet.yml#L30-L36)?

Oh, sorry, I just realized that we shouldn't do that. Ignore what I wrote before :)

.github/workflows/dashboard-ci.yaml
```diff
@@ -66,17 +69,14 @@ jobs:
...
-      # TODO: Use this step once we tag `threshold-network/solidity-contracts`
-      # pacakge with `ropsten` tag to fetch latest ropsten contracts.
-      # - name: Resolve latest ropsten contracts run: | yarn upgrade \
-      #       @keep-network/keep-core@ropsten \
-      #       @threshold-network/solidity-contracts@ropsten
-
       - name: Resolve latest ropsten contracts
         run: |
           yarn upgrade \
+            @threshold-network/solidity-contracts@0ropsten \
```
michalinacienciala, line 75:
```suggestion
            @threshold-network/solidity-contracts@ropsten \
```

georgeweiler (approved):

README.md
michalsmiarowski: @r-czajkowski Maybe we can add additional info to `Prerequisities` to point out that these project should be cloned to the same directory? That might not be so obvious since some of them are in `keep-network` and some of them are in `threshold-network` on github.

src/pages/Overview/index.tsx
```diff
@@ -6,20 +7,25 @@ import Network from "./Network"
...
+  },
+  { text: "tBTC", path: "/tBTC" },
+  { text: "PRE", path: "/pre" },
+]
 
+const Overview = ({}) => {
   const { path } = useRouteMatch()
-
   const isMobile = useChakraBreakpoint("md")
+  const [data, fetchtTvlData] = useFetchTvl()
```
michalsmiarowski, line 24: Typo

```suggestion
  const [data, fetchTvlData] = useFetchTvl()
```

michalsmiarowski (commented):
Overall very good job 💪 Left some non-blocking comments to look at in the future

PR #39, opened 2022-01-11
Staking flow

georgeweiler:
Adds a staking flow for T.

1. Run and seed a local ganache instance using the startup script
2. connect metamask to local RPC (may have to reset)
3. Update the operator address to an address that isn't one of the top 5
4. submit approval TX 
5. Stake your locally created T token


src/web3/hooks/useStakeTransaction.ts
```diff
@@ -0,0 +1,49 @@
...
+    onError
+  )
+
+  const stake = useCallback(
+    async (...args) => {
+      console.log("sending the transaction", args)
+      await sendTransaction(...args)
+    },
+    [sendTransaction, stakingContract?.address]
+  )
```
r-czajkowski, line 46: This hook should be responsible for checking if a `TStaking` contract can spend a given amount of T token. So we should use`useApproveTStaking`(as is defined as above) here. Eg.:
```
const { approve } = useApproveTStaking()

const stake = useCallback(({ amount, operator, beneficiary, authorizer }: StakeRequest ) => {
 await approve(amount)
 await await sendTransaction(operator, beneficiary, authorizer, amount)
}, [<deps here>])
```

What do you think? One advantage of this is that the component does not have to take care of the staking logic as we do in:
`/ConfirmStakingParams/index.ts`
```
const onSubmit = isApprovedForAmount
    ? () => stake(operator, beneficiary, authorizer, stakeAmount)
    : approveTStaking
```

src/web3/hooks/useTStakingAllowance.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from "react"
+import { useToken } from "../../hooks/useToken"
+import { Token } from "../../enums"
+import { useTStakingContract } from "./useTStakingContract"
+import { useWeb3React } from "@web3-react/core"
+
+export const useTStakingAllowance = () => {
```
r-czajkowski, line 7: Let's create something similar to the `useApproval` as I suggested above. Eg.:

```
const useTokenAllowance = (tokenContract, spender, account?: string) => {
 const { account: currentAccount } = useWeb3React()
 const _account = account || currentAccount
  const [allowance, setAllowance] = useState("0")

  useEffect(() => {
    const checkAllowance = async () => {
      setAllowance(
        await tokenContract?.allowance(_account, spender)
      )
    }

    if (token?.allowance && _account && spender) {
      checkAllowance()
    }
  }, [_account, spender, tokenContract])

  return { allowance }
}
```

then in `useTStakingAllowance`

```
const useTStakingAllowance = () => {
 const tTokenContract = useToken(Token.T)
 const tStakingContract = useTStakingContract()
 
 return useTokenAllowance(tTokenContract, tStakingContract?.address)
}
```

r-czajkowski (changes requested):
Left few thoughts. Overall looks very good! 🚀 I tested staking flow locally and works very well.

src/web3/hooks/useTStakingAllowance.ts
r-czajkowski: I would consider to return function from this hook instead of fetching data in `useEffect`- this value will be fetched each time the component is mounted but maybe it would be better if we return function that fetches the allowance amount.🤔

src/hooks/useReduxStaking.ts
```diff
@@ -0,0 +1,37 @@
+import { useSelector, useDispatch } from "react-redux"
```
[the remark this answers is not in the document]
georgeweiler: yeah I'm happy to rename. Including "Redux" in the  filename does feel a bit awkward now that you mention it. No problem

src/hooks/useReduxStaking.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const useReduxStaking: UseReduxStaking = () => {
+  const operator = useSelector((state: RootState) => state.staking.operator)
+  const beneficiary = useSelector(
+    (state: RootState) => state.staking.beneficiary
+  )
+  const authorizer = useSelector((state: RootState) => state.staking.authorizer)
+  const stakeAmount = useSelector(
+    (state: RootState) => state.staking.stakeAmount
+  )
```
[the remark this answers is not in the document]
georgeweiler: nice idea. I exported a more simple "udpateState" action too instead of hard coding each setter in the redux store

src/web3/hooks/useTStakingAllowance.ts
georgeweiler: I like the useEffect because we can run a check like 

```
if (tokenContract?.allowance && account && spender) {
      checkAllowance()
    }
```

in the hook to make sure the method won't fail. We don't have to repeat that logic in other places that use the allowance method

src/web3/hooks/useApproveTStaking.ts
```diff
@@ -0,0 +1,22 @@
...
+  const { contract: TTokenContract } = useToken(Token.T)
+  const tStakingContract = useTStakingContract()
+  const { sendTransaction, status } = useSendTransaction(
+    TTokenContract!,
+    "approve",
+    onSuccess
+  )
+
+  const approveTStaking = useCallback(async () => {
+    await sendTransaction(tStakingContract?.address, MaxUint256.toString())
```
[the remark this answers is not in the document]
Battenfield: Good point @r-czajkowski. I've created a feature ticket to take on in a future sprint so we don't block this PR.

src/web3/hooks/useTStakingAllowance.ts
r-czajkowski: Yeah, I like it too! But my point is that this value will be fetched when the component is mounted, not when the user triggers an `approve` transaction. so e.g. we use the `useApprove` hook in modal and the user opens this modal (the allowance amount is fetching), then closes and re-opens (the allowance amount is fetching again) etc. Ofc non-blocking comment- just flagging 🙂

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const {
+    stakingState: { stakeAmount, operator, beneficiary, authorizer },
+    updateState,
+  } = useStakingState()
+
+  const setStakeAmount = (value: string | number) =>
+    updateState("stakeAmount", value)
+  const setOperator = (value: string) => updateState("operator", value)
+  const setBeneficiary = (value: string) => updateState("beneficiary", value)
+  const setAuthorizer = (value: string) => updateState("authorizer", value)
```
r-czajkowski, line 41: Why do we need store this values in redux store? I think it should be ok if we store this values in component state.

src/components/Modal/StakingSuccessModal/StakingStats.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <Stack spacing="0.5rem">
+      {transactionInfo.map((info) => (
+        <HStack justify="space-between" key={info.text}>
+          <Body3 color="gray.500">{info.text}</Body3>
+          {info.value}
+        </HStack>
+      ))}
+    </Stack>
+  )
```
r-czajkowski, line 58: We already have similar component in `src/components/Modal/UpgradeToTModal/TransactionStats.tsx` so we should think about creating a reusable component.

src/components/Modal/StakingSuccessModal/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  transactionHash: string
+}
+
+const StakingSuccessModal: FC<StakingSuccessProps> = ({
+  transactionHash,
+  closeModal,
+}) => {
+  const {
+    stakingState: { stakeAmount, operator, beneficiary, authorizer },
+  } = useStakingState()
```
r-czajkowski, line 35: Same question as above. I think we can pass these values as a props. Especially if we display this modal in `useSubscribeToOperatorStakedEvent` hook- we could get these values from the `OperatorStaked` event.

src/components/Modal/StakingSuccessPreNeededModal/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <Alert status="warning" mb={4}>
+          <AlertIcon />
+          You haven't added a PRE Worker Adddress
+        </Alert>
+        <Box borderRadius="md" bg="gray.50" p={6} mb={8}>
+          <H5 mb={4} color="gray.800">
+            Your Stake was successful.
+          </H5>
+          <Body1 color="gray.700">
+            You'll need to run a PRE node to receive rewards. You can this this
```
r-czajkowski, line 40: double `this`.

src/components/StakingChecklist/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </Body3>
+      ),
+    },
+    {
+      key: 2,
+      title: "PRE Node Worker address",
+      subTitle: "Make sure you add your Worker address to gain rewards",
+    },
+    {
+      key: 3,
```
r-czajkowski, line 54: we can get these value from the `map` function:

```
{stakingChecklistItems.map((item, index) => (
        <ChecklistItem {...item} key={index} />
      ))}
```

src/hooks/useStakingState.ts
```diff
@@ -0,0 +1,20 @@
...
+  const stakes = useSelector((state: RootState) => state.staking.stakes)
+
+  const dispatch = useDispatch()
+
+  const updateState = (key: StakingStateKey, value: any) =>
+    dispatch(updateStateAction({ key, value }))
+
+  return {
+    stakes,
+    stakingState,
```
r-czajkowski, line 17: Why `stakingState` not `... stakingState `? In this way the staking state is under `stakingState` field : `const { stakingState: { operator, beneficiary, stakeAmount <other> } } = useStakingState()`

src/hooks/useStakingState.ts
```diff
@@ -0,0 +1,20 @@
...
+  const stakingState = useSelector((state: RootState) => state.staking)
+  const stakes = useSelector((state: RootState) => state.staking.stakes)
+
+  const dispatch = useDispatch()
+
+  const updateState = (key: StakingStateKey, value: any) =>
+    dispatch(updateStateAction({ key, value }))
+
+  return {
+    stakes,
```
r-czajkowski, line 16: If you agree with comment above we can skip this and this: `const stakes = useSelector((state: RootState) => state.staking.stakes)`. Eg:
```
const stakingState = useSelector((state: RootState) => state.staking)

const dispatch = useDispatch()

const updateState = (key: StakingStateKey, value: any) =>
 dispatch(updateStateAction({ key, value }))

return {
 ...stakingState,
  updateState
}
```

src/store/modal/modalSlice.ts
```diff
@@ -15,7 +15,13 @@ export const modalSlice = createSlice({
       state.modalType = null
       state.props = {}
     },
+    updateProps: (state, action) => {
+      state.props = {
+        ...state.props,
+        ...action.payload,
+      }
+    },
```
r-czajkowski, line 23: Do we really need this? As I can see we don't use it anywhere.

src/store/index.ts
```diff
@@ -11,8 +12,25 @@ const store = configureStore({
...
+        ignoredActions: ["modal/updateProps", "modal/openModal"],
+        // Ignore these field paths in all actions
+        ignoredPaths: [
+          "modal.props.setOperator",
+          "modal.props.setBeneficiary",
+          "modal.props.setAuthorizer",
+          "modal.props.onSubmit",
+          "modal.props.setAmountToStake",
+          "payload.props.setAmountToStake",
+        ],
```
r-czajkowski, line 31: Same as above.

package.json
```diff
@@ -116,6 +118,7 @@
     "eslint-plugin-react": "^7.25.1",
     "husky": ">=6",
     "lint-staged": ">=10",
-    "prettier": "^2.4.0"
+    "prettier": "^2.4.0",
+    "ts-essentials": "^9.1.2"
```
r-czajkowski, line 122: why do we need `ts-essentials`?

r-czajkowski (changes requested):

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+    setOperator,
+    beneficiary,
+    setBeneficiary,
+    authorizer,
+    setAuthorizer,
+  } = props
+
+  const { isOpen, onToggle } = useDisclosure({ defaultIsOpen: false })
+
+  return (
```
r-czajkowski, line 36: We should probably add validation to check if the text from input is ETH address.

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <Stack spacing={6} mb={6}>
+          <Box>
+            <TokenBalanceInput
+              label={`T Amount`}
+              amount={stakeAmount}
+              setAmount={setStakeAmount}
+              max={maxAmount}
+              icon={ThresholdCircleBrand}
+              mb={2}
+            />
```
r-czajkowski, line 86: We should probably add a validation to check if the amount is greater than `minTStake`.

r-czajkowski (changes requested):

src/components/Modal/ConfirmStakingParams/index.tsx
georgeweiler: It's because this state is used in several different modals. Redux is the easiest way to share the state between them. I tried to implement an "updateModalProps" function at first to avoid using redux but the state was very inconsistent and would not update in many situations.

src/components/StakingChecklist/index.tsx
georgeweiler: I've always heard that using the index is an anti pattern.

https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

what do you think?

r-czajkowski: I agree it's an anti-pattern, but I believe for dynamic values- like fetching data from API where we don't know how many values there will be. In this case, actually, we have a fixed number of items and they will be always in the same order. Don't have a strong opinion- up to you 🙂.

r-czajkowski (approved):
Looks very good, great job! :shipit: We will create tickets for non-blocking stuff and resolve them in follow-up work. Left final approval to @Battenfield.

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -19,6 +19,23 @@ interface AdvancedParamsFormProps {
...
+}: {
+  text: string
+  isInvalid: boolean
+}) => {
+  return (
+    <FormHelperText color={isInvalid ? "red.500" : "gray.500"}>
+      {isInvalid ? "Please enter a valid ETH address" : text}
+    </FormHelperText>
+  )
 }
```
r-czajkowski, line 39: Non-blocking: it should be as a separate component.

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -21,18 +21,31 @@ import { useStakingState } from "../../../hooks/useStakingState"
...
+
+  useEffect(() => {
+    const fetchMinTStake = async () => {
+      setMinTStake(await tStakingContract?.minTStake())
+    }
+
+    if (tStakingContract?.minTStake) {
+      fetchMinTStake()
+    }
+  }, [])
```
r-czajkowski, line 48: Non-blocking: maybe separate hook? or fetch this data only once and save in redux store?

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -63,6 +76,46 @@ const ConfirmStakingParams: FC<BaseModalProps> = () => {
...
   }, [account])
 
+  const isValidBeneficiary = isAddress(beneficiary)
+  const isValidAuthorizer = isAddress(authorizer)
+  const isValidOperator = isAddress(operator)
+  const isMoreThanMax = BigNumber.from(stakeAmount).gt(
+    BigNumber.from(maxAmount)
+  )
+  const isLessThanMin = BigNumber.from(stakeAmount).lt(minTStake)
+  const isZero = stakeAmount == 0
```
r-czajkowski, line 86: Non-blocking: We probably should use lib for forms eg. `formik` and create common validators.

Battenfield (approved):

PR #42, opened 2022-01-14
WIP:  Stakes table UI

georgeweiler:
Screen Shot 2022-01-13 at 10 31 50 PM


src/pages/Staking/StakesTable/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+              <Tr key={stake.operator}>
+                <StakeNameCell stake={stake} index={i} />
+                <StakeAddressesCell stake={stake} />
+                <StakeBalanceCell stake={stake} />
+                <StakeActionsCell stake={stake} />
+              </Tr>
+            )
+          })}
+        </Tbody>
+      </Table>
```
r-czajkowski, line 38: We should probably create a component wrapper for DataTable. something like:

```
<DataTable data={<data here>} >
 <Column header={"name"} fieldId={"operator"}>
  <render content here>
</Column>
<Column header={"addresses"} fieldId={"operator"}>
  <render content here>
</Column>
...
</DataTable>
```

And the `DataTable` should be responsible for mapping data into table cells etc. Ofc let's address it in a followup work but we can just leave a TODO.

src/pages/Staking/index.tsx
```diff
@@ -26,6 +27,8 @@ const StakingPage: FC = () => {
     fn()
   }, [fetchOwnerStakes, account, dispatch])
 
+  console.log("stakes", stakes)
```
r-czajkowski, line 30: Let's remove it.

src/utils/getUsdBalance.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const getUsdBalance = (
+  balance: string | number,
+  usdConversion: number
+): string => {
+  return numeral(
+    FixedNumber.fromString(usdConversion.toString())
+      .mulUnsafe(FixedNumber.fromString(formatUnits(balance)))
+      .toString()
+  ).format("$0,0.00")
```
r-czajkowski, line 13: Let's use `formatFiatCurrencyAmount` util function.

r-czajkowski (approved):
LGTM! just left minor comments

src/pages/Staking/StakesTable/index.tsx
georgeweiler: I made a quick ticket here to capture this: https://keepnetwork.atlassian.net/browse/DAPP-516

PR #43, opened 2022-01-19
Topup flow [depends on unstake flow]

georgeweiler:
depends on #44


src/hooks/useTokenBalance.ts
```diff
@@ -1,9 +1,19 @@
...
 export const useTokenBalance = (token: Token) => {
-  const _token = useToken(token)
+  const tokenMap = {
+    [Token.T]: "t",
+    [Token.Nu]: "nu",
+    [Token.Keep]: "keep",
+    [Token.TBTC]: "tbtc",
+  }
+
+  const tokenState = useTokenState()
```
georgeweiler, line 13: Hey @r-czajkowski let me know if this change makes sense to you. There's a small bug with our balances, especially when they are updated from a transfer event. 

https://keepnetwork.atlassian.net/browse/DAPP-515

This is an attempt to fix, as I think we should be getting token balances from redux, and not the context, right?  This change did not fix the bug. I think it is still needed though.

r-czajkowski: Actually, there is a bug in `useSubscribeToContractEvent` which we pushed in #32. I've opened a PR with a fix- please see #45.

georgeweiler: I think we still have an issue with the updater in that subscription hook. The current wallet balance is always `0`, so when an egress transfer is detected the value goes negative:

Screen Shot 2022-01-19 at 1 28 36 PM

r-czajkowski: As I can see `topup-flow` branch doesn't include changes from #45. I cherry-picked commits from #45 locally to this branch and it works. Please double-check.

PR #45, opened 2022-01-19
Fix `useSubscribeToContractEvent` hook

r-czajkowski:
This PR fixes the `useSubscribeToContractEvent`. We should call callback stored in `callbackRef` on contract event, otherwise, the hook will call obsolete callback and this causes unexpected results.


michalsmiarowski (approved):
Looking good to me

PR #48, opened 2022-01-20
Adds a gray info box component to share across the app

georgeweiler:


r-czajkowski (approved):
LGTM. I have refactored the `InfoBox` component because if we use `useModal` in that component, the styles of **all** `InfoBox` components will change, even if they are used outside of the `Modal` component. Please see f8f9647 and 13048c9.

PR #49, opened 2022-01-20
Staking card [depends on #47, #48]

georgeweiler:
Screen Shot 2022-01-20 at 5 33 54 PM


src/store/index.ts
```diff
@@ -19,9 +19,14 @@ const store = configureStore({
...
         // Ignore these action types
-        ignoredActions: ["modal/openModal", "staking/unstaked"],
+        ignoredActions: [
+          "modal/openModal",
+          "staking/unstaked",
+          "staking/updateStakeAmountForOperator",
+        ],
         // Ignore these field paths in all actions
         ignoredPaths: [
+          "staking.stakedBalance",
```
r-czajkowski, line 29: Let's keep `stakedBalance` as a string.

Battenfield (approved):
already approved

PR #52, opened 2022-01-21
Update "operator in use error" validation UI

georgeweiler:
Screen Shot 2022-01-20 at 7 41 52 PM

### DIFF
--- a/src/web3/hooks/useStakeTransaction.ts
+++ b/src/web3/hooks/useStakeTransaction.ts
@@ -28,10 +28,9 @@ export const useStakeTransaction = (
 
   const onError = (error: any) => {
     if (error?.data?.message.includes(CommonStakingErrors.OperatorInUse)) {
-      openModal(ModalType.TransactionFailed, {
-        error: new Error(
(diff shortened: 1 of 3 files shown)


src/web3/hooks/useStakeTransaction.ts
```diff
@@ -28,10 +28,9 @@ export const useStakeTransaction = (
 
   const onError = (error: any) => {
     if (error?.data?.message.includes(CommonStakingErrors.OperatorInUse)) {
-      openModal(ModalType.TransactionFailed, {
-        error: new Error(
-          "This operator is already in use. Please resubmit with a different operator address."
-        ),
+      // send the user back to the first staking step, but with validated form fields
+      openModal(ModalType.ConfirmStakingParams, {
+        operatorInUse: true,
```
r-czajkowski, line 33: I believe we should implement a kind of an async form validation w/o triggering the transaction. I mean when the user clicks submit button we should trigger async validation not transaction. With `formik` it's super easy https://formik.org/docs/guides/validation#validate. Ofc non-blocking.

r-czajkowski (approved):

PR #51, opened 2022-01-21
Update staking checklist to new designs

georgeweiler:
Screen Shot 2022-01-20 at 7 23 30 PM

PR #55, opened 2022-01-21
Updates external links [depends on #51]

georgeweiler:


src/pages/Overview/Network/StakingOverview.tsx
```diff
@@ -37,6 +38,8 @@ const StakingOverview: FC = () => {
         <Link
           color={useColorModeValue("brand.500", "white")}
           textDecoration="underline"
+          target="_blank"
+          href={ExternalLink.stakingContractLeanMore}
```
r-czajkowski, line 42: We need to create the `ExternaLink` component and it will render `target="_blank` and `rel="noreferrer noopener"` props. Also please see a nice [article](https://www.freecodecamp.org/news/how-to-use-html-to-open-link-in-new-tab/) explaining why we should add `rel` to the `a` tag.

src/enums/externalLink.ts
```diff
@@ -1,4 +1,9 @@
 export enum ExternalLink {
-  THRESHOLD_GITHUB = "https://github.com/threshold-network/token-dashboard",
-  THRESHOLD_DISCORD = "https://discord.gg/hHHZF4rA",
+  thresholdGithub = "https://github.com/threshold-network/token-dashboard",
+  thresholdDiscord = "https://discord.gg/hHHZF4rA",
+  metamaskHomePage = "https://metamask.io/",
+  stakingContractLeanMore = "",
+  preNodeSetup = "",
+  preStakingProvider = "",
+  exchangeRateLearnMore = " ",
```
r-czajkowski, line 8: Just flagging that we need to add these links.

r-czajkowski (dismissed):

Battenfield (approved):

PR #64, opened 2022-01-25
Changed head_ref to ref_name for testnet dashboard upload job

pdyraga:
This should let us have the same format of links on
preview.dashboard.test.threshold.network and
preview.dashboard.threshold.network.

We will always end with / and do not include index.html.

Why it works this way... we are still not sure.


pdyraga:
No, `head_ref` and `ref_name` are not the same in this context. This works differently than documented. Untada.

PR #66, opened 2022-01-25
Route to upgrade sub page based on token balance

georgeweiler:


src/hooks/useRouteByKeepNuBalance.ts
```diff
@@ -0,0 +1,37 @@
...
+      isFirstRender.current = false
+
+      if (keepBalanceBn.gt(nuBalanceBn)) {
+        safeNavigate("/upgrade/keep")
+      } else {
+        safeNavigate("/upgrade/nu")
+      }
+    }
+  }, [active, keepBalanceBn, nuBalanceBn])
+}
```
r-czajkowski, line 35: Something's wrong here:

**Scenario 1**
- I have KEEP and NU tokens. `KEEP balance > NU balance`
- go to KEEP to T page directly via the link `upgrade/keep`
- connect to a wallet
- dapp redirects me to the `upgrade/nu` even if I have more KEEP than NU.

**Scenario 2**
- I have KEEP and NU tokens. `KEEP balance < NU balance`
- go to KEEP to T page directly via the link `upgrade/keep`
- connect to a wallet
- dapp redirects me to the `upgrade/nu` but my intention was to upgrade KEEP to T no NU to T.

src/hooks/useRouteByKeepNuBalance.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (active && isFirstRender.current) {
+      isFirstRender.current = false
+
+      if (keepBalanceBn.gt(nuBalanceBn)) {
+        safeNavigate("/upgrade/keep")
+      } else {
+        safeNavigate("/upgrade/nu")
+      }
+    }
+  }, [active, keepBalanceBn, nuBalanceBn])
```
r-czajkowski, line 34: I believe we should include all necessary dependencies(`safeNavigate`) and they shouldn't be objects (`keepBalanceBn`, `nuBalanceBn`- we could convert to BN in `useEffect` and add `keep.balance` and `nu.balance` to dependency ) or remove dependency array because we have `isFirstRender` guard in `useEffect`.

r-czajkowski (changes requested):

src/hooks/useRouteByKeepNuBalance.ts
georgeweiler: @Battenfield Let's sync on these scenarios. #1 is a bug that I fixed in 47347ec3d3759a9c9c58aa0335b378b604f17c96

but scenario #2 feels like what you intended. We should think thru the several different experiences and nail down exactly what we want

Battenfield: Scenario 1 is a bug and needs to be fixed

Scenario 2 I'm ok with for the first release, we can ticket it out for a future UX consideration

@r-czajkowski after Scenario 1 is resolved and looks good to you, please approve and merge

r-czajkowski (commented):
Hook redirects after successful upgrading `KEEP/NU` to `T` because the balances changes.

**Scenario 1**
- I have `KEEP` and `NU` tokens`KEEP balance > NU balance`
- I upgraded `NU` to `T`
- dapp redirected me to the `upgrade/keep`

^ There is the same issue when I have `KEEP balance < NU balance` and want to upgrade from `KEEP` to `T`

**Scenario 2**
- I have only `KEEP`. `NU balance = 0`
- I upgraded all my`KEEP` tokens to `T`
- dapp redirected me to the `upgrade/nu`

I left final approval to @Battenfield

src/hooks/useRouteByKeepNuBalance.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  // do not route on initial render to prevent the page with a lesser balance from always re-routing to higher balance
+  const isFirstRender = useRef(true)
+
+  useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false
+    }
```
r-czajkowski, line 27: Looks like this flag is unnecessary because:
- `useEffect` will be triggered on first render,
- a condition in the first `if` statement is met- we set the `isFirstRender` to `false`,
- so in a second `if` statement  the  `!isFirstRender.current` condition always be `true`

Maybe I missed something but I tested locally and it looks like `if(active)` is enough.

r-czajkowski (approved):

georgeweiler:
closing, as Chris asked for #77 instead

PR #71, opened 2022-01-27
Revert "Update T dapp to include `TTokenStaking` contract changes "

georgeweiler:
Reverts threshold-network/token-dashboard#68


r-czajkowski:
We published the `@threshold-network/solidity-contracts` for ropsten with the latest changes in `TStaking` contract and the preview link now works as expected. Please see #72.

PR #73, opened 2022-01-27
Bump-up versions of keep-network and t-network dependencies

michalinacienciala:
Some of the semver ranges for keep-network and t-network dependencies
were pointing at old versions.


michalinacienciala:
We're getting a failure at the build stage after updating `tbtc` dependency. A method of pulling address in `useTBTCTokenContract.ts ` needs to be modified - tagging @r-czajkowski for that :)

r-czajkowski:
> We're getting a failure at the build stage after updating `tbtc` dependency. A method of pulling address in `useTBTCTokenContract.ts ` needs to be modified - tagging @r-czajkowski for that :)

I'll take a look at it. Thanks for a heads up @michalinacienciala.

package.json
```diff
@@ -12,9 +12,9 @@
     "@ethersproject/constants": "^5.5.0",
     "@ethersproject/hardware-wallets": "^5.5.0",
     "@keep-network/coverage-pools": ">1.1.0-dev <1.1.0-ropsten",
-    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-ropsten",
-    "@keep-network/keep-ecdsa": ">1.8.0-dev <1.8.0-ropsten",
-    "@keep-network/tbtc": ">1.1.2-pre <1.1.2-rc",
+    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
+    "@keep-network/keep-ecdsa": ">1.9.0-dev <1.9.0-ropsten",
+    "@keep-network/tbtc": ">1.1.2-dev <1.1.2-pre",
```
r-czajkowski, line 17: Shouldn't we use `-ropsten` instead of `-pre`? As I can see in npm registry there is a version with `-ropsten` suffix.

package.json
```diff
@@ -12,9 +12,9 @@
     "@ethersproject/constants": "^5.5.0",
     "@ethersproject/hardware-wallets": "^5.5.0",
     "@keep-network/coverage-pools": ">1.1.0-dev <1.1.0-ropsten",
-    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-ropsten",
-    "@keep-network/keep-ecdsa": ">1.8.0-dev <1.8.0-ropsten",
-    "@keep-network/tbtc": ">1.1.2-pre <1.1.2-rc",
+    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
```
r-czajkowski, line 15: Same as above.

michalinacienciala: There are packages like `1.8.0-pre.16` in the registry. I thought that package version gets resolved to latest alphabetically version matching the range (that's how it behaved for me in case of `tbtc` range). But now, after double checking I see that with the `>1.8.0-dev <1.8.0-ropsten` range, the version gets resolved to `1.8.0-dev.5` which is latest version chronologically. But I would still leave the range as I proposed, until we bump-up numbered version of the package.

georgeweiler (approved):

PR #77, opened 2022-01-31
Create hook for switching the upgrade page href

georgeweiler:
Replaces #66 as per different feature request from @Battenfield


r-czajkowski (approved):
LGTM! :shipit:

PR #84, opened 2022-02-03
Add white color to links in dark mode, prevent close modal on overlay

georgeweiler:


src/components/ViewInBlockExplorer.tsx
```diff
@@ -21,7 +21,7 @@ const ViewInBlockExplorer: FC<
   return (
     <Link
       textDecoration="underline"
-      color={useColorModeValue("brand.500", "brand.300")}
+      color={useColorModeValue("brand.500", "white")}
```
r-czajkowski, line 24: Let's use `ExternalLink` component here.

src/components/ExternalLink/index.tsx
```diff
@@ -19,7 +19,7 @@ const ExternalLink: FC<Props & LinkProps> = ({
       href={href}
       target="_blank"
       rel="noopener noreferrer"
-      color={useColorModeValue("brand.500", "brand.300")}
+      color={useColorModeValue("brand.500", "white")}
```
r-czajkowski, line 22: Shouldn't we change color for an arrow icon? https://github.com/threshold-network/token-dashboard/blob/main/src/components/ExternalLink/index.tsx#L31

src/components/ViewInBlockExplorer.tsx
r-czajkowski: [a5a4c90](https://github.com/threshold-network/token-dashboard/pull/84/commits/a5a4c90536deed20c2dc80db062bc9b01663fa0e)

r-czajkowski (approved):

PR #85, opened 2022-02-07
Display tvl w/o connected wallet

r-czajkowski:
WIP

PR #86, opened 2022-02-09
NuCypher tvl

r-czajkowski:
Depends on: #85 

WIP...

### DIFF
--- a/src/hooks/useFetchTvl.ts
+++ b/src/hooks/useFetchTvl.ts
@@ -8,6 +8,7 @@ import {
   useKeepAssetPoolContract,
   useTStakingContract,
   useKeepTokenStakingContract,
+  useContract,
 } from "../web3/hooks"
 import { useETHData } from "./useETHData"
 import { useToken } from "./useToken"
@@ -20,6 +21,8 @@ interface TVLRawData {
   keepCoveragePoolTVL: string
   keepStakingTVL: string
   tStakingTVL: string
+  stakedNU: string
+  ethInNuNetwork: string
   // TODO: add PRE and NU TVL
 }
 
@@ -38,7 +41,28 @@ const initialState = {
   keepCoveragePoolTVL: "0",
   keepStakingTVL: "0",
   tStakingTVL: "0",
+  stakedNU: "0",
+  ethInNuNetwork: "0",
 }
+
+// TODO: Get contract abi from the package and figure out how to use these
+// contracts on ropste and local network. This only works on mainnet.
+const ESCROW_ABI = [
+  {
+    inputs: [],
+    name: "currentPeriodSupply",
+    outputs: [{ internalType: "uint128", name: "", type: "uint128" }],
+    stateMutability: "view",
+    type: "function",
+  },
+]
+
+const ESCROW_ADDRESS = "0xbbD3C0C794F40c4f993B03F65343aCC6fcfCb2e2"
+const useNuStakingEscrowContract = () => {
+  return useContract(ESCROW_ADDRESS, ESCROW_ABI)
+}
+
+const WORK_LOCK_ADDRESS = "0xe9778E69a961e64d3cdBB34CF6778281d34667c2"
 
 export const useFetchTvl = (): [TVLData, () => Promise<TVLRawData>] => {
   const [rawData, setRawData] = useState<TVLRawData>(initialState)
@@ -48,12 +72,16 @@ export const useFetchTvl = (): [TVLData, () => Promise<TVLRawData>] => {
     keepCoveragePoolTVL,
     keepStakingTVL,
     tStakingTVL,
+    stakedNU,
+    ethInNuNetwork,
   } = rawData
 
   const eth = useETHData()
   const keep = useToken(Token.Keep)
   const tbtc = useToken(Token.TBTC)
   const t = useToken(Token.T)
+  const nu = useToken(Token.Nu)
+  const nuStakingEscrow = useNuStakingEscrowContract()
   const keepBonding = useKeepBondingContract()
   const multicall = useMulticallContract()
   const keepAssetPool = useKeepAssetPoolContract()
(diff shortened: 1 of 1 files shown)


src/hooks/useFetchTvl.ts
```diff
@@ -81,6 +109,24 @@ export const useFetchTvl = (): [TVLData, () => Promise<TVLRawData>] => {
...
+      method: "totalSupply",
+    },
+    {
+      contract: nuStakingEscrow!,
+      method: "currentPeriodSupply",
+    },
+    {
+      contract: nu.contract!,
+      method: "balanceOf",
+      args: [nuStakingEscrow?.address],
```
cygnusv, line 123: Can you clarify why we're tracking here the NU balance of staking escrow?

src/hooks/useFetchTvl.ts
```diff
@@ -93,14 +139,28 @@ export const useFetchTvl = (): [TVLData, () => Promise<TVLRawData>] => {
...
+      nuTotalSupply,
+      nuCurrenctPeriodSupply,
+      nuInEscrow,
+      ethInNuNetwork,
     ] = chainData.map((amount: string) => formatUnits(amount.toString()))
 
+    const stakedNU = FixedNumber.fromString(nuInEscrow)
+      .subUnsafe(
+        FixedNumber.fromString(nuTotalSupply).subUnsafe(
+          FixedNumber.fromString(nuCurrenctPeriodSupply)
```
cygnusv, line 151: This equation seems a bit complex and I'm not sure if this yields a correct figure. Depending on what's the goal, i think there are other endpoints that are simpler as input.

src/hooks/useFetchTvl.ts
r-czajkowski: TBH I implemented based on the https://github.com/nucypher/nucypher-monitor/blob/main/monitor/dashboard.py#L146-L163. Any tips on how to calculate NU TVL correctly would be appreciated 🙂

src/web3/hooks/useNuStakingEscrowContract.ts
```diff
@@ -0,0 +1,18 @@
...
+const ESCROW_ABI = [
+  {
+    inputs: [],
+    name: "currentPeriodSupply",
+    outputs: [{ internalType: "uint128", name: "", type: "uint128" }],
+    stateMutability: "view",
+    type: "function",
+  },
+]
+const ESCROW_ADDRESS = "0xbbD3C0C794F40c4f993B03F65343aCC6fcfCb2e2"
```
georgeweiler, line 14: What do you think about placing this in our `.env`? might be better so that we can just add ropsten & local address when they become available?

src/web3/hooks/useNuWorkLockContract.ts
```diff
@@ -0,0 +1,24 @@
+import { useContract } from "./useContract"
+
+// TODO: Get contract abi from the package and figure out how to use these
+// contracts on ropsten and local network. This only works on mainnet.
+const WORK_LOCK_ADDRESS = "0xe9778E69a961e64d3cdBB34CF6778281d34667c2"
```
georgeweiler, line 5: same as above

cygnusv:
> Would be great if we could have hardhat deployment scripts in https://github.com/nucypher/nucypher-contracts repo to deploy contracts locally and npm packages for mainnet and ropsten contracts.

We recently published an NPM package for this. Check out https://www.npmjs.com/package/@nucypher/nucypher-contracts, the artifacts directory contains the addresses and ABIs of some NU contracts. Let me know if you need something else there (perhaps WorkLock?)

r-czajkowski:
> > Would be great if we could have hardhat deployment scripts in https://github.com/nucypher/nucypher-contracts repo to deploy contracts locally and npm packages for mainnet and ropsten contracts.
> 
> We recently published an NPM package for this. Check out https://www.npmjs.com/package/@nucypher/nucypher-contracts, the artifacts directory contains the addresses and ABIs of some NU contracts. Let me know if you need something else there (perhaps WorkLock?)

Nice thanks! As I can see it only includes the mainnet addresses. Could we have a separate npm package for Ropsten? Something like we have in threshold contracts pacakge? https://www.npmjs.com/package/@threshold-network/solidity-contracts/v/1.1.0-ropsten.0. It will help and simplify the T dapp deployment process on Ropsten network.

src/web3/hooks/useNuStakingEscrowContract.ts
r-czajkowski: I think it's a nice idea 👍 , but hopefully, we can get artifacts for mainnet/ropsten/local network from the package like we do for other contracts. So let's hold for a while and we will see if we can get a separate package for ropsten network.

cygnusv:
> > > Would be great if we could have hardhat deployment scripts in https://github.com/nucypher/nucypher-contracts repo to deploy contracts locally and npm packages for mainnet and ropsten contracts.
> > 
> > 
> > We recently published an NPM package for this. Check out https://www.npmjs.com/package/@nucypher/nucypher-contracts, the artifacts directory contains the addresses and ABIs of some NU contracts. Let me know if you need something else there (perhaps WorkLock?)
> 
> Nice thanks! As I can see it only contains the mainnet addresses. Could we have a separate npm package for Ropsten? Something like we have in threshold contracts pacakge? https://www.npmjs.com/package/@threshold-network/solidity-contracts/v/1.1.0-ropsten.0. It will help and simplify the T dapp deployment process on Ropsten network.
> 
> Would be great if we could add `WorkLock` artifact to the package 🙂.

I'm afraid we don't have a Ropsten deployment of NuCypher contracts. At some point I deployed just the token to facilitate integration of the Vending Machines, but we can't redeploy the whole set of contracts. I think for Ropsten you'd have to mock it somehow. 

For mainnet, yes, we can add the WorkLock artifact to the package.

pdyraga:
> I'm afraid we don't have a Ropsten deployment of NuCypher contracts. At some point I deployed just the token to facilitate integration of the Vending Machines, but we can't redeploy the whole set of contracts. I think for Ropsten you'd have to mock it somehow.

Why can't we have them redeployed? It's just for Ropsten and they would exist under new `-ropsten` version, example here: https://www.npmjs.com/package/@threshold-network/solidity-contracts/v/1.1.0-ropsten.0
Alternatively, deploy just `WorkLock` and whatever is needed and append the new artifacts to the existing package.

PR #91, opened 2022-03-03
Legacy stakes

r-czajkowski:
Closes: #96 

This PR adds support for displaying the legacy KEEP/NU stakes and updates the `Staking` page and staking modals.

**Main changes:**
- displays stake data in a staking card instead of a data table,
- adds amount form to the `STAKED PORTFOLIO`card,
- adds `staking tvl` card to the `Staking` page to display network tvl,
- adds `rewards` card,
- removes amount form from the top-up and unstake modal,
- adds `formik` lib to handle form validation and submission.

**TODO:**
- [x] add support for unstaking legacy stakes- the `TTokenStaking` contract allows to unstake **only the total amount** of KEEP legacy stake but for **NU the unstake amount is required**(the user has to provide a given amount of NU tokens that wants to unstake).


src/components/CopyToClipboard/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+type CopyToClipboardProps = {
+  textToCopy: string
+  textCopiedMsg?: string
+  copyTextMsg?: string
+}
+
+const CopyToClipboard: FC<CopyToClipboardProps> = ({
+  textToCopy,
+  children,
+  textCopiedMsg = "Copied!",
```
georgeweiler, line 22: just a thought... a toast could be nice for this. Not sure if we have designs to follow.

https://chakra-ui.com/docs/feedback/toast

src/pages/Staking/StakeCard/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+          {isInActiveStake ? "inactive" : "active"}
+        </Badge>
+        <NotificationPill colorScheme="brand" mr="2" variant="gradient" />
+        <Label3 textTransform="uppercase" mr="auto">
+          stake - {stakeType}
+        </Label3>
+        <Switcher onClick={setFlag.toggle} isActive={isStakeAction} />
+      </Flex>
+      {/* <Body2 mt="12">Total Rewards</Body2>
+      <Flex mt="4" alignItems="center">
```
georgeweiler, line 75: Should we remove this commented code?

src/pages/Staking/StakeCard/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+      <Flex as="header" alignItems="center">
+        <Badge
+          colorScheme={isInActiveStake ? "gray" : "green"}
+          variant="subtle"
+          size="small"
+          mr="2"
+        >
+          {isInActiveStake ? "inactive" : "active"}
+        </Badge>
+        <NotificationPill colorScheme="brand" mr="2" variant="gradient" />
```
georgeweiler, line 68: I am getting a typescript error on this component because of the variant:

```
Type 'string' is not assignable to type 'undefined'.  TS2322

    66 |           {isInActiveStake ? "inactive" : "active"}
    67 |         </Badge>
  > 68 |         <NotificationPill colorScheme="brand" mr="2" variant="gradient" />
       |                                                      ^
    69 |         <Label3 textTransform="uppercase" mr="auto">
    70 |           stake - {stakeType}
    71 |         </Label3>
```

src/components/ExternalLink/SharedLinks.tsx
```diff
@@ -1,24 +1,28 @@
...
+export const StakingContractLearnMore: FC<StakingContractLearnMoreProps> = (
+  props
+) => {
   const tStakingContract = useTStakingContract()
+  const color = useColorModeValue("gray.500", "gray.300")
 
   if (tStakingContract?.address) {
     return (
-      <HStack justify="center" mt={4}>
+      <Body3 color={color} textAlign="center" {...props}>
```
georgeweiler, line 18: we dropped the `mt={4}` from the previous version. Let's add it back. Or maybe add top margin for each instance? up to you.
Screen Shot 2022-04-21 at 8 42 11 AM

src/components/Forms/SimpleTokenAmountForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+  shouldDisplayMaxAmountInLabel?: boolean
+}
+
+const SimpleTokenAmountFormBase: FC<
+  SimpleTokenAmountFormProps & FormikProps<FormValues>
+> = ({
+  submitButtonText,
+  maxTokenAmount,
+  helperText,
+  label = "Stake Amount",
```
georgeweiler, line 33: I think this label is too specific for the component. Could we use `Token Amount` instead? and pass `Stake Amount` where it's needed?

src/components/Forms/SimpleTokenAmountForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <Form onSubmit={formikProps.handleSubmit}>
+      <FormikTokenBalanceInput
+        name="tokenAmount"
+        label={
+          shouldDisplayMaxAmountInLabel ? (
+            <>
+              <Box as="span">{label} </Box>
+              <Box as="span" float="right">
+                {maxTokenAmount ? formatTokenAmount(maxTokenAmount) : "--"} T
```
georgeweiler, line 48: Is it okay to hard-code `T` as the token? Same for the placeholder below. It might be best to make this more generic so it can be used with Keep and Nu, or other tokens.

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -1,153 +1,88 @@
...
+}
+
+const AdvancedParamsFormBase: FC<ComponentProps & FormikProps<FormValues>> = ({
+  formId,
 }) => {
   return (
-    <FormHelperText color={isInvalid ? "red.500" : "gray.500"}>
-      {isInvalid ? errorText || "Please enter a valid ETH address" : text}
-    </FormHelperText>
+    <Form id={formId}>
```
georgeweiler, line 20: wow this is so much cleaner!  😍

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -1,153 +1,88 @@
...
+      <FormikInput
+        mt="6"
+        name="beneficiary"
+        label="Beneficiary Address"
+        helperText="This address will receive rewards."
+      />
+      <FormikInput
+        mt="6"
+        name="authorizer"
+        label="Authrozier Address"
```
georgeweiler, line 35: typo: `Authorizer`

src/utils/forms.ts
```diff
@@ -0,0 +1,70 @@
...
+type ValidationMsg = string | ((amount: string) => string)
+type ValidationOptions = {
+  greaterThanValidationMsg: ValidationMsg
+  lessThanValidationMsg: ValidationMsg
+  requiredMsg: string
+}
+const DEFAULT_MIN_VALUE = WeiPerEther.toString()
+
+const defaultLessThanMsg: ValidationMsg = (minAmount) => {
+  return `The value should be less than or equal ${formatTokenAmount(
```
georgeweiler, line 15: Is the logic for these strings backwards? I think we want to say 

- `value should be *more* than min`
- `value should be *less* than max

src/components/ChecklistItem/index.tsx
```diff
@@ -6,7 +6,7 @@ import {
   ListItem,
   useColorModeValue,
 } from "@chakra-ui/react"
-import { BsCheckCircleFill } from "react-icons/all"
+import { BsFillArrowRightCircleFill as Icon } from "react-icons/all"
```
georgeweiler, line 9: can you link the designs for this change? just curious

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -1,153 +1,88 @@
...
+        helperText="This authorizer will authorize applications."
```
georgeweiler, line 36: Did sasha provide this copy? If no, can we change to `This address will authorize applications.`?

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -1,153 +1,88 @@
...
 
-  const { isOpen, onToggle } = useDisclosure({
-    defaultIsOpen: stakingProviderInUse,
-  })
+    errors.stakingProvider = validateETHAddress(values.stakingProvider)
+    if (!errors.stakingProvider) {
+      let validationMsg: string | undefined = ""
+      try {
+        const { isProviderUsedForKeep, isProviderUsedForT } =
+          await checkIfProviderUsed(values.stakingProvider)
```
georgeweiler, line 66: we have no way to check for NU, right?

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -1,132 +1,83 @@
...
+        onOpen()
+      }
     }
-  }, [account])
-
-  const isValidBeneficiary = isAddress(beneficiary)
-
-  const isValidAuthorizer = isAddress(authorizer)
+    forceFormValidation()
+  })
```
georgeweiler, line 61: Are we sure we want this effect to run on every single render? I think we are missing some dependencies like `hasBeenValidatedOnMount`. Or perhaps using `[]` as the dependency array could achieve the same result rather than saving and updating the `hasBeenValidatedOnMount` value?

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -138,54 +89,55 @@ const ConfirmStakingParamsModal: FC<
             You are about to make a deposit into the T Staking Contract
           </H5>
           <Body1>
-            Staking is a two step action which requires approve and confirm
-            transactions.
+            Staking uses an ApproveAndCall function which requires one
```
georgeweiler, line 92: is this right?

src/components/Modal/StakingSuccessModal/index.tsx
```diff
@@ -36,22 +38,22 @@ const StakingChecklistModal: FC<
...
+              Go through Step 2 to make sure you get Rewards
+            </H5>
           </InfoBox>
-          <Alert status="warning">
-            <Body1>Complete Step 2 to start earning Rewards</Body1>
-          </Alert>
           <PreSetupSteps />
           {transactionHash && (
-            <Body3 mt="4rem" align="center">
+            <Body3 mt="2.5rem !important" align="center">
```
georgeweiler, line 56: no change required. I just want to say that I hate that the `<Stack>` component forces us to use `!important`. I really disagree with how they built it.

https://github.com/chakra-ui/chakra-ui/issues/1121#issuecomment-655352939

src/components/Modal/UnstakeTModal/Step1.tsx
```diff
@@ -0,0 +1,181 @@
...
+      unstakeType: UnstakeType.ALL,
+    })
+  }
+
+  const hasNuStake = stake.nuInTStake !== "0"
+  const hasKeepStake = stake.keepInTStake !== "0"
+  const hasLegacyStake = hasNuStake || hasKeepStake
+
+  const getLegacyTabTitle = (token: UpgredableToken) => {
+    return `Unstake legacy ${hasKeepStake && hasNuStake ? token : ""} stake`
```
georgeweiler, line 57: is this logic right? Don't we want the `hasLegacyStake` var?

src/components/Modal/UnstakeTModal/Step1.tsx
```diff
@@ -0,0 +1,181 @@
...
+          <TabList mb="8">
+            <Tab onClick={() => setUnstakeType(UnstakeType.NATIVE)}>
+              Unstake T
+            </Tab>
+            {hasKeepStake && (
+              <Tab onClick={() => setUnstakeType(UnstakeType.LEGACY_KEEP)}>
+                {getLegacyTabTitle(Token.Keep)}
+              </Tab>
+            )}
+            {hasNuStake && (
```
georgeweiler, line 112: Have you QA'd a wallet that has all 3 stakes? I don't have one but wondering what the tabs look like if there are 3 of them

src/components/Modal/UnstakeTModal/Step1.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  maxTokenAmount={stake.keepInTStake}
+                  icon={KeepCircleBrand}
+                  helperText="The legacy tokens can be only unstaked in full amount"
+                  isDisabled
+                />
+              </TabPanel>
+            )}
+            {hasNuStake && (
+              <TabPanel>
+                <SimpleTokenAmountForm
```
georgeweiler, line 144: flagging that the props passed to these legacy forms are quite different.  (no helper text for NU, disabled for keep, no `hasLegacyStake` for keep, etc.) Is that intentional? Shouldn't they have fairly similar user interfaces?

src/components/Modal/UnstakeTModal/Step1.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  const hasNuStake = stake.nuInTStake !== "0"
+  const hasKeepStake = stake.keepInTStake !== "0"
+  const hasLegacyStake = hasNuStake || hasKeepStake
+
+  const getLegacyTabTitle = (token: UpgredableToken) => {
+    return `Unstake legacy ${hasKeepStake && hasNuStake ? token : ""} stake`
+  }
+
+  const getUnstakeAllBtnHelperText = () => {
```
georgeweiler, line 60: I think this is a good usecase for `useMemo` instead of a function that is invoked on every render in the JSX:

```
const getUnstakeAllBtnHelperText = useMemo(() => {
  const suffix =
      hasNuStake && hasKeepStake ? "KEEP + NU" : hasKeepStake ? "KEEP" : "NU"
    return `Unstakes max of both native tokens and legacy tokens (${suffix} + T)`
}, [hasNuStake, hasKeepStake])
```

src/web3/hooks/useUnstakeTransaction.ts
```diff
@@ -41,16 +49,21 @@ const useUnstakeTransaction = (
...
     onError
   )
 
   const unstake = useCallback(
     async ({ amount, stakingProvider }: UnstakeRequest) => {
-      await sendTransaction(stakingProvider, amount)
+      const args =
+        type === UnstakeType.NATIVE || type == UnstakeType.LEGACY_NU
+          ? [stakingProvider, amount]
+          : [stakingProvider]
```
georgeweiler, line 62: is this omitting the amount from KEEP unstakes? Is that because it's always a 100% unstake?

src/web3/hooks/useCheckDuplicateProviderAddress.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useKeepTokenStakingContract } from "./useKeepTokenStakingContract"
+import { useMulticallContract } from "./useMulticallContract"
+import { getMulticallContractCall, decodeMulticallResult } from "../utils"
+import { isAddressZero } from "../../web3/utils"
+
+const useCheckDuplicateProviderAddress = (): ((
+  stakingProvider: string
+) => Promise<{
+  isProviderUsedForKeep: boolean
+  isProviderUsedForT: boolean
```
georgeweiler, line 13: I might have already asked, but why can't we include NU here?

src/pages/Staking/RewardsCard.tsx
```diff
@@ -1,25 +1,30 @@
...
-            Rewards are released at the end of each month
-          </Body3>
-        </Box>
-      </Stack>
+      <Label3 textTransform="uppercase">Rewards</Label3>
+      <Body2 mt="6">Total Rewards</Body2>
+      <InfoBox mt="2">
+        <Body2>
+          Rewards are released at the end of each month and can be claimed
+          retroactively for March and April.
```
georgeweiler, line 15: This copy will become stale shortly. Wonder if we should revise or determine the month programmatically?

src/enums/externalHref.ts
```diff
@@ -6,4 +6,7 @@ export enum ExternalHref {
   preNodeSetup = "https://docs.nucypher.com/en/latest/pre_application/running_a_node.html#configure-and-run-a-pre-node",
   preStakingProvidersList = "https://docs.nucypher.com/en/latest/pre_application/node_providers.html",
   exchangeRateLearnMore = "https://blog.threshold.network/threshold-launch/",
+  keepDapp = "https://dashboard.keep.network/",
+  // TODO: Add nu dapp link
+  nuDapp = "URL needed",
```
georgeweiler, line 11: Flagging that this is still needed

src/pages/Staking/StakeCard/index.tsx
r-czajkowski: You have to run `yarn`. I added support for generating types for new components or new variants of components. Please see https://github.com/threshold-network/token-dashboard/pull/91/commits/d4a9374c52c954c9a384c9c74231b2afdbbc198f.

src/components/Forms/SimpleTokenAmountForm.tsx
r-czajkowski: Yeah great suggestion! I'll update!

src/components/ChecklistItem/index.tsx
r-czajkowski: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=2276%3A107185

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
r-czajkowski: I think nope. According to the source code of the staking contract, there are no limitations for NU https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/TokenStaking.sol#L276-L285.

src/components/Modal/ConfirmStakingParams/index.tsx
r-czajkowski: I decided to set up the guard `hasBeenValidatedOnMount` because we use `onOpen` function from `useDisclosure` in the `useEffect` hook so we can't (actually, we can but the react docs says:  if any variable is used inside the hook but is defined outside of it, then it should go in the dependency array.) use an empty dependency array. Inside `forceFormValidation` is a guard `if (hasBeenValidatedOnMount || !formRef.current) return` so the form validation will  be triggered only once.

src/components/Modal/ConfirmStakingParams/index.tsx
r-czajkowski: Good catch! Nope! This is a two step flow (`TTokenContract.approve` then `TStakingContract.stake`).

src/components/Modal/UnstakeTModal/Step1.tsx
r-czajkowski: I think is OK because we want to add the `token` prefix only if user has NU and KEEP stake. Otherwise we want to skip. eg:
- `T + KEEP` -> `Unstake T | Unstake legacy stake`,
- `T + NU` -> `Unstake T | Unstake legacy stake`,
- `T + KEEP + NU` -> `Unstake T | Unstake legacy KEEP stake | Unstake legacy NU stake`,
- `KEEP + NU` -> `Unstake legacy KEEP stake | Unstake legacy NU stake`.

src/components/Modal/UnstakeTModal/Step1.tsx
r-czajkowski: I just set `hasNuStake` to `true` here is a preview:

image

src/components/Modal/UnstakeTModal/Step1.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <TabPanel>
+                <SimpleTokenAmountForm
+                  onSubmitForm={onSubmitForm}
+                  initialTokenAmount={stake.keepInTStake}
+                  label="Unstake amount"
+                  submitButtonText="Unstake"
+                  maxTokenAmount={stake.keepInTStake}
+                  icon={KeepCircleBrand}
+                  helperText="The legacy tokens can be only unstaked in full amount"
+                  isDisabled
```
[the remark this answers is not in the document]
r-czajkowski: Because we can only unstake the all legacy KEEP stake, so user can't type the amount to unstake.

src/components/Modal/UnstakeTModal/Step1.tsx
r-czajkowski: Yes it is intentional. Regarding KEEP - we want to inform users that they can only unstake KEEP legacy stake in full amount. Regarding NU there are no helper text in design.

src/web3/hooks/useUnstakeTransaction.ts
r-czajkowski: Exactly, you can only unstake the legacy KEEP stake in full amount.

src/web3/hooks/useCheckDuplicateProviderAddress.ts
r-czajkowski: Please see https://github.com/threshold-network/token-dashboard/pull/91#discussion_r856130915.

src/pages/Staking/RewardsCard.tsx
r-czajkowski: Hm... we don't have the rewards contract yet and I was taking it as a placeholder 🤔

src/enums/externalHref.ts
r-czajkowski: [4dc8cbb](https://github.com/threshold-network/token-dashboard/pull/91/commits/4dc8cbb1351db51ddb32ce6bcbb94fbc8f2c03d7)

src/pages/Staking/StakeCard/index.tsx
r-czajkowski: [bb7c4fc](https://github.com/threshold-network/token-dashboard/pull/91/commits/bb7c4fca88a182cc1b65a9bd2c8b1df36d603a32)

src/components/ExternalLink/SharedLinks.tsx
r-czajkowski: [0fd1883](https://github.com/threshold-network/token-dashboard/pull/91/commits/0fd18835a8d8754a1eadd5da850c78210c6db32b)

src/components/Forms/SimpleTokenAmountForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+import SubmitTxButton from "../SubmitTxButton"
+import { Form } from "./Form"
+import { getErrorsObj, validateAmountInRange } from "../../utils/forms"
+import { formatTokenAmount } from "../../utils/formatAmount"
+
+type FormValues = {
+  tokenAmount: string
+}
+
+type SimpleTokenAmountFormProps = {
```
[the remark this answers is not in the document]
r-czajkowski: [37f91f8](https://github.com/threshold-network/token-dashboard/pull/91/commits/37f91f814c79dcc0cf5cebc2aefed2b96e1b2fca)

src/components/Forms/SimpleTokenAmountForm.tsx
r-czajkowski: [0b950d9](https://github.com/threshold-network/token-dashboard/pull/91/commits/0b950d9872b7ad1a750c2c5c150d6a0e0047427c)

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
r-czajkowski: [81ee74c](https://github.com/threshold-network/token-dashboard/pull/91/commits/81ee74c3cccd9674196093f0b2c43f1f8a9b07a3)

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
r-czajkowski: [cd29e38](https://github.com/threshold-network/token-dashboard/pull/91/commits/cd29e38b8837c2d5965b38501282c27a7c2d2fab)

src/components/Modal/ConfirmStakingParams/index.tsx
r-czajkowski: [5fccd5a](https://github.com/threshold-network/token-dashboard/pull/91/commits/5fccd5aa93a3cc08100a9420db84f3d7218154bf)

src/components/Modal/UnstakeTModal/Step1.tsx
r-czajkowski: [a9f5cb1](https://github.com/threshold-network/token-dashboard/pull/91/commits/a9f5cb17bd102647ab6a335320f6a26f9b6f1674)

src/components/CopyToClipboard/index.tsx
r-czajkowski: I think we don't have designs for this case at least I can't find them in Figma. I used a `Tooltip` component to indicate a copy action.

obraz

obraz

michalsmiarowski (changes requested):
Overall looking good. Left some comments to look at. I will test it locally now and give some more comments if needed.

src/pages/Staking/StakeCard/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+      <Flex mt="6" mb="8" alignItems="center">
+        <BoxLabel bg="brand.50" color="brand.700" mr="auto">
+          Provider address
+        </BoxLabel>
+        <CopyAddressToClipboard address={stake.stakingProvider} />
+      </Flex>
+      {isStakeAction || !hasLegacyStakes ? (
+        <TokenAmountForm
+          onSubmitForm={onSubmitForm}
+          label="Unstake Amount"
```
michalsmiarowski, line 115: Shouldn't this be dynamic base on the tab we choose? For `Stake` it should be `Stake Amount` and for `Unstake` it should be `Unstake Amount`.


image
_Stake tab is choosen on top, but Unstake amount is shown_

r-czajkowski: Fixed here fefff9d.

src/pages/Staking/index.tsx
```diff
@@ -1,25 +1,40 @@
...
   return (
     <PageLayout {...props}>
-      <Stack direction={{ base: "column", lg: "row" }} w="100%">
+      <SimpleGrid
+        columns={[1, null, null, 2]}
+        spacing="4"
+        w="100%"
+        mt="4"
+        alignItems="self-start"
+      >
```
michalsmiarowski, line 28: Non-blocking:

Something to discuss on desing/dev sync is maybe the positioning of the cards. I think the current position doesn't look very appealing to the user's eye in some cases, because of some unoccupied space. It is happening because legacy stake cards have different height than native stake cards. Example:


image

I think [Masonry Layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Masonry_Layout) would fit better here. Something like this:

image

src/components/Modal/TopupTModal/index.tsx
```diff
@@ -44,56 +38,42 @@ const TopupTModal: FC<
...
         <Stack spacing={6}>
           <InfoBox variant="modal">
-            <H5 mb={4}>You are about to top up your stake</H5>
+            <H5 mb={4} color={useColorModeValue("gray.800", "white")}>
+              You are about to top up your stake
+            </H5>
             <Body1>
               By topping up your stake you will add a new deposit of tokens to
-              your existing stake.
+              your initial stake.
```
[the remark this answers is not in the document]
r-czajkowski: I think this UI is not the latest iteration of legacy top-up flow. Here is the latest iteration for legacy top ups https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=2890%3A115788 and I think we should address it in a separate PR.

src/pages/Staking/StakeCard/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+          submitButtonText={submitButtonText}
+          maxTokenAmount={isStakeAction ? tBalance : stake.tStake}
+          shouldDisplayMaxAmountInLabel
+        />
+      ) : (
+        <Button onClick={onSubmitUnstakeBtn} isFullWidth>
+          {submitButtonText}
+        </Button>
```
michalsmiarowski, line 123: On Figma we also have `Expand View` button, but i'm not sure what it should do exactly. Are we planning to add this?

image

src/web3/hooks/useUnstakeTransaction.ts
```diff
@@ -14,7 +14,15 @@ enum CommonUnStakingErrors {
...
 
+const unstakeTypeToContractFunctionName: Record<UnstakeType, string> = {
+  [UnstakeType.NATIVE]: "unstakeT",
+  [UnstakeType.LEGACY_KEEP]: "unstakeKeep",
+  [UnstakeType.LEGACY_NU]: "unstakeNu",
+  [UnstakeType.ALL]: "unstakeAll",
+}
+
 const useUnstakeTransaction = (
+  type: UnstakeType,
```
michalsmiarowski, line 25: Non-blocking:

@sashatanase Maybe it could be a good idea to inform the user that he won't be able to un-stake for the first 24 hours after doing the stake, before he actually do the stake? Right now we inform him only after he tries to un-stake:

image

src/components/Modal/UnstakeTModal/Step2.tsx
```diff
@@ -0,0 +1,123 @@
...
+            beneficiary={stake.beneficiary}
+            authorizer={stake.authorizer}
+          />
+          <Alert status="warning">
+            <AlertIcon alignSelf="center" />
+
+            <AlertDescription color={useColorModeValue("gray.700", "gray.300")}>
+              Take note! If you fully unstake you will not be able to use the
+              same Operator Address for new stakes. This unstaked stake can be
+              toppped up anytime you want.
```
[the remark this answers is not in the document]
r-czajkowski: 5b70ffe

src/components/TokenBalanceInput/index.tsx
```diff
@@ -57,19 +64,30 @@ const TokenBalanceInput: FC<TokenBalanceInputProps> = ({
           placeholder="Enter an amount"
           paddingLeft="2.5rem"
           paddingRight="4.5rem"
-          value={amount ? formatUnits(amount) : undefined}
+          {...inputProps}
```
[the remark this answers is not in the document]
r-czajkowski: e9bc6b2

src/components/TokenBalanceInput/index.tsx
```diff
@@ -32,23 +37,25 @@ const TokenBalanceInput: FC<TokenBalanceInputProps> = ({
...
   }
 
   return (
-    <Box>
-      {label && (
-        <Body3 mb={2} fontWeight="bold">
-          {label}
-        </Body3>
-      )}
+    <FormControl isInvalid={hasError} isDisabled={inputProps.isDisabled}>
```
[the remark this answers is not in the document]
r-czajkowski: e9bc6b2

src/components/TokenBalanceInput/index.tsx
```diff
@@ -32,23 +37,25 @@ const TokenBalanceInput: FC<TokenBalanceInputProps> = ({
   amount,
   setAmount,
   label,
+  errorMsgText,
+  helperText,
+  hasError = false,
   ...inputProps
```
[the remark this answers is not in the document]
r-czajkowski: e9bc6b2

src/components/NumberInput/index.tsx
```diff
@@ -25,6 +25,7 @@ const NumberInput: FC<NumberInputProps> = (props) => {
       thousandSeparator
       decimalScale={props.decimalScale}
       __css={css}
+      disabled={props.isDisabled}
```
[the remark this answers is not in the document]
r-czajkowski: e9bc6b2

src/components/StakingStats/index.tsx
```diff
@@ -16,10 +17,11 @@ const StakingStats: FC<StakingStatsProps> = ({
...
   authorizer,
+  amountText = "Staked Amount",
 }) => {
   const transactionInfo: TransactionInfo[] = [
     {
-      text: "Stake Amount",
+      text: amountText,
       // todo: Token might not be a string, so this should be updated once we decide on the interface
       value: (
         <TokenBalance
```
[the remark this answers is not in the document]
r-czajkowski: 7771422

src/pages/Staking/RewardsCard.tsx
r-czajkowski: see https://github.com/threshold-network/token-dashboard/pull/107 we display here the total staking bonus reward.

michalsmiarowski (changes requested):
Some more comments to look at:

src/pages/Staking/index.tsx
r-czajkowski: @sashatanase

michalsmiarowski (approved):
LGTM! Tested it locally and it looks good!

Issue #98, opened 2022-04-13
Don't try to upgrade amounts below the conversion precision


cygnusv:
There's a recurring problem with Vending Machine users when they upgrade amounts with too many decimals. Vending machines have a hard-coded precision of 3 decimal places, and below that precision any passed amount is ignored for conversion purposes. The problem is that this leaves some "dust" tokens authorized to the vending machine contract, and this makes upgrade calls to fail later since some ERC20 implementations check for this (at least NuCypher token, see also the ERC20 reference to this respect https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md#approve).

The solution is very simple: don't pass amounts to convert below the third decimal position.
E.g., if the user has `1234.56789 NU` tokens, the UI will only show `1234.567 NU` and use `1234567000000000000000` as input amount to the vending machine, regardless if there's still `0.00089 NU` in the user address.

r-czajkowski:
Fixed in #99.

PR #99, opened 2022-04-14
Fix upgrade flow

r-czajkowski:
Closes: #98

Vending machines have a hard-coded precision of 3 decimal places,  and below that precision any passed amount is ignored for conversion purposes. The problem is that this leaves some "dust" tokens authorized to the vending machine contract, and this makes upgrade calls to fail later since some ERC20 implementations check for this eg. [the NuCypher token](https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/NuCypherToken.sol#L26-L30). To prevent this situation this PR limits the input in the upgrade flow to 3 decimals so a user can't pass the amount to convert below the third decimal position.

### DIFF
--- a/src/components/NumberInput/index.tsx
+++ b/src/components/NumberInput/index.tsx
@@ -10,9 +10,12 @@ export interface NumberInputValues {
   floatValue: number
 }
 
-const NumberInput: FC<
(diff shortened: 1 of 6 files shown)


michalsmiarowski (approved):
LGTM! Tested locally and it works as expected!

PR #105, opened 2022-05-02
Typography

r-czajkowski:
This PR updates typography components:
- sets the correct styles based on the [Threshold Design System](https://www.figma.com/file/zZi2fYDUjWEMPQJWAt8VWv/Threshold-DS?node-id=836%3A18299),
- adds `Inter` font,
- removes conditional rendering of different HTML tags for smaller and larger screens,
- removes unnecessary prop `textDecoration` for `Label` component- this component sets the correct style `textTransform` to `uppercase` by default.

### DIFF
--- a/src/components/Typography/index.tsx
+++ b/src/components/Typography/index.tsx
@@ -7,69 +7,66 @@ import {
   useColorModeValue,
 } from "@chakra-ui/react"
 
-export const Headline: FC<HeadingProps> = (props) => {
+export const H1: FC<HeadingProps> = (props) => {
   return (
     <Heading
-      fontSize="7xl"
-      color={useColorModeValue("gray.700", "gray.300")}
-      {...props}
-    />
-  )
-}
-
-export const H1: FC<TextProps> = (props) => {
-  return (
-    <Text
       as="h1"
-      fontSize="60px"
-      lineHeight="64px"
-      fontWeight="bold"
+      fontWeight="700"
+      fontSize={{ base: "48px", md: "60px" }}
+      lineHeight={{ base: "52px", md: "64px" }}
       color={useColorModeValue("gray.700", "gray.300")}
       {...props}
     />
   )
 }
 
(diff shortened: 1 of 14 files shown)


georgeweiler (dismissed):
thanks for doing this

src/components/Typography/index.tsx
```diff
@@ -143,10 +145,11 @@ export const Label2: FC<TextProps> = (props) => {
 export const Label3: FC<TextProps> = (props) => {
   return (
     <Text
-      fontWeight={600}
+      fontWeight="600"
       fontSize="14px"
       lineHeight="20px"
```
michalsmiarowski, line 150: Shouldn't `lineHeight` be equal to `16px` based on Figma?

src/components/Typography/index.tsx
```diff
@@ -7,69 +7,66 @@ import {
...
-    <Text
+    <Heading
       as="h3"
-      fontSize="36px"
-      lineHeight="40px"
+      fontWeight="600"
+      fontSize={{
+        base: "32px",
+        md: "36px",
+      }}
```
michalsmiarowski, line 44: Non-blocking:

No need to write this in multiple lines - we could keep it in one line as in other `Heading`s since it takse the same amount of space.

michalsmiarowski (changes requested):
Overall looking good! I think Line Height of the `Label3` should have different value based on Figma view, but I might be wrong so please look at the comments belo 😉

src/components/Typography/index.tsx
r-czajkowski: accf3f8

src/components/Typography/index.tsx
r-czajkowski: The eslint and prettier don't report any errors and the build passed. Maybe we can leave it as it is if you agree.

michalsmiarowski (approved):
LGTM

PR #107, opened 2022-05-09
Bonus rewards

r-czajkowski:
Closes: #109

This PR adds support for checking eligibility to staking bonus.

obraz

obraz

obraz

obraz

obraz

obraz


src/components/Modal/StakingBonus/EligibilityCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { StakeData } from "../../../types/staking"
+import { stakingBonus } from "../../../constants"
+
+export const EligibilityCard: FC<{ stake: StakeData }> = ({ stake }) => {
+  const isFirstRequirementMet =
+    stake.bonusEligibility.eligibleStakeAmount !== "0" &&
+    stake.bonusEligibility.hasActiveStake &&
+    !stake.bonusEligibility.hasUnstakeAfterBonusDeadline
+
+  const isSecondRequirementMet = stake.bonusEligibility.hasPREConfigured
```
michalsmiarowski, line 26: Based on [nucypher docs](https://docs.nucypher.com/en/latest/pre_application/running_a_node.html#qualify-node):

`
Nodes must be fully qualified: funded with ETH and bonded to an operator address, in order to fully start. Nodes that are launched before qualification will pause until they have a balance greater than 0 ETH, and are bonded to an Operator address.
`

Tthis means that to have a PRE node configured **and working** they need to have balance greater than 0 ETH. So maybe we should also check here if operator has more than 0 ETH and add it to second requirement?

src/components/StakingBonus/Title.tsx
```diff
@@ -0,0 +1,32 @@
+import { Box, useColorModeValue } from "@chakra-ui/react"
+import { H5 } from "../Typography"
+import { ReactComponent as BonusSwoosh } from "../../static/images/BonusSwoosh.svg"
+
+export const BonusTitle = () => {
+  // TODO: Update dark mode colors.
```
michalsmiarowski, line 6: Just pointing out this TODO here ☝️

src/hooks/useCheckBonusEligibility.ts
```diff
@@ -0,0 +1,249 @@
...
+        provider
+      )
+
+      const stakingProvidersInfo: BonusEligibilityResult = {}
+      for (const stakingProvider of stakingProviders) {
+        const stakingProviderAddress = getAddress(stakingProvider)
+
+        const hasPREConfigured =
+          stakingProviderToPREConfig[stakingProviderAddress]
+            ?.operatorConfirmedAt <= stakingBonus.BONUS_DEADLINE_TIMESTAMP
```
michalsmiarowski, line 83: Could also check here if operator has more than 0 ETH and change the name to `hasPREConfiguredAndRunning`.

See two comments above for more info.

michalsmiarowski (changes requested):
Left some comments to look at 👀

src/components/Modal/StakingBonus/EligibilityCard.tsx
r-czajkowski: Confirmed with David from NU that an `OperatorConfirmed` event is enough as a bonus requirement. cc @cygnusv.

src/hooks/useCheckBonusEligibility.ts
r-czajkowski: Please see https://github.com/threshold-network/token-dashboard/pull/107#discussion_r872050139

src/store/staking/stakingSlice.ts
```diff
@@ -62,6 +88,8 @@ export const stakingSlice = createSlice({
 
       state.stakes = [newStake, ...state.stakes]
       state.stakedBalance = calculateStakedBalance(state.stakes)
+      state.totalBonusBalance = calculateTotalBonusBalance(state.stakes)
+      state.totalRewardsBalance = calculateTotalRewardsBalance(state)
```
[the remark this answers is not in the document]
r-czajkowski: 6c34ea5f3d10072c329e53ced73e94cd7674309f

src/store/staking/stakingSlice.ts
```diff
@@ -99,6 +127,8 @@ export const stakingSlice = createSlice({
         .toString()
 
       state.stakedBalance = calculateStakedBalance(state.stakes)
+      state.totalBonusBalance = calculateTotalBonusBalance(state.stakes)
+      state.totalRewardsBalance = calculateTotalRewardsBalance(state)
```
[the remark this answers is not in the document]
r-czajkowski: 6c34ea5f3d10072c329e53ced73e94cd7674309f

src/store/staking/stakingSlice.ts
```diff
@@ -137,6 +167,8 @@ export const stakingSlice = createSlice({
         .sub(amount)
         .toString()
       state.stakedBalance = calculateStakedBalance(state.stakes)
+      state.totalBonusBalance = calculateTotalBonusBalance(state.stakes)
+      state.totalRewardsBalance = calculateTotalRewardsBalance(state)
```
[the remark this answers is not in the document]
r-czajkowski: 6c34ea5f3d10072c329e53ced73e94cd7674309f

src/store/staking/stakingSlice.ts
```diff
@@ -60,8 +89,19 @@ export const stakingSlice = createSlice({
...
+      const _isBeforeBonusDeadline = isBeforeBonusDeadline()
+      newStake.bonusEligibility = {
+        eligibleStakeAmount: _isBeforeBonusDeadline ? _amount : "0",
+        hasPREConfigured: false,
+        hasActiveStake: _isBeforeBonusDeadline,
+        hasUnstakeAfterBonusDeadline: false,
+        reward: calculateStakingBonusReward(_amount),
+      }
+
       state.stakes = [newStake, ...state.stakes]
```
michalsmiarowski, line 101: Wouldn't it be better to add this `newStake` at the end? I think the stakes are show in chronilogical order when fetched from the chain.

```suggestion
      state.stakes = [...state.stakes, newStake]
```

src/utils/stakingBonus.ts
```diff
@@ -0,0 +1,15 @@
...
+  FixedNumber.fromString(eligibleStakeAmount)
+    .mulUnsafe(
+      FixedNumber.fromString(stakingBonusConstants.STAKING_BONUS_MULTIPLIER)
+    )
+    .toString()
+    // Remove `.` to return an integer.
+    .split(".")[0]
+
+export const isBeforeBonusDeadline = (date: Date = new Date()) =>
+  dateToUnixTimestamp(date) <= stakingBonusConstants.BONUS_DEADLINE_TIMESTAMP
```
michalsmiarowski, line 15: Non-blocking:

I think this function could be named better since it checks if it's before **or** equal. I was thinking about `isBeforeOrEqualBonusDeadline` but I'm not sure. Maybe we should name it event more precise which would be `isCurrentDateBeforeOrEqualBonusDeadline`? What do you think?

michalsmiarowski (changes requested):
Left some more comments to look at

src/utils/stakingBonus.ts
r-czajkowski: The date doesn't always point to the current date. You can pass any date as a param. It sets the current date by default. So I think `isBeforeOrEqualBonusDeadline` is better.

src/store/staking/stakingSlice.ts
r-czajkowski: Tbh from my point of view is better to see the added record at the top of the list. So we should probably reverse the staked events array here https://github.com/threshold-network/token-dashboard/blob/main/src/hooks/useFetchOwnerStakes.ts#L28.

r-czajkowski: 6c1ec13

src/utils/stakingBonus.ts
r-czajkowski: 4ad0551

michalsmiarowski (approved):
LGTM

PR #111, opened 2022-05-14
Update staking bonus params

r-czajkowski:
The staking bonus will be 3% and the bonus deadline has been extended to
May 31 23:59:59 UTC.


src/types/staking.ts
```diff
@@ -33,10 +33,10 @@ export interface UseStakingState {
 export interface BonusEligibility {
   hasPREConfigured: boolean
   hasActiveStake: boolean
-  // No unstaking after the May 15th "snapshot" and until July 15th (not even
-  // partial amounts).
+  // No unstaking after the bonus deadline and until mid-July (not even partial
+  // amounts).
   hasUnstakeAfterBonusDeadline: boolean
-  // Only total staked amount before May 15th(May 15 2022 23:59:59) is taking
+  // Only total staked amount before bonus deadline is taking
```
michalsmiarowski, line 39: Non-blocking:

We could change this to a variable name so everyone knows exactly where that bonus deadline is, and where they can change it (f needed)

```suggestion
  // Only total staked amount before {BONUS_DEADLINE_TIMESTAMP} is taking
```
```suggestion
  // Only total staked amount before bonus deadline is taking
```

michalsmiarowski (approved):
LGTM

PR #113, opened 2022-05-18
Low pre funds display fix

michalsmiarowski:
Related to: https://github.com/threshold-network/solidity-contracts/pull/107

Fix displaying `low PRE funds` label.

Changes:
- display the label only when the PRE is not set up (and of course if the funds 
on PRE are low)
- change the unit of LOW_FUNDS_THRESHOLD constant to ETH


r-czajkowski (approved):
:shipit:

PR #117, opened 2022-05-31
Add How it Works page to the Staking page

r-czajkowski:
Add a page about how the staking works in T network according to the
Figma views.

### DIFF
--- a/src/components/StakingChecklist/index.tsx
+++ b/src/components/StakingChecklist/index.tsx
@@ -1,6 +1,6 @@
 import { FC } from "react"
 import { Box, Stack, useColorModeValue } from "@chakra-ui/react"
-import { Body3 } from "../Typography"
+import { Body2, Body3 } from "../Typography"
 import ChecklistGroup from "../ChecklistGroup"
 import { ExternalHref } from "../../enums"
 import ExternalLink from "../ExternalLink"
@@ -23,6 +23,36 @@ export const StakingDepositSteps: FC = () => {
                 withArrow
               />
             </Body3>
+          ),
+        },
+      ]}
+    />
+  )
+}
+
+export const LegacyStakesDepositSteps: FC = () => {
+  return (
+    <ChecklistGroup
+      title="Step 1 - T Staking Contract Authorization"
+      checklistItems={[
+        {
+          title: (
+            <Body2>
+              Authorize your NuCypher legacy stake{" "}
+              <ExternalLink text="here" href={ExternalHref.nuDapp} withArrow />
+            </Body2>
+          ),
+        },
+        {
+          title: (
+            <Body2>
+              Authorize your Keep Network legacy stake{" "}
+              <ExternalLink
+                text="here"
+                href={ExternalHref.keepDapp}
+                withArrow
+              />
+            </Body2>
           ),
         },
       ]}
--- a/src/components/StakingChecklist/index.tsx
+++ b/src/components/StakingChecklist/index.tsx
@@ -49,7 +49,7 @@ export const LegacyStakesDepositSteps: FC = () => {
               Authorize your Keep Network legacy stake{" "}
               <ExternalLink
                 text="here"
-                href={ExternalHref.keepDapp}
+                href={ExternalHref.keepDappAuthPage}
                 withArrow
               />
             </Body2>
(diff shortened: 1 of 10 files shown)


src/pages/Staking/HowItWorks/ProvidersCard.tsx
```diff
@@ -0,0 +1,105 @@
...
+      <Body1>{name}</Body1>
+      <Body3>{email}</Body3>
+    </Box>
+    <ExternalLink
+      fontWeight="600"
+      color="gray.700"
+      ml="auto !important"
+      text="Learn more"
+      // @ts-ignore
+      link={link as ExternalHref}
```
michalsmiarowski, line 46: `<ExternalLink>` doesn't have `link` prop. I think you meant `href`.

```suggestion
      href={link as ExternalHref}
```

src/components/StakingChecklist/index.tsx
```diff
@@ -30,6 +30,36 @@ export const StakingDepositSteps: FC = () => {
...
+        },
+        {
+          title: (
+            <Body2>
+              Authorize your Keep Network legacy stake{" "}
+              <ExternalLink
+                text="here"
+                href={ExternalHref.keepDapp}
+                withArrow
+              />
```
michalsmiarowski, line 54: Wouldn't it be better to link it to `Authorization -> Threshold` page in Keep Dashboard page instead of the homepage?

src/pages/Staking/HowItWorks/ProvidersCard.tsx
```diff
@@ -0,0 +1,105 @@
...
+      withArrow
```
michalsmiarowski, line 47: I think it's non-blocking but there is different icon on Figma:

image

src/pages/Staking/HowItWorks/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        tStakingContractAddress={tStakingContract?.address!}
+      />
+      <NewTStakesCard gridArea="new-t-stakes" />
+      <LegacyStakesCard
+        gridArea="legacy-stakes"
+        tStakingContractAddress={tStakingContract?.address!}
+      />
+      <StakingActionsCard gridArea="staking-actions" />
+      <AboutAddressesCard gridArea="addresses" alignSelf="flex-start" />
+      <ProvidersCard gridArea="providers" />
```
michalsmiarowski, line 48: Also non-blocking but nice to have:

On Figma there are `+` signs in top-right corner of every card which probably means, that they can be expanded and rolled-up. Non-blocking issue but leaving this comment for the future.

src/pages/Staking/HowItWorks/StakingActionsCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+    items: [
+      "Top-ups are done in T tokens.",
+      "If you want to top up your Legacy stake with Legacy tokens you have to go to the Legacy dashboard in order to do that.",
+    ],
+  },
+
+  {
+    sectionName: "Unstaking",
+    items: [
+      "Unstaking can be total or partial. For a total unstake you will not be able to use the same Operator Address for new stakes. This unstaked stake becomes an inactive stake and  can be toppped up anytime you want.",
```
michalsmiarowski, line 45: Typo:

```suggestion
      "Unstaking can be total or partial. For a total unstake you will not be able to use the same Operator Address for new stakes. This unstaked stake becomes an inactive stake and  can be topped up anytime you want.",
```

src/pages/Staking/HowItWorks/ProvidersCard.tsx
r-czajkowski: Can we address it in a separate PR? We can't change icon in `ExternalLink` with the current implementation.

src/pages/Staking/HowItWorks/ProvidersCard.tsx
r-czajkowski: 8b77f29

src/components/StakingChecklist/index.tsx
r-czajkowski: b31aede

michalsmiarowski (approved):
LGTM 🚀

PR #122, opened 2022-06-06
Fix `MAX` amount btn in `TokenBalanceInput` component

r-czajkowski:
If the user clicks `MAX` button in input, the component doesn't limit the
value to decimal scale and passes to a submit function the whole amount.
For example:
- the input in a form is limited to 3 decimals,
- a user has `2.123456789` T tokens,
- a user clicks `MAX` amount,
- a component passes the `2.123456789` value but should `2.123`.

This PR fixes this bug.

### DIFF
--- a/src/components/TokenBalanceInput/index.tsx
+++ b/src/components/TokenBalanceInput/index.tsx
@@ -13,10 +13,13 @@ import {
 } from "@chakra-ui/react"
 import { createIcon } from "@chakra-ui/icons"
 import { formatUnits, parseUnits } from "@ethersproject/units"
+import { Zero } from "@ethersproject/constants"
+import { BigNumber } from "@ethersproject/bignumber"
 import NumberInput, {
   NumberInputValues,
   NumberInputProps,
 } from "../NumberInput"
+import { web3 as web3Constants } from "../../constants"
 
 export interface TokenBalanceInputProps
   extends InputProps,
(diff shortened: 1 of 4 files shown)


src/components/TokenBalanceInput/index.tsx
```diff
@@ -59,7 +62,20 @@ const TokenBalanceInput: FC<TokenBalanceInputProps> = ({
   })
 
   const setToMax = () => {
-    _setAmount(formatUnits(max))
+    let remainder = Zero
+    const { decimalScale } = inputProps
```
georgeweiler, line 66: I can't see where we're passing the decimal scale? Is it defaulted somewhere that I'm missing?

r-czajkowski: eg here: https://github.com/threshold-network/token-dashboard/blob/main/src/components/UpgradeCard/index.tsx#L44-L46

The `decimalScale` prop comes from`NumberFormat` component from`react-number-format` lib and this component takes care of displaying formatted amount. In this component, we use the `decimalScale` prop to format the amount if a user clicks `MAX` btn. If the `decimalScale` is undefined we format the full amount in `setToMax` fn.

georgeweiler (approved):

dougvk:
thanks @r-czajkowski !

PR #123, opened 2022-06-08
How it works page updates

r-czajkowski:
Fix some styling issues on How it works page based on the feedback from https://coda.io/d/Building-Keep_d-fmEgBNFVH/06-06-2022-Staking-Cards-How-it-works-page_suvmH#_lutJM. Mainly changes related to
wrong spacing between elements and wrong font colors.

### DIFF
--- a/src/components/ExternalLink/index.tsx
+++ b/src/components/ExternalLink/index.tsx
@@ -1,4 +1,4 @@
-import { FC } from "react"
+import { FC, ReactElement } from "react"
 import { Icon, Link, LinkProps, useColorModeValue } from "@chakra-ui/react"
 import { FiArrowUpRight } from "react-icons/all"
 
@@ -6,12 +6,14 @@ interface Props {
   href: string
   text: string
   withArrow?: boolean
+  icon?: ReactElement
 }
 
 const ExternalLink: FC<Props & LinkProps> = ({
   text,
   href,
   withArrow,
+  icon,
   ...props
 }) => {
   const defaultColor = useColorModeValue("brand.500", "white")
@@ -27,9 +29,10 @@ const ExternalLink: FC<Props & LinkProps> = ({
       {...props}
     >
       {text}{" "}
-      {withArrow && (
+      {withArrow && !icon && (
         <Icon boxSize="12px" as={FiArrowUpRight} color={finalColor} />
       )}
+      {icon && icon}
     </Link>
   )
 }
--- a/src/pages/Staking/HowItWorks/AboutAddressesCard.tsx
+++ b/src/pages/Staking/HowItWorks/AboutAddressesCard.tsx
@@ -1,23 +1,23 @@
 import { FC, ComponentProps } from "react"
 import { List, ListItem } from "@chakra-ui/react"
 import Card from "../../../components/Card"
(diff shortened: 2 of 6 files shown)


src/components/ExternalLink/index.tsx
```diff
@@ -27,9 +29,10 @@ const ExternalLink: FC<Props & LinkProps> = ({
       {...props}
     >
       {text}{" "}
-      {withArrow && (
+      {withArrow && !icon && (
```
georgeweiler, line 32: nit: maybe we could simplify to 

```
{ withArrow ? <Icon boxSize="12px" as={FiArrowUpRight} color={finalColor} /> : icon }
```

georgeweiler (dismissed):

georgeweiler:
one nit, looks good 👍

georgeweiler:
@r-czajkowski  here's a related PR that might be nice to include: #119

src/components/ExternalLink/index.tsx
r-czajkowski: d039dd3

src/components/ExternalLink/index.tsx
```diff
@@ -27,8 +29,10 @@ const ExternalLink: FC<Props & LinkProps> = ({
       {...props}
     >
       {text}{" "}
```
michalsmiarowski, line 31: Can we move that `{" "}` space to the `withArrow` condition, and display it only when there is arrow or icon is not a null? Without arrow it makes that space a link also and it look strange.

image

See the space between `ThresholdStakingContract` and `supports`

michalsmiarowski: Actually `Learn more` also has unnecessary underlined space and it has an `icon`:

image

So maybe a better approach would be to just wrap the space in `span` and remove the underline? What do you think?

michalsmiarowski (changes requested):
Very good! Left some comments though 🚀

src/components/ExternalLink/index.tsx
```diff
@@ -1,17 +1,19 @@
-import { FC } from "react"
+import { FC, ReactElement } from "react"
 import { Icon, Link, LinkProps, useColorModeValue } from "@chakra-ui/react"
 import { FiArrowUpRight } from "react-icons/all"
 
 interface Props {
   href: string
   text: string
   withArrow?: boolean
+  icon?: ReactElement
```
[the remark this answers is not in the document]
r-czajkowski: I think currently is ok because we pass React Element not React Component. So if we passed the icon prop like this: `icon = {ExternalLinkIcon}` the `iconComponent` would fit better. Also, as I can see, the chakra has the same pattern please see https://chakra-ui.com/docs/components/form/button#button-with-icon.

src/components/ExternalLink/index.tsx
georgeweiler: one option could be to use a `<Button/>` component with `variant="link"` and the `rightIcon` prop. It would probably need some additional styling to match perfectly but maybe it's better than adding in an extra HTML element?

Screen Shot 2022-06-10 at 7 28 13 AM

r-czajkowski: 3ffdc66

PR #127, opened 2022-06-17
Legacy Top ups

r-czajkowski:
Depends on: https://github.com/threshold-network/solidity-contracts/pull/109/

This PR adds support for the legacy top-ups.

**Main changes:**
- adds a new top-up modal for legacy stakes,
- displays the top-up form in the Stake card only for native stakes; for legacy stakes we open a new modal ^,
- updates the `useTopupTransaction` hook to support the legacy top-ups,
- updates `useTConvertedAmount` hook- adds a function that converts a given amount of KEEP/NU token into T and returns it from a hook so we can use this function in case we can't pass the token amount into a hook directly,
- recognizes the top-up type based on the transaction data- the `ToppedUp` event doesn't include the `StakeType` so we need to recognize which function was called based on the transaction input data.

**How to top up using legacy stake (Ropsten/local development)**

1. Create a stake in the legacy Keep network- https://dashboard.test.keep.network/delegations/wallet.
2. Move created stake into T- https://dashboard.test.keep.network/applications/threshold.
3. Go to https://dashboard.test.keep.network/delegations/wallet top-up a given stake under the `Delegations` table and then commit a top-up under the `Top-ups` table. 
obraz
obraz
4. Go to https://dashboard.test.threshold.network/staking connect wallet as an owner and confirm a legacy top-up in stake card.
obraz

<details>
  Legacy stake:
  
obraz

obraz
obraz

Confirm legacy top-up:
obraz

Native Stake
obraz

</details>


src/components/Modal/TopupTModal/index.tsx
```diff
@@ -84,3 +89,4 @@ const TopupTModal: FC<
 }
 
 export default withBaseModal(TopupTModal)
+export { LegacyTopUpModal }
```
michalsmiarowski, line 92: Non-blocking:

Wouldn't it be cleaner to move the TopUpTModal to `Modal/TopupTModal/TopupTModal.jsx` and then just import `TopupTModal` and `LegacyTopUpModal` in `Modal/TopupTModal/index.jsx`?

src/hooks/useFetchOwnerStakes.ts
```diff
@@ -62,6 +74,29 @@ export const useFetchOwnerStakes = () => {
...
+        getMulticallContractCall
+      )
+
+      const [, eligibleKeepStakesData] = await multicallContract?.aggregate(
+        eligibleKeepStakeRequests
+      )
+      const eligibleKeepStakes = decodeMulticallResult(
+        eligibleKeepStakesData,
+        eligibleKeepStakeMulticalls
+      )
```
michalsmiarowski, line 94: Non-blocking:

We have a `useMulticall` hook that we don't really use that often. Couldn't we use it here?

src/components/Modal/TopupTModal/LegacyTopUpModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+  TopUpType,
+} from "../../../enums"
+import withBaseModal from "../withBaseModal"
+import { TokenAmountForm } from "../../Forms"
+import { useTokenBalance } from "../../../hooks/useTokenBalance"
+import { StakingContractLearnMore } from "../../ExternalLink"
+import { useModal } from "../../../hooks/useModal"
+import { StakeData } from "../../../types/staking"
+
+const stakeTypeToDaapHref: Record<StakeType.KEEP | StakeType.NU, ExternalHref> =
```
michalsmiarowski, line 39: Non-blocking:

Typo

```suggestion
const stakeTypeToDappHref: Record<StakeType.KEEP | StakeType.NU, ExternalHref> =
```

src/store/staking/stakingSlice.ts
```diff
@@ -112,40 +114,42 @@ export const stakingSlice = createSlice({
...
+        stakes[stakeIdxToUpdate].possibleNuTopUpInT = "0"
+      }
 
-      const amountUnstaked = BigNumber.from(amount)
+      const fieldName =
+        topUpType === TopUpType.NATIVE
+          ? "tStake"
+          : topUpType === TopUpType.LEGACY_KEEP
+          ? "keepInTStake"
+          : "nuInTStake"
```
michalsmiarowski, line 143: Non-blocking:

Would be nice if we could somehow retrieve the property name from the `state.object` (although I'm not sure how yet). This way when we change the name of the prop, for example `keepInTStake` to some other name the tslint will underline this as an error. In this case if we change the name and we forgot to change it here the tslint will not underline this as an error.

src/store/staking/stakingSlice.ts
```diff
@@ -112,40 +114,42 @@ export const stakingSlice = createSlice({
...
       const stake = stakes[stakeIdxToUpdate]
 
-      const originalStakeAmount = BigNumber.from(
-        stakes[stakeIdxToUpdate].tStake
-      )
+      if (topUpType === TopUpType.LEGACY_KEEP) {
+        stakes[stakeIdxToUpdate].possibleKeepTopUpInT = "0"
+      } else if (topUpType === TopUpType.LEGACY_NU) {
+        stakes[stakeIdxToUpdate].possibleNuTopUpInT = "0"
+      }
```
michalsmiarowski, line 136: Shouldn't we add something like this to the unstaked reducer also? I think user should not be able to have any possible top up after the unstake

michalsmiarowski (changes requested):
Left some minor comments - most of them are non-blocking. Only one need to be looked at before I can approve. Overall very good job 🔥

src/store/staking/stakingSlice.ts
r-czajkowski: I don't think so. A user can, for example, unstake an amount of T tokens but the top-up can still be possible.

src/store/staking/stakingSlice.ts
r-czajkowski: VSC code underlines this as an error see:
obraz

src/components/Modal/TopupTModal/LegacyTopUpModal.tsx
r-czajkowski: b3b450c

src/components/Modal/TopupTModal/index.tsx
r-czajkowski: f2dc6fd

src/components/Modal/TopupTModal/LegacyTopUpModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+  TabList,
+  Tab,
+  Tabs,
+  useColorModeValue,
+  Alert,
+  AlertDescription,
+  AlertIcon,
+  Link,
+  Divider,
+} from "@chakra-ui/react"
```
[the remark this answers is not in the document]
r-czajkowski: Updated in merge commit 0134b62.

michalsmiarowski (approved):
LGTM 🐦

PR #129, opened 2022-07-11
Change argument used to deploy keep-core

michalsmiarowski:
Change argument used to deplo `keep-core` in `start_dahsboard.sh` script.

Tha argument name was changed from `--contracts-only` to `--skip-client-build`.

Please see: https://github.com/keep-network/keep-core/pull/3026


r-czajkowski (approved):
:shipit:

PR #130, opened 2022-07-12
Update package json and fix dependencies that blocked compiling

georgeweiler:


georgeweiler:
Closing because we've discussed other options for updating our local env and configuring for gorli.

PR #131, opened 2022-07-12
Separates StakeCard components to prepare for the MAS card

georgeweiler:
The StakeCard  component consisted of one file with several smaller components. The complexity of the stake card will grow, so this PR separates those components into individual files.


r-czajkowski (approved):
👍

PR #132, opened 2022-07-13
Use `1.0.1-ropsten.0` package of cov-pools during deploy on Ropsten

michalinacienciala:
The `@keep-network/coverage-pools@1.0.1-ropsten.0` package that we want
to use for deployment on Ropsten is not tagged as `ropsten` (as there are
higher-versioned packages currently in the NPM registry). We need to
explicitly provide the version that we want to use.


michalinacienciala:
Closing this PR. The change wasn't meant to be merged to main. We're going to use the `ropsten`-tagged `coverage-pools` package.

PR #133, opened 2022-07-15
Deploy to GCP bucket after manual dispatch of workflow

michalinacienciala:
So far the step publishing the T dashboard to
`dashboard.test.threshold.network` GCP bucket was executed only when new
changes were being introduced to `main` branch.
There may be cases when we want to re-deploy T dashboard on Testnet
without pushing anything to `main`. This may happen when there's a
change in the dependecncies of the project that we want to see reflected
in the dashboard. To make such deploy possible, we're allowing for
execution of `Deploy to GCP bucket` step for 'workflow_dispatch' event.

### DIFF
--- a/.github/workflows/dashboard-ci.yaml
+++ b/.github/workflows/dashboard-ci.yaml
@@ -144,9 +144,12 @@ jobs:
               body: 'Preview uploaded to https://preview.dashboard.test.threshold.network/${{ github.head_ref }}/index.html.'
             })
 
-      # A push event is triggered on main branch merge; deploy to testnet bucket.
+      # A push event is triggered on main branch merge; deploy to testnet
+      # bucket. Also triggered by manual dispatch.
       - name: Deploy to GCP bucket
-        if: github.event_name == 'push'
+        if: |
+          github.event_name == 'push'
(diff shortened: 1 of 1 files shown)


.github/workflows/dashboard-ci.yaml
```diff
@@ -144,9 +144,12 @@ jobs:
...
             })
 
-      # A push event is triggered on main branch merge; deploy to testnet bucket.
+      # A push event is triggered on main branch merge; deploy to testnet
+      # bucket. Also triggered by manual dispatch.
       - name: Deploy to GCP bucket
-        if: github.event_name == 'push'
+        if: |
+          github.event_name == 'push'
+          || github.event_name == 'workflow_dispatch'
```
r-czajkowski, line 152: In that case, someone can build the testnet dapp from other branches than `main`. Maybe we could add a condition to check if the `wokfrlow_dispatch` was triggered from the `main` branch?

michalinacienciala: Good idea, added this in 533fc147c78bba3aba6ddc39433545f70bedf933.

r-czajkowski (approved):
LGTM!

PR #138, opened 2022-07-21
Add support for deployment on Goerli testnet

michalinacienciala:
Görli became a recommended test network after Ropsten's deprecation
notice (https://blog.ethereum.org/2022/06/21/testnet-deprecation/).
We're modifying our GitHub Actions workflows for deploying contracts
and dashboards to support the deployment on Görli. In this PR we're
modifying workflow for building/publishing T dashboard for testing
purposes. We also modify T dApp code which in couple places was
Ropsten-specific. Among some other changes, we have specified:
* merkle drop contract address for Goerli: `0x55F836777302CE096CC7770142a8262A2627E2e9`
* stub `SimplePreApplication` for Goerli: `0x829fdCDf6Be747FEA37518fBd83dF70EE371fCf2`
* multicall contract address for Goerli: `0x77dca2c955b15e9de4dbbcf1246b4b85b651e50e`

TODO:
- [x] deploy merkle drop contract address at Goerli network (@r-czajkowski): `0x55F836777302CE096CC7770142a8262A2627E2e9`
- [x] deploy stub `SimplePreApplication` on Goerli (@r-czajkowski): ``
- [x] switch Ropsten-related code in `token-dashboard` repo to Goerli


src/enums/web3.ts
```diff
@@ -1,6 +1,7 @@
 export enum ChainID {
   Ethereum = 1,
-  Ropsten = 3,
+  // Ropsten = 3,
```
michalinacienciala, line 3: I commented this out. Would it be better to remove it completely? (Same applies to other similar changes)

README.md
```diff
@@ -37,31 +37,31 @@ To make sure the changes made in local depository of `components` lib are implem
...
+## Install Görli contracts
 
 ```
-yarn upgrade @threshold-network/solidity-contracts@ropsten \
-  @keep-network/keep-core@ropsten \
-  @keep-network/keep-ecdsa@ropsten \
-  @keep-network/tbtc@ropsten \
-  @keep-network/coverage-pools@ropsten
+yarn upgrade @threshold-network/solidity-contracts@goerli \
+  @keep-network/keep-core@goerli \
```
r-czajkowski, line 61: Shouldn't we use `@keep-network/keep-core@1.8.1-goerli.0` as described in the workflow?

> We provide explicit version of the `keep-core` package, because using `goerli` tag results in `expected manifest` error - probably caused by bug in Yarn: https://github.com/yarnpkg/yarn/issues/4731.

package.json
```diff
@@ -12,10 +12,10 @@
...
     "@ethersproject/hardware-wallets": "^5.5.0",
     "@fontsource/inter": "^4.5.10",
-    "@keep-network/coverage-pools": ">1.1.0-dev <1.1.0-ropsten",
-    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
-    "@keep-network/keep-ecdsa": ">1.9.0-dev <1.9.0-ropsten",
-    "@keep-network/tbtc": ">1.1.2-dev <1.1.2-pre",
+    "@keep-network/coverage-pools": "development",
+    "@keep-network/keep-core": "development",
+    "@keep-network/keep-ecdsa": "development",
+    "@keep-network/tbtc": "development",
```
r-czajkowski, line 18: Why do we use `development`? Shouldn't we use a range?

package.json
```diff
@@ -25,7 +25,7 @@
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
     "@threshold-network/components": "^1.0.0-dev.0",
-    "@threshold-network/solidity-contracts": ">1.1.0-dev <1.1.0-ropsten",
+    "@threshold-network/solidity-contracts": "development",
```
r-czajkowski, line 28: Same as above.

src/enums/web3.ts
r-czajkowski: I think we can remove it completely. We can always use `git revert` if we want to restore the ropsten config.

r-czajkowski (changes requested):

package.json
michalinacienciala: Generally, we want the dependencies to be resolved to the latest packages meant for the development environment during `yarn install`. In the past we've been using ranges (which had a disadvantage of need of often updates - every time a version got bumped up), but since we started to tag the latest `-dev` packages with `development` tag, we can use that in the `package.json`. We use similar notation for example for `random-beacon`'s dependencies: https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/package.json#L38.

package.json
michalinacienciala: Same as above :)

src/enums/web3.ts
michalinacienciala: 78c0db2e6e804ff86b9b7ff9b74022e60d067fe9

README.md
michalinacienciala: You're right. I wrote README before realizing that the command does not work properly and forgot to update it later. Fixed in 59e2529426118ebe85ec635a338a5861adf8b43d.

package.json
r-czajkowski: Ahh I see, the `development` is the tag name. Nice 🚀.

r-czajkowski (approved):
LGTM! Tested on preview and it works! :shipit:

PR #139, opened 2022-07-27
Extracts the stake type to a util for reuse in other components

georgeweiler:
There are some new designs that require named stakes. This PR adds a util for generating the "stake type" that is used for naming them.


r-czajkowski (approved):
👍

PR #141, opened 2022-07-29
Adds the min-mid-max radio group for MAS

georgeweiler:
This selector/radio component will be used for the multi-app-staking feature. It includes a simple API with an `onRadioClick` function that accepts enum `MinMidMaxValue` to catch click events and update necessary parent state.

We can update tooltip copy at a later point.

Screen Shot 2022-07-29 at 9 42 04 AM
.

### DIFF
--- a/src/components/MinMidMaxRadio/index.tsx
+++ b/src/components/MinMidMaxRadio/index.tsx
@@ -1,0 +1,172 @@
+import { FC, useState } from "react"
+import { BsInfoCircleFill } from "react-icons/all"
+import {
+  BodySm,
+  BodyXs,
+  Box,
+  BoxProps,
+  Card,
+  HStack,
+  LabelMd,
+  Progress,
+  Flex,
+  Radio,
+  RadioGroup,
+  Stack,
+  StackDivider,
+  Tooltip,
+  Icon,
+} from "@threshold-network/components"
+
+export enum MinMidMaxValue {
+  Min = "MIN",
+  Mid = "MID",
+  Max = "MAX",
+}
+
+export interface MinMidMaxRadioProps {
+  label: string
+  progressValue: number
+  badgeColor: string
+  badgeBg: string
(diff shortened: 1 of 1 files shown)


src/components/MinMidMaxRadio/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    </Card>
+  )
+}
+
+const MinMidMaxRadioGroup: FC<{
+  onRadioClick: (value: MinMidMaxValue) => any
+}> = ({ onRadioClick }) => {
+  const [value, setValue] = useState<MinMidMaxValue | undefined>(undefined)
+
+  const _onRadioClick = (value: MinMidMaxValue) => {
```
michalsmiarowski, line 118: Can we change the name of the prop to `val`, because right now it has the same name as the `value` from `useState`. Would prefer to not mix that.

```suggestion
  const _onRadioClick = (val: MinMidMaxValue) => {
```

src/components/MinMidMaxRadio/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+      onClick={() => setValue(value)}
+      borderColor={isActive ? "brand.300" : undefined}
+      boxShadow="none"
+    >
+      <HStack
+        divider={<StackDivider />}
+        justifyContent="space-between"
+        w="100%"
+      >
+        <Radio size="lg" colorScheme="brand" defaultChecked value={value}>
```
michalsmiarowski, line 79: Why we use `defaultChecked` prop here? Is it really needed?

michalsmiarowski (changes requested):
Left some comments to look at

sashatanase:
@georgeweiler we will change this component so please do not focus on it.

michalsmiarowski:
@georgeweiler This component did not make it to the final version, so I think we should close this PR

PR #149, opened 2022-08-09
Update merkle distribution JSON file

r-czajkowski:
We've detected an accounting bug with the merkle distribution amounts so
here we update the merkle distribution JSON file.

Additionally, I set the correct deployment block([`15146501`](https://etherscan.io/tx/0x902eae448671b8f0eeb0d59bf796c55dac5ae4988fe8b71da1645d9d014583d6)) of the merkle drop
contract. We should've updated this value when deploying v1.2.0.


michalsmiarowski (approved):
LGTM

PR #151, opened 2022-08-10
Update dashboard-ci.yaml with tbtc-v2 packages

michalsmiarowski:
There are problems with installing tbtc-v2 and tbtc-v2.ts packages using
node v16.  Because of that we decided to change the version of node in
dashboard-ci.yaml to 14 and we also upgrade those two libs in 
`resolve latest contracts` and `resolve latest goerli contracts` steps.

We also change the node version to 14 in `dashboard-mainnet.yaml` file

More info: https://github.com/threshold-network/token-dashboard/pull/120#issuecomment-1210665779

### DIFF
--- a/.github/workflows/dashboard-ci.yaml
+++ b/.github/workflows/dashboard-ci.yaml
@@ -114,7 +114,7 @@ jobs:
             @keep-network/tbtc@goerli \
             @keep-network/coverage-pools@goerli \
             @keep-network/tbtc-v2@goerli \
-            @keep-network/tbtc-v2.ts@goerli
+            @keep-network/tbtc-v2.ts
 
       - name: Run postinstall script
         # `yarn upgrade` doesn't trigger the `postinstall` script.
--- a/.github/workflows/dashboard-ci.yaml
+++ b/.github/workflows/dashboard-ci.yaml
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: "16"
+          node-version: "14"
           cache: "yarn"
 
       # This step forces Git to download dependencies using `https://` protocol,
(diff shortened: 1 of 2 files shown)


.github/workflows/dashboard-ci.yaml
```diff
@@ -110,7 +112,9 @@ jobs:
             @keep-network/keep-core@1.8.1-goerli.0 \
             @keep-network/keep-ecdsa@goerli \
             @keep-network/tbtc@goerli \
-            @keep-network/coverage-pools@goerli
+            @keep-network/coverage-pools@goerli \
+            @keep-network/tbtc-v2@goearli \
```
michalinacienciala, line 116:
```suggestion
            @keep-network/tbtc-v2@goerli \
```

.github/workflows/dashboard-ci.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: "16"
+          node-version: "14"
```
michalinacienciala, line 21: We need to do the same change in L47 as well. And in the [dashboard-mainnet.yml](https://github.com/threshold-network/token-dashboard/blob/main/.github/workflows/dashboard-mainnet.yml).

michalinacienciala (changes requested):

.github/workflows/dashboard-ci.yaml
michalsmiarowski: Fixed in 3095c034a60a13ef9dbaf2a49e24b02499edc213 and 360f567f903c829c0eecad8344930a494ed0949e

.github/workflows/dashboard-ci.yaml
```diff
@@ -110,7 +112,9 @@ jobs:
             @keep-network/keep-core@1.8.1-goerli.0 \
             @keep-network/keep-ecdsa@goerli \
             @keep-network/tbtc@goerli \
-            @keep-network/coverage-pools@goerli
+            @keep-network/coverage-pools@goerli \
+            @keep-network/tbtc-v2@goearli \
+            @keep-network/tbtc-v2.ts@goerli
```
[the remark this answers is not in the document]
michalsmiarowski: Fixed in 685d01ecf98b81b27a04a65c1a7f0d10c57f4948

michalinacienciala (approved):
Looking good. The https://github.com/threshold-network/token-dashboard/runs/7768484090?check_suite_focus=true check is failing, but that's due to changes on `tBTC-template` branch, so shouldn't block the merge of #151.

PR #155, opened 2022-08-15
Remove staking bonus banner

michalsmiarowski:
The banner is not needed anymore so we are getting rid of it.

image


georgeweiler (dismissed):

r-czajkowski (approved):
LGTM! 👍

PR #159, opened 2022-08-17
Replace GitHub action with reusable workflow

nkuba:
Instead of using an action, we defined a reusable workflow.

A workflow feels like suits the use case better, it gives us a step-by-step view of the action execution, unlike the action that logs just one step (example: https://github.com/threshold-network/token-dashboard/runs/7882894088?check_suite_focus=true).
    
We replaced `testnet` environment with `protected`, as it may better reflect the reason for it.
    
References:
- https://docs.github.com/en/actions/using-workflows/reusing-workflows
- https://github.com/orgs/community/discussions/29827

Simulated using a protected environment with a manual approval in:
https://github.com/threshold-network/token-dashboard/actions/runs/2878285294

TODO:
- [ ] remove `testnet` environment once this PR is merged.


.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,169 @@
...
+          ETH_HOSTNAME_HTTP: ${{ secrets.ethUrlHttp }}
+          ETH_HOSTNAME_WS: ${{ secrets.ethUrlWS }}
+
+      - name: Deploy to GCP
+        uses: thesis/gcp-storage-bucket-action@v3.1.0
+        with:
+          service-key: ${{ secrets.gcpServiceKey }}
+          project: ${{ env.GOOGLE_PROJECT_ID }}
+          bucket-name: ${{ inputs.gcpBucketName }}
+          bucket-path: ${{ github.head_ref }}
```
michalinacienciala, line 153: Have you tested if deployment works ok when the reusable workflow gets executed by the parent job that got triggered by the `workflow_dispatch` or `push` events? Normally (for non-reusable workflows) the `github.head_ref` variable is available only when event triggering workflow is `pull_request`. For other triggers, the variable resolves to an empty value and the `bucket-path` input won't get passed to the runner, causing the mixup in the parameters in the `thesis/gcp-storage-bucket-action` action. I would expect to see similar behaviour here in the step executed via reusable workflow. If this doesn't work correctly, we can do a workaround similar to what we had in the code previously, as got introduce in this PR: https://github.com/threshold-network/token-dashboard/pull/160/.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,169 @@
...
+          Tag which should be used to pull latest non-legacy `threshold-network` and
+          `keep-network` packages with contracts (required if
+          `useUpstreamBuilds==false`). For example `dapp-dev-goerli`.
+        required: false
+        type: string
+      gcpBucketName:
+        description: The name of the bucket where the code will be deployed.
+        required: true
+        type: string
+      gcpBucketPath:
```
michalinacienciala, line 38: We define the input, but never use it.

michalinacienciala (changes requested):

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,169 @@
...
+            echo "::set-output name=threshold-contracts-version::${{ inputs.dependentPackagesTag }}"
+          fi
+
+      # Currently we only support `environment` = `goerli`. We provide explicit
+      # version of the `keep-core` package, because using `goerli` tag results in
+      # `expected manifest` error - probably caused by bug in Yarn:
+      # https://github.com/yarnpkg/yarn/issues/4731.
+
+      # TODO: Add upgrade of @keep-network/random-beacon, @keep-network/ecdsa,
+      # @keep-network/tbtc-v2 once they'll be added as dashboard's dependencies.
```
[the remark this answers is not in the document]
michalinacienciala: Also, there's an open PR that adds `random-beacon` and `ecdsa` dependency to the dashboard: #165. If it gets merged before #159 we'll need to include the new dependencies here as well.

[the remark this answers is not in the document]
michalsmiarowski: #165 was merged today @michalinacienciala

.github/workflows/build-and-publish.yml
michalinacienciala: We now use the input, see [eae00fb](https://github.com/threshold-network/token-dashboard/pull/159/commits/eae00fb73671914a1d690a761b14e4584c1d2a04).

.github/workflows/build-and-publish.yml
michalinacienciala: I confirmed that using `bucket-path: ${{ github.head_ref }}` does not work when parent workflow gets triggered via `workflow_dispatch` or `push` events. Addressed in [eae00fb](https://github.com/threshold-network/token-dashboard/pull/159/commits/eae00fb73671914a1d690a761b14e4584c1d2a04).

.github/workflows/build-and-publish.yml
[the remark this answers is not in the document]
michalinacienciala: I added the support for new dependencies when resolving merge conflicts (see [9f8c600](https://github.com/threshold-network/token-dashboard/pull/159/commits/9f8c600e270fee13ee9c6c2e6f60d0a68fc356db)).

michalinacienciala:
@nkuba, please review my additions. I verified the `workflow_dispatch` path in https://github.com/threshold-network/token-dashboard/actions/runs/3045394375/jobs/4907017216 - it failed because the `-wip` suffixed bucket that I used for testing was non-existing, but it can be seen in the logs that correct command was executed.

michalinacienciala (approved):
I'm marking the PR as approved, to unblock the merging. But before we merge let's have a :+1: from @nkuba on changes that I've added.

nkuba:
✅ LGTM

PR #160, opened 2022-08-18
Specify `bucket-path` when `github.head_ref` is not set

michalinacienciala:
In the workflow we're executing the `thesis/gcp-storage-bucket-action@v3.1.0`
action. One of its inputs is `bucket-path`, which is an optional
input and is used to configure bucket path.

If in the workflow step executing the action we don't use the
`bucket-path` property, the action will set its `bucket-path` input to
the default value which is `.`. This is the value that we want to use
when workflow is triggered manually or by merge to main.
Previously we thought that setting the action to use
`bucket-name: ${{ github.head_ref }}` property would do the trick of
using `.` for `workflow_dispatch` and `push` events and using
`head_ref` for `pull_request` events (because we thougt the default
value will be used when `bucket-name: ` (without a value) is used in
the action. But this turned out to not be truth. When we use
`bucket-name: ` the behavior is different than when we don't provide the
`bucket-name` property at all. In the first case we don't see the
`bucket-name` property in the runner's log, in the second case we see
the property populated with the default `.` value.
Not having `bucket-name` property in the runner results in wrong
behavior of the action (order of the inputs gets messed up and the
address of bucket gets resolved to gs://<BUCKETNAME>/<BUILDFOLDER>".

As a solution we add a `set-bucket-path` step which creates an output
which value is either `.` or `github.head_ref`, depending if the
latter is configured. We then use this output as value of `bucket-path`
parameter of action `thesis/gcp-storage-bucket-action@v3.1.0`.

Read more about how runner handles empty values here:
https://github.com/actions/runner/issues/924.

### DIFF
--- a/.github/actions/build-and-deploy-to-bucket/action.yml
+++ b/.github/actions/build-and-deploy-to-bucket/action.yml
@@ -117,13 +117,29 @@ runs:
         ETH_HOSTNAME_HTTP: ${{ inputs.ethUrlHttp }}
         ETH_HOSTNAME_WS: ${{ inputs.ethUrlWS }}
 
+    # We need this step to set correct value of `bucket-path` in the
+    # `cp-storage-bucket-action`. We can't use `${{ github.head_ref }}` directly
+    # there, because then when `head_ref` is empty, the parameter `bucket-path`
+    # is not forwarded to the runner and empty value is assigned to the
+    # `bucket-path` input instead of the default value (`.`). This results in
+    # incorrect passing of docker arguments and setting wrong bucket address.
+    - name: Configure `bucket-path`
+      shell: bash
+      id: set-bucket-path
+      run: |
+        if [ ${{ github.head_ref }} == '' ]; then
+          echo "::set-output name=bucket-path::."
+        else
+          echo "::set-output name=bucket-path::${{ github.head_ref }}"
+        fi
+
     - name: Deploy to GCP
       uses: thesis/gcp-storage-bucket-action@v3.1.0
       with:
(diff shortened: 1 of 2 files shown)


.github/actions/build-and-deploy-to-bucket/action.yml
```diff
@@ -117,13 +117,29 @@ runs:
...
+        fi
+
     - name: Deploy to GCP
       uses: thesis/gcp-storage-bucket-action@v3.1.0
       with:
         service-key: ${{ inputs.gcpServiceKey }}
         project: ${{ env.GOOGLE_PROJECT_ID }}
         bucket-name: ${{ inputs.gcpBucketName }}
-        bucket-path: ${{ github.head_ref }}
+        bucket-path: ${{ steps.set-bucket-path.outputs.bucket-path }}
```
nkuba, line 142: Another solution would be to define `gcpBucketPath` with the `default: .` and expect the value to be provided in the caller.

```suggestion
        bucket-path: ${{ inputs.gcpBucketPath }}
```

michalinacienciala: Yeah, that could work, as long as we make sure that the `gcpBucketPath` will not have the empty value. In our current use case (using action to build dashboard) it's not problematic. I'll prepare the change.

michalinacienciala: See fdcd76ff6901372ee6c4694d476297646614ab98. Unfortunately I had to split the `Build` step to two steps, depending on condition. Couldn't think of a cleaner solution.

nkuba (approved):

PR #162, opened 2022-08-18
New apps authorize modal

georgeweiler:
Creates the modal that will display for users who have not authorized any applications yet.

designs:
https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=4872%3A387918


src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </RadioGroup>
+      </ModalBody>
+      <ModalFooter>
+        <Button onClick={closeModal} variant="outline" mr={2}>
+          Dismiss
+        </Button>
+        <Button
+          onClick={routeToStake}
+          mr={2}
+          disabled={selectedAuthorizeAddress === ""}
```
michalsmiarowski, line 84: I think we can simplify it to just:

```suggestion
          disabled={!selectedAuthorizeAddress}
```

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const NewAppsToAuthorizeModal: FC<BaseModalProps> = ({ closeModal }) => {
+  const { stakes } = useStakingState()
+
+  const [selectedAuthorizeAddress, setSelectedAuthorizeAddress] = useState("")
+
+  const navigate = useNavigate()
+
+  const routeToStake = useCallback(() => {
+    if (selectedAuthorizeAddress.length > 0) {
```
michalsmiarowski, line 31: Same as above I'm pretty sure this can be simplified to:

```suggestion
    if (selectedAuthorizeAddress) {
```

src/theme/Radio.ts
```diff
@@ -0,0 +1,19 @@
...
+          backgroundColor: "brand.500",
+        },
+      },
+    },
+  }
+}
+
+export const Radio = {
+  baseStyle,
+}
```
michalsmiarowski, line 19: Non-blocking: 

I think this should probably be moved to the `components` repo, but maybe let's merge this like this first, and them move it to `components` repo in separate PRs

src/pages/Overview/Network/StakingOverview.tsx
```diff
@@ -26,6 +30,9 @@ const StakingOverview: FC = () => {
       <Button size="lg" isFullWidth mt={4} as={RouterLink} to="/staking">
         Go to Staking
       </Button>
+      <Button onClick={() => openModal(ModalType.NewAppsToAuthorize)}>
+        modal
+      </Button>
```
michalsmiarowski, line 35: Let's remember to remove that before the merge

michalsmiarowski (changes requested):
Looks good! 👍  Left some comments to look at before the merge

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    : undefined
+                }
+              >
+                <Radio value={stake.authorizer} size="lg">
+                  <LabelMd ml={4}>
+                    STAKE {i + 1} {getStakeType(stake)}
+                  </LabelMd>
+                </Radio>
+              </Card>
+            ))}
```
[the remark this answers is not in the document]
georgeweiler: 🙈

src/theme/Radio.ts
georgeweiler: why move to components? This feels like a pure theme file to me

michalsmiarowski: I think it suits the `components` repository better because it's a theme for a Radio component from chakra, and we export all chakra components form the `components` lib.

georgeweiler: Ah, yes I misread - thought you wanted to move to the `/src/components` directory. Yes, I agree this should be in the `@threshold-network/components` repo. I do wonder if it's a simple task to "extend" the theme for imports from chakra like this one 🤔

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+const NewAppsToAuthorizeModal: FC<BaseModalProps> = ({ closeModal }) => {
+  const { stakes } = useStakingState()
+
+  const [selectedAuthorizerAddress, setSelectedAuthorizerAddress] = useState("")
+
+  const navigate = useNavigate()
+
+  const routeToStake = useCallback(() => {
+    if (selectedAuthorizerAddress) {
+      navigate(`/authorize/${selectedAuthorizerAddress}`)
```
michalsmiarowski, line 33: Is this route correct? Shouldn't it be:

```suggestion
      navigate(`/staking/authorize/${selectedAuthorizerAddress}`)
```

michalsmiarowski (changes requested):
Left some more comments to look at 👀

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+            any time.
+          </BodyLg>
+        </InfoBox>
+        <BodyLg my={6}>Choose a stake to continue:</BodyLg>
+        <RadioGroup
+          onChange={setSelectedAuthorizerAddress}
+          value={selectedAuthorizerAddress}
+        >
+          <Stack>
+            {stakes.map((stake, i) => (
```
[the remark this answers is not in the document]
michalsmiarowski: I think this modal is displayed from the banner on the Overview page. We can probably hide the banner for not connected users but like you said, the modal itself should not care about when or why it's rendered 😃

georgeweiler:
Hey @michalsmiarowski  I just fixed up those small issues. Thanks for catching

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        </Button>
+        <Button
+          mr={2}
+          as={RouterLink}
+          to={`/staking/authorize/${selectedAuthorizerAddress}`}
+          onClick={closeModal}
+          disabled={!selectedAuthorizerAddress}
+        >
+          Continue
+        </Button>
```
michalsmiarowski, line 108: Since the button is now wrapped as `RouterLink` the additional bug showed up. When there are no stakes, or there aren't any selected stakes, the Continue button, even though it looks disabled, still works when you click it.

I found a workaround in [this issue](https://github.com/remix-run/react-router/issues/1082#issuecomment-160348163) but it seems little dirty. For our case it will look like this:

```suggestion
        <Button
          mr={2}
          as={RouterLink}
          to={`/staking/authorize/${selectedAuthorizerAddress}`}
          onClick={closeModal}
          disabled={!selectedAuthorizerAddress}
          style={{ pointerEvents: selectedAuthorizerAddress ? "auto" : "none" }}
        >
          Continue
        </Button>
```

Let me know what you think. Feel free to search for another solution because, like I said, this one is not pretty 😅

michalsmiarowski (changes requested):
@georgeweiler Found additional issue after the changes. See comment below

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
georgeweiler: great catch. I don't mind the css to disable it.

src/components/Modal/NewAppsToAuthorizeModal/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { useStakingState } from "../../../hooks/useStakingState"
+import { getStakeType } from "../../../utils/getStakeType"
+import { isSameETHAddress } from "../../../web3/utils"
+import { Alert, AlertDescription, AlertIcon, Link } from "@chakra-ui/react"
+import { isAddress } from "web3-utils"
+
+const NewAppsToAuthorizeModal: FC<BaseModalProps> = ({ closeModal }) => {
+  const { stakes } = useStakingState()
+
+  const [selectedAuthorizerAddress, setSelectedAuthorizerAddress] = useState("")
```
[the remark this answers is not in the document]
michalsmiarowski: That's actually a great catch! I think we should check `stakingProvider`, not `authorizer`, because when there are multiple stakes with same `authorizer` then when we select one of them, then all of them are being selected:

image

Stake 1 and Stake 3 have the same authorizer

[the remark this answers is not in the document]
michalsmiarowski: That's actually a great catch! I think we should check `stakingProvider`, not `authorizer`, because when there are multiple stakes with same `authorizer` then when we select one of them, then all of them are being selected:

image

Stake 1 and Stake 3 have the same authorizer

michalsmiarowski (approved):
LGTM

michalsmiarowski (approved):
LGTM 🚀

PR #163, opened 2022-08-19
Deauthorization Modal - Step 1

georgeweiler:
Adds the confirmation modal for deauthorizing an application

Screen Shot 2022-08-19 at 9 54 12 AM


src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
```diff
@@ -0,0 +1,119 @@
...
+          h="88px"
+          mx="auto"
+          my={8}
+        />
+        <LineDivider />
+        <Stack mb={8}>
+          <Flex justifyContent="space-between">
+            <BodySm color="gray.500">Decrease amount</BodySm>
+            <BodySm color="gray.700">
+              {numeral(decreaseAmount).format("0,0")} T
```
michalsmiarowski, line 67: Didn't test but maybe we can use `TokenAmount` component here?

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
```diff
@@ -0,0 +1,119 @@
+import { FC } from "react"
```
michalsmiarowski, line 1: (line above not related)

Let's try to avoid naming commits without any context like this one - 31ac478e49317519ca982396c4ec8b4bad66a9d3

Commit names should be generic, but associated with changes and if needed then additional, more detailed, description should be added where you can describe why you did such change in this way and add any additional important notes. This really helps when doing a review to understand the author's thinking faster and better 😄

michalsmiarowski (changes requested):
Looking good 👍 Left some comments to look at before the merge

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
```diff
@@ -0,0 +1,119 @@
...
+            This is 1 transaction
+          </FlowStep>
+          <FlowStep
+            preTitle="Step 2"
+            title="14 day cooldown"
+            status={FlowStepStatus.inactive}
+          >
+            You must wait a 14 day cooldown to then confirm the deauthorization.
+            This is 1 transaction.
+          </FlowStep>
```
michalsmiarowski, line 99: Flow step does not match the designs on Figma.

Figma:
image

Your version:
image

What's your version missing:
- no margin-bottom between "Step x" and the rest of the flow step
- "Step x" has different color
- it looks like the flow step title is too big? I think it uses `LabelSm` in the Figma version
- arrow icon next to `This is 1 transaction` is smaller than in Figma designs

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
```diff
@@ -0,0 +1,119 @@
...
+} from "../../StakingApplicationIcon"
+import numeral from "numeral"
+import shortenAddress from "../../../utils/shortenAddress"
+
+const InitiateDeauthorization: FC<BaseModalProps & { stake: StakeData }> = ({
+  closeModal,
+  stake,
+}) => {
+  const chosenStakingApplication = StakingApplication.TBTC
+  const decreaseAmount = 500000
```
michalsmiarowski, line 36: This should probably be passed to the modal

src/components/Modal/DeauthorizeApplicationModal/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+import InitiateDeauthorization from "./InititateDeauthorization"
+import withBaseModal from "../withBaseModal"
+import { BaseModalProps, StakeData } from "../../../types"
+import { useStakingState } from "../../../hooks/useStakingState"
+
+export interface DeauthorizeApplicationModalProps {
+  stake: StakeData
+}
+
+const DeauthorizeApplicationModal: FC<
```
michalsmiarowski, line 11: When opening modal there is an error in the console:

`react-remove-scroll-bar: cannot calculate scrollbar size because it is removed (overflow:hidden on body`

Not sure if it's related to this PR but can you check it out? I don't remember seeing it earlier.

michalsmiarowski (changes requested):
Did second part of the review after testing it out locally. Left some more comments, mostly related to UI.

src/components/StakingApplicationIcon/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+export enum StakingApplication {
+  TBTC = "TBTC",
+  RANDOM_BEACON = "RANDOM_BEACON",
+  PRE = "PRE",
+}
+
+const StakingApplicationIcon: FC<
+  {
+    stakingApplication: StakingApplication
+    increaseOrDecrease: "increase" | "decrease"
```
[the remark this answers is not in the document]
georgeweiler: yeah great idea. Was trying to come up with a better variable name myself. `operation` sounds great.

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
georgeweiler: see: https://github.com/threshold-network/components/pull/21

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
georgeweiler: What do you think about waiting until we have a better idea of the data interface and store? I think we might be able to pull this from a context or store depending on how we set it up. For now, this PR was just meant to build the UI components. Wdyt?

src/components/Modal/DeauthorizeApplicationModal/index.tsx
georgeweiler: I did some debugging. There's this issue with chakra: https://github.com/chakra-ui/chakra-ui/issues/6213


Looks like the modal is adding "overflow: hidden" to the body and complains when it tried to add it again?

in any case, it's only a warning in the development env: https://github.com/theKashey/react-remove-scroll-bar/blob/3dd80e28c92b8aac025e41f4258ff926cdc88af9/src/utils.ts#L22

So I think it's okay to ignore. WDYT?

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
michalsmiarowski: @georgeweiler I merged that PR and updated the `components` version to test it but the issue still persist when i test it

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
michalsmiarowski: Ok. Let's keep this conversation as unresolved, just in case. Will be easier to find this discussion if we decide to check all MAS-related PRs one day to see if we missed something.

src/components/Modal/DeauthorizeApplicationModal/index.tsx
michalsmiarowski: @georgeweiler 
Oh so it looks like this error occurs [when opening more than one modal](https://github.com/chakra-ui/chakra-ui/issues/6213). It's happening because we've got `DeauthorizeApplicationModal`, which is exported as `withBaseModal(DeauthorizeApplicationModal)` and it returns `InitiateDeauthorization` which is also exported as `withBaseModal(InitiateDeauthorization)`. So when we open `DeauthorizeApplicationModal` we have two modals opened (DeauthorizeApplicaitonModal and InitiateDeauthorization).

Is there a reason we separate those two like this?

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
georgeweiler: Screen Shot 2022-09-05 at 11 36 36 AM

interesting. my local components are updated. I'll try to reproduce

src/components/Modal/DeauthorizeApplicationModal/index.tsx
georgeweiler: Warning removed 👍

michalsmiarowski (approved):
🚀 🌔

michalsmiarowski (approved):
LGTM 🦉

PR #156, opened 2022-08-15
Multi app staking - authorize applications

michalsmiarowski:
Adds some new features mostly related to Multi App Staking:
- Applications section in Stake Card (UI without logic)
- Authorize Applications page (UI without logic)
- Multi app staking feature flag
- new `hideFromMenu` prop for the `RouteProps`
- (additionally) Stake Card code refactor (by @georgeweiler)

# Changes details

### Applications section in Stake Card (UI without logic)

Added `Applications` section in Stake Card. This section will allow user to authorize the apps (random beacon and tbt) and see which one is authorized and which one is not. Bear in mind that this is only UI - no logic is connected to that section yet.

Figma design: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=4782%3A264606

<details>
  <summary>Example</summary>

image

</details>

### Authorize Applications page (UI without logic)

Added the`Authorize Applications` page which can be viewed by clicking "Configure Apps" or "Authorize Application" buttons. This will allow user to see in details what apps are authorized and which one he wants to authorize. Bear in mind that this is only UI - no logic is connected to that section yet. The Authorize Applications component is fully responsive - you can see the examples below.

This page has it's own link - `staking/authorize/<authorizer_address>`

Figma design: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=4872%3A387073

Examples:

<details>
  <summary>Authorize applications page in normal view</summary>

image

</details>

<details>
  <summary>Authorize applications page ~800px width</summary>

image

</details>

<details>
  <summary>Authorize applications page ~600px width</summary>

image

</details>

<details>
  <summary>Authorize applications page ~400px width</summary>

Note: not sure if the checkbox should be on top in this case, but for now we can keep it like that

image

</details>

### Multi app staking feature flag

The feature flags were first introduced in https://github.com/threshold-network/token-dashboard/pull/120 (TBTC_V2 flag). This PR introduce new flag - MULTI_APP_STAKING. The default value is set as false so **all the ui changs made in this PR are hidden by default.**. The flags are set in `.env` file, but for local development we can just set:

```
export const MULTI_APP_STAKING = true
```

in `src/constants/featureFlags.ts`.

**Note**: If your App is running you have to stop it with ctrl+c and then run it again! This is very important, because otherwise routes might not work as expected and pages related to MULTI_APP_STAKING feature flag might not show.

[...]

PR #164, opened 2022-08-19
Adds MAS content to stake flow (pre txn submit)

georgeweiler:
Screen Shot 2022-08-19 at 11 02 45 AM

The multi-app staking feature includes some more content in the staking modal transaction flow. This PR adds the pre-submission content for informing the user about approving apps and running nodes


src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -76,16 +55,20 @@ const ConfirmStakingParamsModal: FC<
...
-      <ModalHeader>Stake Tokens</ModalHeader>
+      <ModalHeader display="flex" alignItems="baseline">
+        <H5 mr={2}>Stake Tokens</H5>
+        <BodySm>(Step 1)</BodySm>
+      </ModalHeader>
       <ModalCloseButton />
       <ModalBody>
         <InfoBox variant="modal">
           <H5>You are about to make a deposit into the T Staking Contract.</H5>
+          <BodyLg>Staking requires 2 transactions.</BodyLg>
```
michalsmiarowski, line 71: We should add some top padding to it:

image

It's too close to the Header in comparison with Figma designs

michalsmiarowski (changes requested):
Left some comments to look at:

src/components/Modal/SubmitStake/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <InfoBox variant="modal">
+          <H5>You are about to make a deposit into the T Staking Contract.</H5>
+          <BodyLg>Staking requires 2 transactions.</BodyLg>
+        </InfoBox>
+        <StakingStats
+          {...{
+            stakeAmount,
+            beneficiary: account as string,
+            stakingProvider: account as string,
+            authorizer: account as string,
```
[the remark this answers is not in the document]
georgeweiler: I've updated to use the correct values for `stakingProvider`, `beneficiary`, `authorizer`. Pulled them from state instead of passing as props.

src/components/StakingTimeline/index.tsx
```diff
@@ -109,17 +116,45 @@ export const PreSetupSteps: FC = () => {
...
+        changed at any time.
+      </FlowStep>
+      <FlowStep
+        size="sm"
+        title="Set up node"
+        preTitle="Step 3"
+        status={FlowStepStatus.inactive}
+      >
+        Set up and run a node for any of the applications authorized.
+      </FlowStep>
```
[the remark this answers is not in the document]
georgeweiler: Great catch, I've updated the component here: https://github.com/threshold-network/components/pull/21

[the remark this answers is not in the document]
georgeweiler: If you approve that PR, let's merge it & deploy, then I'll upgrade the components version in this PR so I can QA the new changes. (same for https://github.com/threshold-network/token-dashboard/pull/163#discussion_r956105418)

src/components/StakingTimeline/index.tsx
```diff
@@ -109,17 +116,45 @@ export const PreSetupSteps: FC = () => {
...
-      </Box>
+    <Stack spacing={6}>
+      <BoxLabel>Staking Timeline</BoxLabel>
+      <FlowStep
+        size="sm"
+        title="Stake Tokens"
+        preTitle="Step 1"
+        status={FlowStepStatus.active}
+      >
+        Enter the Provider, Beneficiary, and Authorizer addresses. These will be
```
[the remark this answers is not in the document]
liz-shinn: This is the Oxford comma; it can be omitted in a series but imo it should be kept there for clarity.

src/components/StakingTimeline/index.tsx
```diff
@@ -109,17 +116,45 @@ export const PreSetupSteps: FC = () => {
...
+        You can authorize 100% of your stake for each app. This amount can be
+        changed at any time.
+      </FlowStep>
+      <FlowStep
+        size="sm"
+        title="Set up node"
+        preTitle="Step 3"
+        status={FlowStepStatus.inactive}
+      >
+        Set up and run a node for any of the applications authorized.
```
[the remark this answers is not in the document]
liz-shinn: Yes, I think this makes more sense to switch the order here.

src/components/Modal/ConfirmStakingParams/index.tsx
michalsmiarowski: @georgeweiler  I think this one is still unresolved so I unresolved the conversation

src/pages/Staking/HowItWorks/NewTStakesCard.tsx
```diff
@@ -10,7 +10,7 @@ export const NewTStakesCard: FC<ComponentProps<typeof Card>> = (props) => {
         Before you start staking on Threshold Network, make sure you are aware
         of the following requirements:
       </BodyMd>
-      <StakingChecklist />
+      <StakingTimeline />
```
michalsmiarowski, line 13: We probably should use feature flag here. We've renamed `StakingChecklist` component to `StakingTimeline`, but maybe we should keep both components for now, until we remove feature flag (which means we deploy Multi App Staking to the production)? This way we can still show `StakingChecklist` when the MAS feature flag is set to false, and `StakingTimeline` when the MAS featuer flag is set to true.

For this we would need to have newest changes from `main` on this branch. The MAS feature flag was introduced in this PR - https://github.com/threshold-network/token-dashboard/pull/156.

Let's make sure if there are not any more situations like this one in this PR. When something is changes in a way that show some info about authorizing an app for the stake, we should wrap it in `featureFlag` to make sure it is not showed on `main` branch until the feature is fully released on production.

Thoughts?

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
```diff
@@ -35,6 +43,15 @@ const AdvancedParamsFormBase: FC<ComponentProps & FormikProps<FormValues>> = ({
         label="Authorizer Address"
         helperText="This address will authorize applications."
       />
+      {isSameETHAddress(authorizer, account as string) && (
```
michalsmiarowski, line 46: The Alert says `Authorizer address is different than your wallet address` but based on this consition, the alert is shown when the addresses are the same. I think it should be:

```suggestion
      {!isSameETHAddress(authorizer, account as string) && (
```

or maybe I've missed something?

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -76,16 +53,20 @@ const ConfirmStakingParamsModal: FC<
...
```
michalsmiarowski, line 69: (Above line not related, but file is related)

Maybe we could also addres this section in this PR:

image

I think this is not displayed anymore. Based no Figma this section is changed to:

image

Figma link: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=5658%3A336987

Link to code section:
https://github.com/threshold-network/token-dashboard/blob/5d49bf39c70b920f60d112cc4b89064c6024d2c6/src/components/Modal/ConfirmStakingParams/index.tsx#L79-L85

michalsmiarowski (changes requested):
Left some more comments to look at before the merge

src/components/Modal/ConfirmStakingParams/AdvancedParamsForm.tsx
georgeweiler: 🙈

src/pages/Staking/HowItWorks/NewTStakesCard.tsx
georgeweiler: Agreed. I've added some flags where the pull requests refers to authorizing.  We can keep the same `StakingTimeline` component, I just added a condition for the feature flag.

src/components/Modal/ConfirmStakingParams/index.tsx
georgeweiler: Sorry about that. Should be resolved now. I've updated our InfoBox component so the margin props will work now.

src/components/Modal/ConfirmStakingParams/index.tsx
```diff
@@ -95,34 +95,18 @@ const ConfirmStakingParamsModal: FC<
...
-            formId="advanced-staking-params-form"
-            initialAddress={account as string}
-            onSubmitForm={onSubmit}
-            checkIfProviderUsed={checkIfProviderUsed}
-          />
-        </Collapse>
+        <Alert my={6}>
+          <AlertIcon />
+          Take note! These addresses cannot be changed later.
+        </Alert>
```
michalsmiarowski, line 101: Based on Figma views the Alert should have status Info and it should be "small".

We should probalby add `sizes` property to the theme in `components` repo and add two sized - large and small (these are the only two we have in Threshold designs). Although' I've tried doing that in Alert theme:

```
sizes: {
    sm: {
      fontSize: "sm",
    },
    lg: {
      fontSize: "lg",
    },
  },
  ```

and it didn't work when i add `size=sm` to the Alert (but it worked when I added fontSize="sm" to the Alert).

We should also reduce the size of the Alert Icon for small size.

We can just add:

```suggestion
        <Alert my={6} status="info" fontSize="sm">
          <AlertIcon />
          Take note! These addresses cannot be changed later.
        </Alert>
```

for now if the there will be any problems with above solution.

michalsmiarowski (changes requested):
Some additional comment:

georgeweiler:
closing because I've opened https://github.com/threshold-network/token-dashboard/pull/185 which has signed commits.

Issue #134, opened 2022-07-20
Data access layer for authorization functions


pdyraga:
We want to have a separate data access layer for fetching and submitting data to the chain instead of having all the code implemented inside a React hook. This issue aims at implementing this layer for functions needed to implement authorization-related features in the dashboard. The data access layer should be covered with unit tests.

Functionalities needed (possibly separate PRs):
- [ ] Fetch authorized amount for TBTC and Random Beacon. The function to check is [`authorizedStake(address stakingProvider, address application)`](https://github.com/threshold-network/solidity-contracts/blob/61b2e45a2525d88e332595afd239c64afa97d153/contracts/staking/TokenStaking.sol#L1188). The application address is `RandomBeacon` contract address and `WalletRegistry` (TBTC) contract address that can be obtained from the NPM package.
- [ ] Fetch the amount being deauthorized and the time until the deauthorization can be completed. The function to check is [`pendingAuthorizationDecrease(address stakingProvider)`](https://github.com/keep-network/keep-core/blob/615c5872be8f8fa1992222369c0ac2ac990b86d5/solidity/ecdsa/contracts/WalletRegistry.sol#L1091) and [`remainingAuthorizationDecreaseDelay(address stakingProvider)`](https://github.com/keep-network/keep-core/blob/615c5872be8f8fa1992222369c0ac2ac990b86d5/solidity/ecdsa/contracts/WalletRegistry.sol#L1104) on `WalletRegistry` (TBTC) and `RandomBeacon` contracts.
[...]

PR #165, opened 2022-08-22
Data access layer for authorization functions

r-czajkowski:
Closes: #134

This PR creates the data access layer for the Threshold network- we want a separate layer for fetching on-chain data and submitting data to the chain. Thanks to this layer we can separate work between UI and contract integration. Also, it helps to unit test created services and move the "business logic" from the React hooks. For now, this lib supports authorization-related features and multicall contract that allows multiple smart contract constant function calls to be grouped into a single call and the results aggregated into a single result.

### Application
An interface representing the staking applications. Responsible for fetching authorization-related parameters and application data for a given staking provider such as authorized amount, amount being deauthorized and the time until the deauthorization can be completed.

### Staking
An interface representing the staking contract- should be responsible for fetching and submitting staking-related data. Currently, only requires implementing the `authorizedStake` function. In follow-up work, we are going to add more functions to this interface eg. a function that fetches all owner stakes etc.

### Multicall
Allows multiple smart contract constant function calls to be grouped into a single call and the results aggregated into a single result. This reduces the number of separate JSON RPC requests that need to be sent.


src/threshold-ts/applications/index.ts
```diff
@@ -0,0 +1,129 @@
+import { BigNumber, Contract, ContractInterface } from "ethers"
+import { getContract, isAddress, isAddressZero } from "../utils"
+import { IStaking } from "../staking"
+
+/**
+ * Represents an applciation.
+ */
+export interface IApplication {
```
georgeweiler, line 8: Is it worth defining this as a "StakingApplication"? I think `Application` is a very generic term

src/threshold-ts/applications/index.ts
```diff
@@ -0,0 +1,129 @@
...
+  async isEligibleForRewards(stakingProvider: string): Promise<boolean> {
+    const operator = await this._application.stakingProviderToOperator(
+      stakingProvider
+    )
+
+    if (isAddress(operator) && isAddressZero(operator)) {
+      return false
+    }
+
+    const isInPool: boolean = await this._application.isOperatorInPool(operator)
```
georgeweiler, line 118: Will this work for PRE app?

src/threshold-ts/index.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: Implement Threshold wrapper that includes all services
+export const Threhosld = {}
```
georgeweiler, line 2: Typo

src/threshold-ts/mas/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  public readonly ecdsa: IApplication
+
+  constructor(staking: IStaking, multicall: IMulticall) {
+    this._staking = staking
+    this._multicall = multicall
+    this.randomBeacon = new Application(this._staking, {
+      address: RandomBeacon.address,
+      abi: RandomBeacon.abi,
+    })
+    this.ecdsa = new Application(this._staking, {
```
georgeweiler, line 33: I feel like there's a naming mismatch between ecdsa and tbtc in this class. We should leave some comments explaining the relationship between them. or maybe ecsda is just a stub for now before tbtc is ready?

src/threshold-ts/applications/index.ts
r-czajkowski: Nope, we will have to create a new impl of `Application` interface for PRE but I think we can skip the PRE implementation for now because the PRE app doesn't require authorization and the `SimplePreApplication` contract doesn't even implement the `IApplication`interface required by the Threshold staking contract.

src/threshold-ts/applications/index.ts
r-czajkowski: Both options work for me, but I kept the same convention as on the contract side. See https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/IApplication.sol. I am curious about @michalsmiarowski  opinion here.

src/threshold-ts/applications/index.ts
```diff
@@ -0,0 +1,241 @@
...
+   * Delay in seconds that needs to pass between the time authorization decrease
+   * is requested and the time that request gets approved. Protects against
+   * free-riders earning rewards and not being active in the network.
+   */
+  authorizationDecreaseDelay: BigNumber
+  /**
+   * Authorization decrease change period in seconds. It is the time, before
+   * authorization decrease delay end, during which the pending authorization
+   * decrease request can be overwritten. If set to 0, pending authorization
+   * decrease request can not be overwritten until the endire
```
michalsmiarowski, line 24: Typo?

```suggestion
   * decrease request can not be overwritten until the entire
```

src/threshold-ts/applications/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Calculates reward eligibility for an application.
+   * @returns `true` if the staking provider is eligible for rewards, otherwise
+   * returns `false`.
+   */
+  isEligibleForRewards(stakingProvider: string): Promise<boolean>
+
+  /**
+   * Returns the application data for a given staking provider. This function
+   * uses the multicall interface to aggragate contract calls into a single one.
```
michalsmiarowski, line 115:
```suggestion
   * uses the multicall interface to aggregate contract calls into a single one.
```

src/threshold-ts/applications/index.ts
```diff
@@ -0,0 +1,241 @@
...
+   */
+  pendingAuthorizationDecrease: BigNumber
+  /**
+   * Time in seconds until the deauthorization can be completed.
+   */
+  remainingAuthorizationDecreaseDelay: BigNumber
+}
+
+/**
+ * Represents an applciation.
```
michalsmiarowski, line 47: Typo:

```suggestion
 * Represents an application.
```

src/threshold-ts/mas/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    ]
+
+    const [tbtcMinAuthorizationParams, randomBeaconMinAuthorizationParams] =
+      await this._multicall.aggregate(calls)
+
+    return {
+      tbtc: tbtcMinAuthorizationParams,
+      randomBeacon: randomBeaconMinAuthorizationParams,
+    }
+  }
```
michalsmiarowski, line 69: I'm not a master in solidity so I might be wrong here, but are you sure `tbtc` and `randomBeacon` that are returned from the function are `BigNumbers`?

The [authorizationParameters function in randomBeacon contract](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/RandomBeacon.sol#L1457) returns three values not one.

We could probably benefit from a tests for that function. I think there are not any right now.

src/threshold-ts/multicall/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const MULTICALL_ABI = [
+  "function aggregate(tuple(address target, bytes callData)[] calls) view returns (uint256 blockNumber, bytes[] returnData)",
+  "function getEthBalance(address addr) view returns (uint256 balance)",
+]
+export const MULTICALL_ADDRESSESS = {
+  1: "0xeefba1e63905ef1d7acba5a8513c70307c1ce441",
+  5: "0x77dca2c955b15e9de4dbbcf1246b4b85b651e50e",
+  1337: process.env.REACT_APP_MULTICALL_ADDRESS || AddressZero,
+} as Record<number | string, string>
```
michalsmiarowski, line 37: Maybe we could extract those two variables to a separate file, like `config.ts` and keep it in the same directory?

What do you think?

michalsmiarowski (changes requested):
Overall looks awesome 🔥 

I checked the code and it looks fantastic! 🎖️ Left some comments to look at.

I haven't reviewed the tests yet but I'll do it soon.

src/contexts/ThresholdContext.tsx
```diff
@@ -0,0 +1,37 @@
+import { createContext, FC, useContext, useMemo } from "react"
+import { useWeb3React } from "@web3-react/core"
+import { JsonRpcProvider, Provider } from "@ethersproject/providers"
+import { Threshold } from "../threshold-ts"
+import { EnvVariable } from "../enums"
+import { getEnvVariable, supportedChainId } from "../utils/getEnvVariable"
+
+const getThresholdLib = (providerOrSigner?: Provider) => {
```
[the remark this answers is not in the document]
r-czajkowski: We can pass the provider instance or signer to the Contract instance please see https://docs.ethers.io/v5/api/contract/contract/#Contract--creating so that's why there is a `provixerOrSigner`. I'll update type.

src/threshold-ts/multicall/index.ts
r-czajkowski: Those two variables are strictly related to the Multicall contract and IMO here, is the best place for them.

src/threshold-ts/mas/index.ts
r-czajkowski: Good catch! We should use `SupportedAppAuthorizationParameters` type. Will update.

src/contexts/ThresholdContext.tsx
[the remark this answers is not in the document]
r-czajkowski: 4a49da7

src/threshold-ts/mas/index.ts
r-czajkowski: Types: 79f29ee
Unit tests: d40342d

michalsmiarowski (dismissed):
I haven't found any more issues with this PR so I'm gonna approve this.

Not gonna merge this one yet in case @pdyraga would like to take a look on it too.

package.json
```diff
@@ -13,8 +13,10 @@
     "@ethersproject/hardware-wallets": "^5.5.0",
     "@fontsource/inter": "^4.5.10",
     "@keep-network/coverage-pools": "development",
+    "@keep-network/ecdsa": "development",
     "@keep-network/keep-core": "development",
     "@keep-network/keep-ecdsa": "development",
+    "@keep-network/random-beacon": "development",
```
michalinacienciala, line 19: Adding dependency to the `random-beacon` and `ecdsa` packages means that we also need to update CI (so that the correct version of contracts would be used when testing/deploying).
There is couple of PRs in works that change the CI workflows, and depending on when #165 is going to be merged, we may need to do different changes.
No matter what, we need to upgrade the new dependencies in this step: https://github.com/threshold-network/token-dashboard/blob/main/.github/workflows/dashboard-ci.yml#L69-L78.

If #165 were to be merged now, we would also need to update the https://github.com/threshold-network/token-dashboard/blob/main/.github/actions/build-and-deploy-to-bucket/action.yml#L69-L103 in the action that builds and publishes the dashboard (similarly to how `solidity-contracts` are handled).
If before merge of #165, the https://github.com/threshold-network/token-dashboard/pull/159 gets merged, then as part of #165 we would need to update the `.github/workflows/build-and-publish.yml`.

michalsmiarowski (approved):
🌳 🦌

PR #166, opened 2022-08-22
Update preparation of the artifacts for `solidity-contracts` in `start_dashboard.sh` script 

michalsmiarowski:
In this PR - https://github.com/threshold-network/solidity-contracts/pull/113 - there was a change that changed the way of how we prepare the artifacts in `solidity-contracts` repo. Instead of calling `prepare-artifacts.sh` file we use hardhat helper task (please see this commit - https://github.com/threshold-network/solidity-contracts/pull/113/commits/0c843c620fa0d7e54474736933ad56d5aa3ba3ea).

This PR updates the `start_dashboard` script in `token-dashboard` to do the same, since `prepare-artifacts.sh` file was removed.


r-czajkowski (approved):
:shipit:

PR #168, opened 2022-08-22
Extends input interfaces to allow JSX as helper/error messages

georgeweiler:
New designs have some links and alerts as the helper text for a few inputs, so this PR extends the interface to allow for custom components


src/components/Forms/HelperErrorText.tsx
```diff
@@ -0,0 +1,26 @@
+import { FC } from "react"
+import { FormErrorMessage, FormHelperText } from "@chakra-ui/react"
+
+const HelperErrorText: FC<{
+  errorMsgText?: string | JSX.Element
+  hasError?: boolean
+  helperText?: string | JSX.Element
+}> = ({ errorMsgText, helperText, hasError }) => {
+  if (hasError) {
+    return typeof errorMsgText === "string" ? (
```
r-czajkowski, line 10: Why can't we render the `JSX.Element` inside `FormErrorMessage` or `FormHelperText`? It would simplify this component.

```jsx
if(hasError) {
 return (
  <FormErrorMessage>
   {errorMsgText || "Please enter a valid value"}
  </FormErrorMessage>
 )
}

return helperText && (
 <FormHelperText>
  {helperText}
 </FormHelperText>
)
```

r-czajkowski (changes requested):

src/components/Forms/HelperErrorText.tsx
georgeweiler: My thinking is that the `FormErrorMessage`/`FormHelperText` would be an unnecessary dom node and might apply unintended styles to its children but I guess it's not a big deal. The code you suggested is cleaner so I changed it, but I'm returning null if there's no text to render since it's a JSX component.

https://github.com/threshold-network/token-dashboard/pull/168/commits/1c822323b3bc62ed52588bb1d678486179409535

r-czajkowski (approved):

michalsmiarowski (approved):
LGTM 🐔

PR #169, opened 2022-08-22
Update validation for min stake

georgeweiler:
Adds loading and error variables for the min-stake fetch and fixes two small bugs in the gorli development env: 

1. The min stake is set to 0 for Gorli, so the input incorrectly shows "min stake loading"
2. The input no longer validates on Gorli, fixing the "value must be less than 0" bug


src/pages/Staking/StakedPortfolioCard.tsx
```diff
@@ -53,12 +66,11 @@ const StakedPortfolioCard: FC<ComponentProps<typeof Card>> = (props) => {
...
-        placeholder={`Minimum stake ${
-          minStakeAmount === "0"
-            ? "loading..."
-            : `${formatTokenAmount(minStakeAmount)} T`
-        }`}
+        placeholder={placeholder}
         minTokenAmount={minStakeAmount}
+        shouldValidateForm={
+          supportedChainId.toString() === ChainID.Ethereum.toString()
+        }
```
r-czajkowski, line 73: TBH I don't understand why we need this. Could you please explain? I tested locally w/o this prop and T dapp works as expected.

src/hooks/useMinStakeAmount.ts
```diff
@@ -9,17 +9,25 @@ export const useMinStakeAmount = () => {
...
         dispatch(setMinStake({ amount: minStakeAmount.toString() }))
+        setLoading(false)
+        setHasError(false)
       } catch (error) {
+        setLoading(false)
+        setHasError(true)
         console.error("Could not fetch the min stake amount: ", error)
       }
     }
     if (minStakeAmount === "0" && tStakingContract) fetchMinStakeAmount()
```
r-czajkowski, line 29: I've noticed that if the minimum stake is `0` this hook triggers the `fetchMinStakeAmount` function every time we visit the `staking` page. So my suggestion is:
- set the `minimumStake` in store as `undefined` as the default value- it means that the dapp has not fetched the minimum stake amount yet,
- condition in hook change to `if (minStakeAmount === undefined && tStakingContract)`,
- change the initial value for `isLoading` to `false`: `const [isLoading, setLoading] = useState(false)`.

src/pages/Staking/StakedPortfolioCard.tsx
```diff
@@ -14,21 +14,34 @@ import { TokenAmountForm } from "../../components/Forms"
...
 
+  const placeholder = useMemo(() => {
+    if (hasError) {
+      return "Error fetching min stake"
+    }
+
+    return `Minimum stake ${
+      isLoading ? "loading..." : `${formatTokenAmount(minStakeAmount)} T`
+    }`
+  }, [isLoading, hasError])
```
r-czajkowski, line 43: Missing `minStakeAmount` dependency.

r-czajkowski (changes requested):

cygnusv:
Note there's an upcoming change to the staking contract that changes how the stake amount is validated:

https://github.com/threshold-network/solidity-contracts/pull/122

In particular, staking 0 tokens is disallowed

src/pages/Staking/StakedPortfolioCard.tsx
georgeweiler: Whoops, good catch. Removing.

src/hooks/useMinStakeAmount.ts
```diff
@@ -9,17 +9,26 @@ export const useMinStakeAmount = () => {
   const { minStakeAmount } = useStakingState()
   const dispatch = useDispatch()
 
+  const [isLoading, setLoading] = useState(false)
+  const [hasError, setHasError] = useState(false)
+
   useEffect(() => {
     const fetchMinStakeAmount = async () => {
+      console.log("rendering the hook")
```
r-czajkowski, line 17: console log.

r-czajkowski (changes requested):
Let's remove unnecessary console log and :shipit:

r-czajkowski (approved):
🚢

PR #173, opened 2022-08-31
Update rewards emission dates

r-czajkowski:
Next emissions date is October 1st, and then every 1st day of the month
until the end of year.


michalsmiarowski (approved):
🥇

PR #172, opened 2022-08-25
Updates the detailed link component

georgeweiler:
- Updates the detailed link list item component to include an image with fallback
- Adds staking provider logos

PR #175, opened 2022-08-31
WIP: Staking how it works updates

georgeweiler:
Creates the updated "How it works" page with tabs.

WIP as designs for the first two sections are still in flux.  Staking Providers tab is complete.

Depends on #172


georgeweiler:
Closing because #198 has signed commits and is updated to use the new Filter Tabs routing system

PR #178, opened 2022-09-05
MAS UI <-> data access layer integration

r-czajkowski:
~Depends on: #165~

This PR integrates the data access layer added in #165 with MAS UI.

### Listener middleware
Add a [listener middleware](https://redux-toolkit.js.org/api/createListenerMiddleware) to redux store. Lets us define "listener" entries that contain an "effect" callback with additional logic, and a way to specify when that callback should run based on dispatched actions or state changes. It's helpful when we want to fetch data based on other fetched data and stored in the redux store. Eg. We want to fetch the application data for each owner stake- so we add a listener which will trigger an effect when the `setStakes` action will be dispatched. We could aslo create a one-shot listener by unconditionally calling `unsubscribe()` in the body - the effect callback would run the first time the relevant action is seen, then immediately unsubscribe and never run again.

### Threshold-ts "lib"
- Update  `EthereumConfig` type- add an `account` field to be able to create a contract instance with a signer to sign
transaction,
- Add `updateConfig` function to `Threshold` class- to reinitiate services with updated config eg. we should call this function when a user connects to a wallet and an Ethereum provider changes,
- Add `increaseAuthorization` function- to authorize a given staking provider for the given application by the given amount,
- Add missing unit tests.

### Threshold React context
- Update the Threshold config when a user connects to a wallet because the provider has changed- we need to update provider to be able to sign transactions,
- Create one instance of `Threshold` lib to use the same instance with the same config in redux middleware and react context.

### Redux Store
- Add `staking-applications` slices to store the staking applications data such as authorization parameters for a given application and to store staking providers applications data,
- Add listener effects that fetch auth-related data and save data in a store.

### UI integration
- Display data from redux store in UI components
<details>
  
  
obraz
obraz

</details>


src/hooks/store.ts
```diff
@@ -0,0 +1,5 @@
+import { useDispatch, useSelector, TypedUseSelectorHook } from "react-redux"
+import { RootState, AppDispatch } from "../store"
+
+export const useAppDispatch: () => AppDispatch = useDispatch
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
```
michalsmiarowski, line 5: Wouldn't it be better to keep the rule of one hook in one file and naming the file as the hook? We could put those two hook inside a `store` directory, and also fours hooks inside 'hooks/application.ts' to separate directory:


```
hooks
├── application
│   ├── useApplicationState.ts
│   ├── useAppParameters.ts
│   ├── useAppMinAuthorizationAmount.ts
│   └── useAppDataByStakingProvider.ts
├── store
│   ├── useAppDispatch.ts
│   └── useAppSelector.ts
├── useAddErc20ToMetamask.ts
├── useChakraBreakpoint.ts
{...rest of the hooks}
```

Also if you agree with aboce comments then don't forget to change `application` to `stakingApplication` in proper places

src/pages/Staking/AuthorizeStakingApps/index.tsx
```diff
@@ -34,8 +38,17 @@ const AuthorizeStakingAppsPage: PageComponent = (props) => {
...
+  const randomBeaconMinAuthAmount = useAppMinAuthorizationAmount("randomBeacon")
+
+  const tbtcApp = useAppDataByStakingProvider(
+    "tbtc",
+    stakingProviderAddress || AddressZero
+  )
+  const randomBeaconApp = useAppDataByStakingProvider(
+    "randomBeacon",
+    stakingProviderAddress || AddressZero
+  )
```
michalsmiarowski, line 51: ~~I understand the idea but don't you think this might be a little bit over-engineered? We could simply just get the state with:~~

~~useSelector((state: RootState) => state.applications)~~

~~and then get the data we want and assign it to `appsAuthData` variable few lines below.~~

EDIT:

I read a little more about this and actually I think your solution makes more sense. Here is a good thread on this subject on stack overflow: [https://stackoverflow.com/a/63277926](https://stackoverflow.com/a/63277926). Gonna keep this comment anyway if you would like to add some more about it

src/utils/percentage.ts
```diff
@@ -1,3 +1,23 @@
...
+    BigNumber.from(amount).isZero() ||
+    !totalAmount ||
+    BigNumber.from(totalAmount).isZero()
+  )
+    return 0
+
+  return FixedNumber.fromString(amount.toString())
+    .divUnsafe(FixedNumber.fromString(totalAmount.toString()))
+    .mulUnsafe(FixedNumber.fromString("100"))
+    .toUnsafeFloat()
```
michalsmiarowski, line 18: Why we use `FixedNumber` here instead of `BigNumber`?

src/pages/Staking/index.tsx
```diff
@@ -9,15 +9,21 @@ import { useFetchTvl } from "../../hooks/useFetchTvl"
...
 import AuthorizeStakingAppsPage from "./AuthorizeStakingApps"
+import { applicationsSlice } from "../../store/applications/slice"
 
 const StakingPage: PageComponent = (props) => {
   const [data, fetchtTvlData] = useFetchTvl()
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    dispatch(applicationsSlice.actions.getSupportedApps({}))
+  }, [dispatch])
```
michalsmiarowski, line 26: Does that mean that we are gonna fetch the app data only on `Staking` page? We've got a banner on overview page that says about the app authorization and which let the user choose the stake he wants to authorize. We might need this data there too.

src/store/applications/slice.ts
```diff
@@ -0,0 +1,141 @@
...
+        data: ApplicationDataByStakingProvider
+      }>
+    ) => {
+      const { appName, data } = action.payload
+      // @ts-ignore
+      state[appName].stakingProviders = {
+        isFetching: false,
+        error: "",
+        ...data,
+      }
```
michalsmiarowski, line 108: Unnecessary `@ts-ignore`?

michalsmiarowski (changes requested):
First part of review done:

src/utils/percentage.ts
r-czajkowski: Because `BigNumber` doesn't support decimals.

michalsmiarowski: Ahh, that's true! 🤦 😅

src/store/applications/slice.ts
```diff
@@ -0,0 +1,141 @@
...
+  parameters: AuthorizationParameters<string> & FetchingState
+  stakingProviders: ApplicationDataByStakingProvider & FetchingState
+}
+
+export interface ApplicationsState {
+  tbtc: ApplicationState
+  randomBeacon: ApplicationState
+}
+
+export type AppName = "tbtc" | "randomBeacon"
```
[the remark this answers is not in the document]
r-czajkowski: In terms of threshold staking applications, we have only tbtc v2 app so I think from code's/T dapp's perspective is ok to use `tbtc` w/o `v-2` because it's obvious that only tbtc-v2 works on threshold network. So maybe let's use suffix for `tbtc-v1`? What do you think? We only added the `tbtc v1` token to the `TokenContext` to get the token price in usd and calculate tvl.

src/hooks/application.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  AppName,
+  selectAppByStakingProvider,
+  selectAppStateByAppName,
+} from "../store/applications"
+import { useAppSelector } from "./store"
+
+export const useApplicationState = (appName: AppName) => {
```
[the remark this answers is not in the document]
r-czajkowski: 367be12

src/hooks/store.ts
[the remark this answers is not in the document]
r-czajkowski: 367be12

r-czajkowski: 5b96c75

src/App.tsx
```diff
@@ -48,6 +49,8 @@ const Web3EventHandlerComponent = () => {
   useSubscribeToStakedEvent()
   useSubscribeToUnstakedEvent()
   useSubscribeToToppedUpEvent()
+  useSubscribeToAuthorizationIncreasedEvent("tbtc")
```
georgeweiler, line 52: do you think we should use enums for these? And in other places in the d'app?

src/contexts/ThresholdContext.tsx
```diff
@@ -1,34 +1,28 @@
...
 
 export const ThresholdProvider: FC = ({ children }) => {
-  const { library, active } = useWeb3React()
+  const { library, active, account } = useWeb3React()
 
-  const threshold = useMemo(() => {
-    return getThresholdLib(active ? library : undefined)
-  }, [library, active])
+  useEffect(() => {
+    if (active && library && account) {
```
georgeweiler, line 16: do we need a cleanup function here if the user disconnects?

src/App.tsx
georgeweiler: nevermind I see the type `StakingAppName`

src/pages/Staking/StakeCard/StakeApplications/index.tsx
```diff
@@ -11,22 +11,31 @@ import {
...
+    "tbtc",
+    stake.stakingProvider
+  )
+  const randomBeaconApp = useStakingAppDataByStakingProvider(
+    "randomBeacon",
+    stake.stakingProvider
+  )
 
   // TODO: This will probably be fetched from contracts
   const appsAuthData = {
```
georgeweiler, line 28: I think it makes sense to store all of this information in the redux slice, rather than rebuilding these objects here. wdyt?

georgeweiler:
@r-czajkowski I see this error when pulling changes. Is there a special setup I need?

```
Could not fetch supported apps auth parameters Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="authorizationParameters()", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.6.4)
    at Logger.makeError (index.ts:261:1)
    at Logger.throwError (index.ts:273:1)
    at Interface.decodeFunctionResult (interface.ts:427:1)
    at index.ts:66:1
    at Array.map (<anonymous>)
    at Multicall.aggregate (index.ts:64:1)
    at async MultiAppStaking.getSupportedAppsAuthParameters (index.ts:58:1)
    at async Object.getSupportedAppsEffect [as effect] (effects.ts:25:1)
```

src/hooks/staking-applications/useStakingAppDataByStakingProvider.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  selectStakingAppByStakingProvider,
+  StakingAppName,
+} from "../../store/staking-applications"
+import { useAppSelector } from "../store"
+
+export const useStakingAppDataByStakingProvider = (
+  appName: StakingAppName,
+  stakingProvider: string
+) => {
```
georgeweiler, line 10: Do we have an interface defined somewhere for the return value of this?

src/store/applications/slice.ts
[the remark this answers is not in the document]
michalsmiarowski: @r-czajkowski  I think we might do it that way. It's still might get confusing since we use the terms `tbtc` and `tbtc-v2` outside of the project, but since we don't use `tbtc-v1` that much in the project (as you said) I think your solution is also very good 👍

src/pages/Staking/StakeCard/StakeApplications/index.tsx
r-czajkowski: Based on the docs https://redux.js.org/usage/deriving-data-selectors:
> should keep the Redux state minimal, and derive additional values from that state whenever possible

I think it's better eg. calculate percentage in selector because less logic is needed to calculate this value because it depends on the total staked amount of a given stake- so we can keep them it sync with the staking data.

Regarding `appsAuthData` and `AppAuthDataProps`, I'm going to refactor this interface because it's problematic and gives a lot of unnecessary logic (see my PR #189). But I would like to address it in a separate PR.

src/contexts/ThresholdContext.tsx
r-czajkowski: 5ac843f

src/pages/Staking/index.tsx
```diff
@@ -9,15 +9,21 @@ import { useFetchTvl } from "../../hooks/useFetchTvl"
...
 import AuthorizeStakingAppsPage from "./AuthorizeStakingApps"
+import { stakingApplicationsSlice } from "../../store/staking-applications/slice"
 
 const StakingPage: PageComponent = (props) => {
   const [data, fetchtTvlData] = useFetchTvl()
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    dispatch(stakingApplicationsSlice.actions.getSupportedApps({}))
+  }, [dispatch])
```
michalsmiarowski, line 26: There is a bug with getting minimum amount that is required to authorize for TBTC and Random Beacon. We fetch the data here in `Staking/index.tsx` and it works fine when you go to the authorize apps page through the staking page:

image

Ther problem is when you refresh the page (or simply go straight to authorize apps url) and connect the wallet:

image

The minimum amount for TBTC and Random Beacon is not fetched properly as it shows `0`

michalsmiarowski (changes requested):
Found 1 bug, see comment below

michalsmiarowski (approved):
I'm gonna approve it since i haven't found any other issues. 

Let's not forget about this bug: https://github.com/threshold-network/token-dashboard/pull/178#discussion_r972830485 and address it in a separate PR @r-czajkowski 

Let's goo 🤙

PR #182, opened 2022-09-07
Add StakeCardContext

michalsmiarowski:
Main changes:
- adds context to a StakeCard
- refactor `getStakeType` function
- refactor child components of StakeCard component

# Changes details

### adds context to a StakeCard

Adds context to a StakeCard with these values:
```
  isInactiveStake: boolean
  canTopUpKepp: boolean
  canTopUpNu: boolean
  hasLegacyStakes: boolean
  isPRESet: boolean
```

This way we can avoid props drilling.

Related discussion: https://github.com/threshold-network/token-dashboard/pull/156#discussion_r957398910

### refactor `getStakeType` function

Renames the `getStakeType` function to `getStakeTitle`. It is renamed because `getStakeType` returned a type (string) that needed to be concatenated with a "stake" word to create a title, so `getStakeType` might be misleading. Instead of that we just return a full stake title from this function.

### refactor child components of StakeCard component
There were a lot of components inside StakeCard components that requred the whole stake to be passed to them even though it only needs some specific information about it. I've change it to not pass the whole stake data where it's
not needed.


src/pages/Staking/StakeCard/index.tsx
```diff
@@ -26,16 +26,40 @@ import StakeRewards from "./StakeRewards"
 import StakeBalance from "./StakeBalance"
 import StakeAddressInfo from "./StakeAddressInfo"
 import { featureFlags } from "../../../constants"
+import { StakeCardContext } from "../../../contexts/StakeCardContext"
+import { useStakeCardContext } from "../../../hooks/useStakeCardContext"
+
+const StakeCardContainer: FC<{ stake: StakeData }> = ({ stake }) => {
```
r-czajkowski, line 32: Maybe `StakeCardProvider`? The `Container` term seems to be related to the UI like `div/section` wrapper.

src/utils/getStakeTitle.ts
```diff
@@ -0,0 +1,13 @@
+// import { StakeData } from "../types"
```
r-czajkowski, line 1: Unnecessary comment?

r-czajkowski (commented):
Left 2 non-blocking comments.

src/pages/Staking/StakeCard/index.tsx
michalsmiarowski: Fixed in 38cfb32de3fa544d6a285513b913f6c1a9685a03

src/utils/getStakeTitle.ts
michalsmiarowski: Fixed in 92d9b11a8218da3ee5d69170f52357487713630e (also removed other unnecessary comments)

r-czajkowski (approved):

PR #189, opened 2022-09-12
MAS UI integration v2

r-czajkowski:
Depends on: #178 

This PR adds support for multi-staking apps authorization.

## Changes
- display new apps to authorize when a user connects to a wallet and at least one staking provider has an unauthorized staking application, <details>
  
  obraz
</details>

- display `authorized` status in staking application card, <details>
  
  obraz
</details>

- add validation for multi app staking form- if a user selects an application to authorize they must fill the form otherwise we should display validation messages,<details>
  
  obraz
</details>

- multi staking applications authorization flow- pre-submit tx modal and success modal. <details>
  
  obraz
  obraz
</details>


src/pages/Staking/AuthorizeStakingApps/AuthorizeApplicationsCardCheckbox/index.tsx
```diff
@@ -38,25 +86,43 @@ export const AuthorizeApplicationsCardCheckbox: FC<
...
+          {
+            appName: appAuthData.label,
+            authorizationAmount: tokenAmount,
+            address: stakingAppAddress,
+          },
+        ],
+      })
+    } else {
+      // TODO: Create a increase authroziation modal.
+      console.log("open increase authroziation modal")
```
r-czajkowski, line 116: Done in #196.

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    type={ExplorerDataType.TRANSACTION}
+                  />
+                  {index + 1 === authorizedStakingApplications.length
+                    ? " "
+                    : " and "}
+                </Fragment>
+              ))}
+              on Etherscan
+            </>
+          )}
```
michalsmiarowski, line 136: Maybe we could specify here which transaction is which by displaying the name of the app? For example:

`View TBTC authorization transaction and Random Beacon transaction on Etherscan`

michalsmiarowski (changes requested):
First part of the review done. Left some comments below 👀

src/store/staking-applications/effects.ts
```diff
@@ -150,3 +155,41 @@ const getKeepStakingAppStakingProvidersData = async (
...
+    )
+    .some(
+      (stakingProviderAppInfo) =>
+        stakingProviderAppInfo.authorizedStake &&
+        stakingProviderAppInfo.authorizedStake === "0"
+    )
+
+  if (hasAnyUnauthroizedStakes) {
+    listenerApi.dispatch(openModal({ modalType: ModalType.NewAppsToAuthorize }))
+  }
```
michalsmiarowski, line 182: I think this, or even the whole function should be wrapped with `MULTI_APP_STAKING` feature flag

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  return (
+    <>
+      <ModalHeader>Authorize Apps</ModalHeader>
+      <ModalCloseButton />
+      <ModalBody>
+        <InfoBox variant="modal" mb="6" mt="0">
+          <H5>
+            You are authorizing the Threshold Application
+            {numberOfApps > 1 ? "s" : ""} to use your stake.
```
michalsmiarowski, line 64: Just pointing out that on Figma it says:

image

Link: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A178394

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <ListItem key={app.appName}>
+              <StakingApplicationToAuth
+                {...app}
+                stakingProvider={stakingProvider}
+                totalInTStake={totalInTStake}
+              />
+            </ListItem>
+          ))}
+        </List>
+      </ModalBody>
```
michalsmiarowski, line 83: Are you gonna add the alerts that are in Figma designs?

image


Link: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A178394

And also:

image


Link: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A178369

michalsmiarowski: And also this one in stake Card:

Link: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A178187

michalsmiarowski (changes requested):

src/components/ExternalLink/index.tsx
```diff
@@ -29,11 +29,7 @@ const ExternalLink: FC<Props & LinkProps> = ({
       {...props}
     >
       {text}
-      {withArrow ? (
-        <Icon boxSize="12px" ml="1" as={FiArrowUpRight} color={finalColor} />
```
georgeweiler, line 33: Is this just a code cleanup change because the Icon inherits the final color?

src/pages/Staking/AuthorizeStakingApps/index.tsx
```diff
@@ -60,29 +88,76 @@ const AuthorizeStakingAppsPage: FC = () => {
...
       percentage: randomBeaconApp.percentage,
+      authorizedStake: randomBeaconApp.authorizedStake,
       isAuthRequired: true,
     },
     pre: {
+      stakingAppId: "pre",
       label: "PRE",
       isAuthorized: false,
       percentage: 0,
+      authorizedStake: stake.totalInTStake,
```
georgeweiler, line 117: This crashes the app if you try to access a URL such as `/staking/0xF408dE50D788183d9F4F46fbb34E3EbAC9D90d50/authorize` when the wallet is not connected, or when stakes are not loaded. It happens when you do a simple page refresh so I think we should create a fallback page (or redirect?) to capture this edgecase

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const { authorizeMultipleApps } = useAuthorizeMultipleAppsTransaction()
+  const onAuthorize = async () => {
+    await authorizeMultipleApps(
+      applications.map((_) => ({
+        address: _.address,
+        amount: _.authorizationAmount,
+      })),
+      stakingProvider
+    )
+  }
```
georgeweiler, line 53:
```suggestion
  const onAuthorize = useCallback(async () => {
    await authorizeMultipleApps(
      applications.map((_) => ({
        address: _.address,
        amount: _.authorizationAmount,
      })),
      stakingProvider
    )
  }, [authorizeMultipleApps, applications, stakingProvider])
```

src/pages/Staking/AuthorizeStakingApps/AuthorizeApplicationsCardCheckbox/AppAuthorizationInfo.tsx
```diff
@@ -60,6 +76,16 @@ export const AppAuthorizationInfo: FC<AppAuthorizationInfoProps> = ({
           {`${formatPercentage(slashingPercentage, 0, true)}`}
         </BoxLabel>
       </HStack>
+      {isAuthorizationRequired && isAuthorized && (
```
georgeweiler, line 79: I'm running into an issue where my `authorizedStake` value is `undefined` for a moment before it loads. This causes `formatTokenAmount` to crash the app because BigNumber can't handle `undefined`. Maybe it's a local config issue? The value eventually loads for me, so adding this fixed the issue. Up to you if you think it's worth it. Do you think we we see this bug in production?

```
{isAuthorizationRequired && isAuthorized && !!authorizedStake && <>content</>}
```

src/pages/Staking/AuthorizeStakingApps/index.tsx
```diff
@@ -60,29 +88,76 @@ const AuthorizeStakingAppsPage: FC = () => {
...
   }
 
-  const onAuthorizeApps = () => {
-    console.log("Authorize Apps!!", selectedApps)
+  const isAppSelected = (stakingAppName: AppAuthDataProps["stakingAppId"]) => {
+    return selectedApps.map((app) => app.stakingAppId).includes(stakingAppName)
+  }
+
+  const onAuthorizeApps = async () => {
+    const isTbtcSelected = isAppSelected("tbtc")
```
georgeweiler, line 127: if think you pull these out into their own variables in the component state such as:

```
    const isTbtcSelected = useMemo(() => isAppSelected("tbtc"))
```

then you can reference them in this function and in the JSX below without re-rendering the component an 2 extra times when that value is calculated.

src/pages/Staking/AuthorizeStakingApps/index.tsx
```diff
@@ -121,30 +196,33 @@ const AuthorizeStakingAppsPage: FC = () => {
           </AlertDescription>
         </AlertBox>
         <AuthorizeApplicationsCardCheckbox
+          formRef={tbtcAppFormRef}
           mt={5}
           appAuthData={appsAuthData.tbtc}
+          totalInTStake={stake.totalInTStake}
           onCheckboxClick={onCheckboxClick}
-          isSelected={selectedApps.map((app) => app.label).includes("tBTC")}
+          isSelected={isAppSelected("tbtc")}
```
georgeweiler, line 204: see comment above about extracting this to a variable

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
```diff
@@ -0,0 +1,159 @@
...
+const StakingApplicationsAuthorizedBase: FC<
+  StakingApplicationsAuthorizeProps
+> = ({ stakingProvider, authorizedStakingApplications, closeModal }) => {
+  const stake = useAppSelector((state) =>
+    selectStakeByStakingProvider(state, stakingProvider)
+  )
+  const navigate = useNavigate()
+  const onAuthorizeOtherApps = async () => {
+    closeModal()
+    navigate(`/staking/${stakingProvider}/authorize`)
```
georgeweiler, line 57: I think this is OK for now because I am doing the same in many modals, and I don't have a good solution to implement before the launch. But it's going to throw a warning, see: 

https://dev.to/jexperton/how-to-fix-the-react-memory-leak-warning-d4i

src/store/staking-applications/effects.ts
```diff
@@ -150,3 +155,41 @@ const getKeepStakingAppStakingProvidersData = async (
...
+      Object.values(
+        selectStakingAppStateByAppName(listenerApi.getState(), "randomBeacon")
+          .stakingProviders.data
+      )
+    )
+    .some(
+      (stakingProviderAppInfo) =>
+        stakingProviderAppInfo.authorizedStake &&
+        stakingProviderAppInfo.authorizedStake === "0"
+    )
```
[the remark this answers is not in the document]
r-czajkowski: That's weird. I use VSC as well and looks OK.

obraz

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const StakingApplicationToAuth: FC<{
+  appName: string
+  authorizationAmount: string
+  stakingProvider: string
+  totalInTStake: string
+}> = ({ appName, authorizationAmount, stakingProvider, totalInTStake }) => {
+  const percentage = formatPercentage(
+    calculatePercenteage(authorizationAmount, totalInTStake)
+  )
```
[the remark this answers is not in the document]
r-czajkowski: Good catch! I'll ping @sashatanase and @liz-shinn to make sure this solution will be ok.

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
r-czajkowski: Implemented based on Figma views https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6098%3A141093 but I'll ping @sashatanase and @liz-shinn as well 😆.

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
r-czajkowski: Yep, I'm going to add alerts in a separate PR- there are too many files to review now.

src/components/ExternalLink/index.tsx
r-czajkowski: Yes exactly. The previous impl won't work in a case like that:
```jsx
<Button
  variant="outline"
  as={ExternalLink}
  mr={2}
  href={link}
  text="text"
  color="red"
  withArrow
/>
```
This will render the arrow in default color even if we pass `color` prop- the icon won't inherit color from parent.

src/pages/Staking/AuthorizeStakingApps/index.tsx
r-czajkowski: Added conditional props in https://github.com/threshold-network/token-dashboard/pull/197/commits/34cf8066e1c72e3648df164e5f689d6202411d0d so the `authorizedStake` is not required if the `isAuthRequired` is set to `false`. But yeah we need to add support for fetching data when the user goes to `/staking/{some_provider}/authorize` directly and I'm going to add it in a separate PR. Does it work for you?

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
r-czajkowski: It won't memoize the submit function because the `applications` props is an array. You can check with simple test:
```jsx
const [increment, setIncrement] = useState(1) // run `setIncrement((i) => i+1)` on button click
  console.log("increment", increment)
  const onAuthorize = useCallback(async () => {
    await authorizeMultipleApps(
      applications.map((_) => ({
        address: _.address,
        amount: _.authorizationAmount,
      })),
      stakingProvider
    )
  }, [authorizeMultipleApps, applications, stakingProvider])

  useEffect(() => {
    console.log("updated")
  }, [onAuthorize])
```
Also I noticed that we use `openModal` in `useAuthorizeMultipleAppsTransaction` and in `useSendTranacion` which is not memoized and hook always returns a new instance of function on every render. See https://github.com/threshold-network/token-dashboard/blob/main/src/hooks/useModal.ts#L15-L16. We need to update it but I would address it in a separate PR.

src/pages/Staking/AuthorizeStakingApps/AuthorizeApplicationsCardCheckbox/AppAuthorizationInfo.tsx
r-czajkowski: I think it will be addressed once we add support for fetching data when a  user goes to the auth page directly see https://github.com/threshold-network/token-dashboard/pull/189#discussion_r976455968. Does it work for you if I address it in a separate PR?

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
r-czajkowski: Yeah we need to get rid of the warning some day. Reagrding async- it is unnecessary. I'll remove.

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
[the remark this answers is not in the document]
sashatanase: I agree with this solution. They both make sense to me.

src/pages/Staking/AuthorizeStakingApps/index.tsx
georgeweiler: yes works for me!

src/pages/Staking/AuthorizeStakingApps/AuthorizeApplicationsCardCheckbox/AppAuthorizationInfo.tsx
georgeweiler: yes that works

georgeweiler (approved):

src/pages/Staking/AuthorizeStakingApps/index.tsx
r-czajkowski: The same thing as here https://github.com/threshold-network/token-dashboard/pull/189#discussion_r976480676. It won't be memoized value because `isAppSelected` function uses the `selectedApps` array. To achieve that we should refactor this component to be more "static"- I mean store the boolean for each app in a separate variable like that instead of storing selected apps in an array:
```jsx
const [isTbtcAppSelected, setTbtcAppSelected] = useState(false)
const [isRandomBeaconAppSelected, setRandomBeaconAppSelected] = useState(false)
...

<AuthorizeApplicationsCardCheckbox
 <...otherProps>
 onCheckboxClick={(_, isChecked: boolean) => setTbtcAppSelected(isChecked)}
 isSelected={isTbtcAppSelected}
/>
```

I think this approach is OK, at least for now, because we have only 2 staking apps. Let me know what you think and maybe let's refactor in a separate PR.

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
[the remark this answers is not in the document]
r-czajkowski: da5ad3e

src/store/staking-applications/effects.ts
r-czajkowski: 9e4e044

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
r-czajkowski: fec8758

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
r-czajkowski: @michalsmiarowski let's update that once we get the final copy, ok?

src/store/staking-applications/effects.ts
```diff
@@ -150,3 +155,41 @@ const getKeepStakingAppStakingProvidersData = async (
     throw error
   }
 }
+
+export const displayNewAppsToAuthrozieModalEffect = async (
+  action: AnyAction,
+  listenerApi: AppListenerEffectAPI
+) => {
+  listenerApi.unsubscribe()
+  const hasAnyUnauthroizedStakes = Object.values(
```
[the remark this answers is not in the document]
r-czajkowski: bd465ca

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
r-czajkowski: 303b6ab

src/store/staking-applications/effects.ts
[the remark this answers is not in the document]
michalsmiarowski: Yup, works for me as well, maybe it was synchronizing or something 🤔

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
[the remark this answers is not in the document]
michalsmiarowski: @r-czajkowski Can we also change it in stake card?

image

https://github.com/threshold-network/token-dashboard/blob/c8622a7b096223909045a26953b2bb5d563324ef/src/pages/Staking/StakeCard/StakeApplications/AuthorizeApplicationRow.tsx#L51

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
michalsmiarowski: Gonna keep this convo as unresolved to not lost the track of it

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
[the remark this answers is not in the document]
michalsmiarowski: @r-czajkowski Sorry, but I just noticed another error related to this:

image

App is `>99%` authorized but it doesn't have checkbox as if it were 100% authorized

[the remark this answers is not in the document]
michalsmiarowski: And also it shows `>99%` and `<1%` for maximum T staked and 0 T staked. I think we should upgrade `formatPercentage` function:

https://github.com/threshold-network/token-dashboard/blob/c8622a7b096223909045a26953b2bb5d563324ef/src/utils/percentage.ts#L21-L30

to something like this (note: i did not test this, but I think it should work):

```
if (percentage < 1 && percentage > 0 && displayLessThanGreaterThanSigns) {
    return "<1%"
  } else if (percentage > 99 && percentage < 100 && displayLessThanGreaterThanSigns) {
    return ">99%"
  }
```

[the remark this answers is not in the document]
r-czajkowski: Based on the Figma views, my understanding is that we only want to display checkboxes when applications are not authorized eg here https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A175881. Let's clarify with @sashatanase and @liz-shinn.

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
```diff
@@ -0,0 +1,159 @@
...
+        </Alert>
+        <List spacing="2" mb="6">
+          <ListItem>
+            <HStack justifyContent="space-between">
+              <BodySm>Provider Address</BodySm>
+              <BodySm>{shortenAddress(stakingProvider)}</BodySm>
+            </HStack>
+          </ListItem>
+          {authorizedStakingApplications.map((_) => (
+            <ListItem>
```
michalsmiarowski, line 77: Should have unique `key` or warning will appear after authorazing both apps

michalsmiarowski: After authorizing both apps I alos see the warning with ref, but I can't figure out why it appears as of now

<details>
<summary>Screenshot</summary>
<br>
image
</details>

michalsmiarowski (changes requested):
Left few minor comments that I would like to be addressed before the merge

src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
r-czajkowski: 1c939e8 0ca4fe5

src/components/Modal/StakingApplications/AuthorizeStakingApps.tsx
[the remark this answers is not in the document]
r-czajkowski: > And also it shows `>99%` and `<1%` for maximum T staked and 0 T staked. I think we should upgrade `formatPercentage` function:
> 
> https://github.com/threshold-network/token-dashboard/blob/c8622a7b096223909045a26953b2bb5d563324ef/src/utils/percentage.ts#L21-L30
> 
> to something like this (note: i did not test this, but I think it should work):
> 
> ```
> if (percentage < 1 && percentage > 0 && displayLessThanGreaterThanSigns) {
>     return "<1%"
>   } else if (percentage > 99 && percentage < 100 && displayLessThanGreaterThanSigns) {
>     return ">99%"
>   }
> ```

37a06fb

michalsmiarowski (approved):
LGTM 🏋️

PR #199, opened 2022-09-22
Map operator to staking provider feature

michalsmiarowski:
Full flow of the mapping operator to the staking provider.

Figma designs: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A175846

Main changes:
- flow for mapping both apps at the same time
- flow for mapping one of the app if user rejects one of the transactions from the previous step
- simple modal queue

**TODO**: (in separate PR) **(addressed in #204)**
- add the mapping component to the Staking page
- refactor the staking page so that all the stakes are on the left side
- separate `new stake` and `staked portfolio`

Figma link: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6179%3A143943

<details>
  <summary>TODO screenshot from figma</summary>

image

</details>

# Changes details

### flow for mapping both apps at the same time

The whole flow consists two modals + success/pending/failure transaction modals. We display the `MapOperatorToStakingProviderModal` with the input form where the user can write down the operator address that he wants to map. This modal was heavily based on `ConfirmStakingParamsModal`. We validate it if it's not used by other staking providers and if we check if user already has operator mapped for one of the apps (if yest then we do the flow from the second point). Then we display confirmation modal (`MapOperatorToStakingProviderConfirmationModal`), and if user confirm we send two transaction using `useRegisterMultipleOperatorsTransaction` (which was heavily based on `useAuthorizeMultipleAppsTransaction`).

### flow for mapping one of the app if user rejects one of the transactions from the previous step

This is the flow in which user rejects one of the transaction from the previous flow. In such cases we display same modal (but little bit changed) and after confirmation we are sending one transaction using `useRegisterMultipleOperatorsTransaction` hook.

### simple modal queue

There are two modals that we have to display (if all conditions are met) after user connects the wallet - `MapOperatorToStakingProviderModal` and `NewAppsToAuthorizeModal`. Because of that I created a simple pseudo-queue for modal whichc is basically a state that have information if the current flow is finished:

```
export interface ModalQueueState {
  isSuccessfullLoginModalClosed: boolean
  isMappingOperatorToStakingProviderModalClosed: boolean
}
```

Thank to this we can display modals in this order: Successful log in modal -> `MapOperatorToStakingProviderModal` -> `NewAppsToAuthorizeModal`. We don't need a flag for the last modal since we don't have to display any other modals after that.

Screenshots:

<details>
  <summary>MapOperatorToStakingProviderModal for two unmapped apps</summary>

image

</details>

<details>
  <summary>MapOperatorToStakingProviderConfirmationModal for two unmapped apps</summary>

image

</details>

<details>
  <summary>MapOperatorToStakingProviderModal for one unmapped app</summary>

image

</details>

<details>
  <summary>MapOperatorToStakingProviderConfirmationModal for one unmapped app</summary>

image

</details>


src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { ModalType } from "../../../enums"
+import { openModal } from "../../../store/modal"
+import { updateState } from "../../../store/staking"
+import { FormValues } from "../../Forms"
+import { useModal } from "../../../hooks/useModal"
+import StakeAddressInfo from "../../../pages/Staking/StakeCard/StakeAddressInfo"
+import { useWeb3React } from "@web3-react/core"
+import { AddressZero } from "@ethersproject/constants"
+import { useThreshold } from "../../../contexts/ThresholdContext"
+import { isAddressZero, isSameETHAddress } from "../../../web3/utils"
```
r-czajkowski, line 34: Related to the comment above- let's add `AddressZero` here.

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        openModal(ModalType.TransactionFailed, {
+          error: new Error(
+            "Transaction rejected. You are required to map the Operator Address for both apps."
+          ),
+          closeModal: () => {
+            closeModal()
+            dispatch(mapOperatorToStakingProviderModalClosed())
+          },
+        })
+      }
```
r-czajkowski, line 114: we can pass onError to the `useRegisterOperatorTransaction`.

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
```diff
@@ -0,0 +1,174 @@
...
```
r-czajkowski, line 127: Same as above.

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+const MapOperatorToStakingProviderConfirmationModal: FC<
+  BaseModalProps & {
+    operator: string
+  }
+> = ({ operator, closeModal }) => {
+  const { account } = useWeb3React()
+  const { registerMultipleOperators } =
+    useRegisterMultipleOperatorsTransaction()
+  const dispatch = useDispatch()
```
r-czajkowski, line 68: Let's use `useAppDispatch`.

src/App.tsx
```diff
@@ -127,6 +128,7 @@ const AppBody = () => {
 
   useCheckBonusEligibility()
   useFetchStakingRewards()
+  useSaveConnectedAddressToStore()
```
r-czajkowski, line 131: See https://github.com/threshold-network/token-dashboard/pull/199#discussion_r983220907 I think this hook is unnecessary.

r-czajkowski (changes requested):

georgeweiler (approved):
Approving so we can merge before launch but we should remember to address the comments that were flagged for code quality

r-czajkowski (approved):

src/hooks/useRolesOf.ts
```diff
@@ -0,0 +1,11 @@
+import { useSelector } from "react-redux"
+import { RootState } from "../store"
+import { RolesOf } from "../threshold-ts/staking"
+
+export const useRolesOf: () => RolesOf = () => {
+  const rolesOf = useSelector(
```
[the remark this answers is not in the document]
michalsmiarowski: I know that TypeScript set those types automatically so you can mouse over the variable and see the type in your IDE, but wouldn't it be a good idea to keep the types wrote as well? I think this would help to understand code more when for example doing the "raw" review of the code on GItHub.

PR #206, opened 2022-09-28
Removes node statuses, APY and slashing percentages

georgeweiler:
We are not able to calculate these values yet so they should be removed.


r-czajkowski (approved):
LGTM!

michalsmiarowski (approved):
LGTM 🦖

PR #209, opened 2022-09-30
Updates announcement banner styles

georgeweiler:
The announcement banner component needs to be updated to match the design system: https://www.figma.com/file/zZi2fYDUjWEMPQJWAt8VWv/Threshold-DS?node-id=3234%3A23878


src/components/AnnouncementBanner/index.tsx
```diff
@@ -53,7 +55,10 @@ const AnnouncementBanner: FC<AnnouncementBannerProps & BoxProps> = ({
         sx={styles.subContainer}
       >
         <Image maxW={size == "sm" ? "146px" : "280px"} src={imgSrc} />
-        <H4 sx={styles.title}>{title}</H4>
+        <Stack>
+          <LabelMd sx={styles.preTitle}>{preTitle}</LabelMd>
```
r-czajkowski, line 59: `preTitle` is an optional prop so let's check if exists and then render the `LabelMd`.

src/theme/AnnouncementBanner.ts
```diff
@@ -17,6 +18,14 @@ const baseStyle: PartsStyleFunction<typeof parts> = (props) => {
...
       px: "16",
       mb: 4,
+      // TODO: update to gradient.4 when components PR is merged
+      // https://github.com/threshold-network/components/pull/25
+      bg: mode(
+        "linear-gradient(to bottom right, #F0E1FF, #D7C9FF)",
+
+        // TODO: Figure out if this is a new color in the DS (gray.1000 ?)
+        "#13171B"
+      )(props),
```
r-czajkowski, line 28: We need to improve styles for dark mode.

src/pages/Staking/HowItWorks/index.tsx
```diff
@@ -57,7 +57,6 @@ const HowItWorksPage: PageComponent = () => {
         buttonText="Start Staking"
         size="lg"
         hideCloseBtn
-        variant="secondary"
```
r-czajkowski, line 60: We should have variants or pass correctly props because the AnnacumentBanner on `How It Works page` is different than `Overview`.

r-czajkowski (changes requested):

src/components/AnnouncementBanner/index.tsx
r-czajkowski: 734e706

src/theme/AnnouncementBanner.ts
r-czajkowski: 899438a

src/pages/Staking/HowItWorks/index.tsx
r-czajkowski: My mistake- they are the same based on the TDS https://www.figma.com/file/zZi2fYDUjWEMPQJWAt8VWv/Threshold-DS?node-id=3234%3A24113

r-czajkowski (approved):

PR #220, opened 2022-09-30
Update auth/dauth form validation

r-czajkowski:
We need to update the validation for authorize/deauthorize form to match the validation on the smart contract side. There are 3 scenarios:

1. Application is unauthorized- first-time authorization:
- the authorization amount should be greater than the minimum authorization amount for a given application and less than/equal to the total staked balance of the stake.

2. Increase authorization- an application is already authorized:
- the increase amount should be greater than 0 and less than/equal to the remaining stake amount: Eg:
  - total staked balance in T: 100 000 T,
  - first authorization amount: 40 000 T,
  - max increase amount is 60 000 T.

3. Deatuhorization:
- the decrease amount should be greater than 0 and the authorization amount after deauthorization should be greater than/equal to the minimum authorization amount eg:
  - minimum authorization amount: 40 000 T,
  - authorization amount: 100 000 T,
  - the authorizer can deauthorize the staking application by up to 60 000 T. An authorizer can also fully deauthorize, so the value can be 100 000T.

Ref:
https://github.com/keep-network/keep-core/blob/4df5ceb01fd1bd2d55a39eb65db6c14d5a275d18/solidity/random-beacon/contracts/libraries/BeaconAuthorization.sol#L278-L281
https://github.com/keep-network/keep-core/blob/main/solidity/ecdsa/contracts/libraries/EcdsaAuthorization.sol#L281-L284
https://github.com/threshold-network/solidity-contracts/blob/main/contracts/staking/TokenStaking.sol#L1305-L1313

### DIFF
--- a/src/components/StakingApplicationForms/index.tsx
+++ b/src/components/StakingApplicationForms/index.tsx
@@ -1,0 +1,163 @@
+import { FC, Ref, useEffect, useState } from "react"
+import { BigNumber } from "ethers"
+import { FormikErrors, FormikProps, withFormik } from "formik"
+import { BodyMd, BodySm } from "@threshold-network/components"
+import {
+  TokenAmountFormBase,
+  TokenAmountFormBaseProps,
+  FormValues,
+} from "../Forms"
+import {
+  DEFAULT_MIN_VALUE,
+  getErrorsObj,
+  validateAmountInRange,
+} from "../../utils/forms"
+import { formatTokenAmount } from "../../utils/formatAmount"
+
+type ComponentProps = {
+  totalStake: string
+  authorizedAmount: string
+  isAuthorization?: boolean
+} & Omit<TokenAmountFormBaseProps, "label" | "maxTokenAmount">
+
+const Label: FC<{ label: string; remainigAmount: string }> = ({
+  label,
+  remainigAmount,
+}) => {
+  return (
+    <>
+      <BodyMd as="span" fontWeight="bold">
+        {label}
+      </BodyMd>
+      <BodySm as="span" color="gray.500" float="right">
+        Remaining Balance: {formatTokenAmount(remainigAmount)} T
+      </BodySm>
+    </>
+  )
+}
+
+const StakingApplicationFormBase: FC<
+  ComponentProps & FormikProps<FormValues>
+> = ({
+  totalStake,
+  submitButtonText,
+  isDisabled,
+  helperText,
+  authorizedAmount,
+  isAuthorization = true,
+  ...formikProps
+}) => {
+  const { values } = formikProps
+  const { tokenAmount } = values
+  const [remainigAmount, setRemainingAmount] = useState("0")
+  const [maxAmount, setMaxAmount] = useState("0")
+
+  useEffect(() => {
+    if (!isAuthorization) {
+      setMaxAmount(authorizedAmount)
+    } else {
+      setMaxAmount(BigNumber.from(totalStake).sub(authorizedAmount).toString())
+    }
+  }, [authorizedAmount, totalStake, isAuthorization])
+
+  useEffect(() => {
+    const _tokenAmount = BigNumber.from(tokenAmount || "0")
+    const _max = BigNumber.from(maxAmount)
+    setRemainingAmount(
+      tokenAmount
+        ? _tokenAmount.gte(_max)
+          ? "0"
+          : _max.sub(_tokenAmount).toString()
+        : maxAmount
+    )
+  }, [tokenAmount, authorizedAmount, totalStake, maxAmount])
+
+  return (
+    <TokenAmountFormBase
+      label={
+        <Label
+          label={isAuthorization ? "Increase Amount" : "Decrease Amount"}
+          remainigAmount={remainigAmount}
+        />
+      }
+      submitButtonText={submitButtonText}
+      isDisabled={isDisabled}
+      maxTokenAmount={totalStake}
+      placeholder={"Enter amount"}
+      helperText={helperText}
+      {...formikProps}
+    />
+  )
+}
+
+type StakingAppFormBaseProps = {
+  initialAmount?: string
+  minimumAuthorizationAmount: string
+  innerRef?: Ref<FormikProps<FormValues>>
+  onSubmitForm: (tokenAmount: string) => void
+} & ComponentProps
+
+const authorizationValidation = (
+  values: FormValues,
+  props: StakingAppFormBaseProps
+) => {
+  const errors: FormikErrors<FormValues> = {}
+
+  const { tokenAmount } = values
+  const { authorizedAmount, totalStake, minimumAuthorizationAmount } = props
+
+  const _authorizedAmount = BigNumber.from(authorizedAmount)
+  const _max = BigNumber.from(totalStake).sub(_authorizedAmount)
+  const _minimumAuthorizationAmount = BigNumber.from(minimumAuthorizationAmount)
+
+  const min = _authorizedAmount.gt(_minimumAuthorizationAmount)
+    ? DEFAULT_MIN_VALUE
+    : _minimumAuthorizationAmount.sub(_authorizedAmount)
+
+  errors.tokenAmount = validateAmountInRange(
+    tokenAmount,
+    _max.toString(),
+    min.toString()
+  )
+
+  return getErrorsObj(errors)
+}
+
+const deauthorizationValidation = (
+  values: FormValues,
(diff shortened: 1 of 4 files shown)


src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+        : maxAmount
+    )
+  }, [tokenAmount, authorizedAmount, totalStake, maxAmount])
+
+  return (
+    <TokenAmountFormBase
+      label={
+        <Label
+          label={isAuthorization ? "Increase Amount" : "Decrease Amount"}
+          remainigAmount={remainigAmount}
```
georgeweiler, line 80: typo

georgeweiler:
left 2 very small comments. OK to merge otherwise.

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+const Label: FC<{ label: string; remainigAmount: string }> = ({
+  label,
+  remainigAmount,
+}) => {
+  return (
+    <>
+      <BodyMd as="span" fontWeight="bold">
+        {label}
+      </BodyMd>
+      <BodySm as="span" color="gray.500" float="right">
```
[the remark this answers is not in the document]
r-czajkowski: As I can see on Figma https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6028%3A175671 the color is `gray.500` in dark mode as well.

src/components/StakingApplicationForms/index.tsx
r-czajkowski: 03163e6

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const max = BigNumber.from(authorizedAmount).sub(minimumAuthorizationAmount)
+  const _tokenAmount = BigNumber.from(tokenAmount)
+
+  if (!_tokenAmount.eq(authorizedAmount)) {
+    errors.tokenAmount = validateAmountInRange(
+      tokenAmount,
+      max.toString(),
+      DEFAULT_MIN_VALUE
+    )
+  }
```
michalsmiarowski, line 143: There is an error when you remove the value from the deauthorization form. It happend because `tokenAmount` is `undefined` so it can't be changed to `BigNumber`.

We have a check in `validateAmountInRange` that displays the "Required" message but here, since the valued is undefined, it doesn't go inside the if and the check is not done

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    const _tokenAmount = BigNumber.from(tokenAmount || "0")
+    const _max = BigNumber.from(maxAmount)
+    setRemainingAmount(
+      tokenAmount
+        ? _tokenAmount.gte(_max)
+          ? "0"
+          : _max.sub(_tokenAmount).toString()
+        : maxAmount
+    )
+  }, [tokenAmount, authorizedAmount, totalStake, maxAmount])
```
michalsmiarowski, line 73: Non-blocking:

Do we need `authorizedAmount` here since it's not used in `useEffect()`?

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+  props: StakingAppFormBaseProps
+) => {
+  const errors: FormikErrors<FormValues> = {}
+
+  const { tokenAmount } = values
+  const { authorizedAmount, totalStake, minimumAuthorizationAmount } = props
+
+  const _authorizedAmount = BigNumber.from(authorizedAmount)
+  const _max = BigNumber.from(totalStake).sub(_authorizedAmount)
+  const _minimumAuthorizationAmount = BigNumber.from(minimumAuthorizationAmount)
```
michalsmiarowski, line 111: Non-blocking:

I think we usually name BigNumber variables as `<variable_name>InBN` and not `_<variable_name>`. Check for example `validateAmountInRange` method in `utils/forms.ts`. I think we should keep the naming convention similar throught the whole project.

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    if (!isAuthorization) {
+      setMaxAmount(authorizedAmount)
+    } else {
+      setMaxAmount(BigNumber.from(totalStake).sub(authorizedAmount).toString())
+    }
+  }, [authorizedAmount, totalStake, isAuthorization])
+
+  useEffect(() => {
+    const _tokenAmount = BigNumber.from(tokenAmount || "0")
+    const _max = BigNumber.from(maxAmount)
```
michalsmiarowski, line 65: Non-blocking:

See above comment

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+const deauthorizationValidation = (
+  values: FormValues,
+  props: StakingAppFormBaseProps
+) => {
+  const errors: FormikErrors<FormValues> = {}
+
+  const { tokenAmount } = values
+  const { authorizedAmount, minimumAuthorizationAmount } = props
+  const max = BigNumber.from(authorizedAmount).sub(minimumAuthorizationAmount)
+  const _tokenAmount = BigNumber.from(tokenAmount)
```
michalsmiarowski, line 135: Non-blocking:

See above comment

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+      }
+      submitButtonText={submitButtonText}
+      isDisabled={isDisabled}
+      maxTokenAmount={totalStake}
+      placeholder={"Enter amount"}
+      helperText={helperText}
+      {...formikProps}
+    />
+  )
+}
```
michalsmiarowski, line 91: Non-blocking but I think it's pretty important:

Max button on deauthorization flow sets max amount of the token staked in that stake instead of max amount that the user can deauthorize.

image

I've had a 83,000 T stake and i authorized 42,000T. When i want to deauthorize and click `MAX` it sets the value to 83,000 and says `The vlaue should be less or equal than 2,000`. It should be set up to 42,000 T.

And also the validation message might be misleading, because it says that you can only deauthorize the value to less or equal 2,000 when in reality, in this case, you can deauthorize less or equal 2,000 or 42,000 T.

michalsmiarowski (changes requested):
Left some comments to look at before merge

src/components/StakingApplicationForms/index.tsx
r-czajkowski: 6910e97

r-czajkowski: Max button: 84ba3bd

src/components/StakingApplicationForms/index.tsx
r-czajkowski: 098b64f

src/components/StakingApplicationForms/index.tsx
r-czajkowski: 5054679

src/components/StakingApplicationForms/index.tsx
r-czajkowski: 5054679

src/components/StakingApplicationForms/index.tsx
r-czajkowski: 5054679

src/components/StakingApplicationForms/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  if (!tokenAmountInBN.eq(authorizedAmount)) {
+    errors.tokenAmount = validateAmountInRange(
+      tokenAmount,
+      max.toString(),
+      DEFAULT_MIN_VALUE,
+      {
+        ...defaultValidationOptions,
+        lessThanValidationMsg(amount) {
+          return `${defaultLessThanMsg(amount)} or ${formatTokenAmount(
+            authorizedAmount.toString()
```
michalsmiarowski, line 150: I think this would be more readable:

```suggestion
        lessThanValidationMsg(amount) {
          return `${defaultLessThanMsg(amount)} or equal to ${formatTokenAmount(
            authorizedAmount.toString()
```

In your case it might indicate that the value should be less than or equal `amount` or less than or equal `authorized` amount.

michalsmiarowski (approved):
LGTM 🌌

PR #221, opened 2022-09-30
Fixes authorizer input bug  that caused an app crash

georgeweiler:
The authorizer input would crash the app if the user changed the value because of the `isSameETHAddress` validation that we used.

I think we should think about impacts of adding this address validation to the `isSameETHAddress` method to avoid future bugs, but for now this will patch the issue and prepare us for the multi app staking launch


michalsmiarowski (approved):
LGTM 🐺

r-czajkowski (approved):

michalsmiarowski (approved):

PR #222, opened 2022-10-03
Update text color in alert for dark mode

r-czajkowski:
Update the text color in alert for dark mode on the authorization page.


r-czajkowski:
Fixed in #221

PR #223, opened 2022-10-03
Operator mapping - code cleaning

### LINKED ISSUE #232: Operator Mapping Modal Copy needs an update
Screen Shot 2022-10-03 at 9 32 55 AM

### LINKED ISSUE #255: Etherscan link incorrect for operator registration tx
After operator registration the link is: https://etherscan.io/tx/[object%20Object]

michalsmiarowski:
Closes #232 (64677b39e1b6c5234354bcdb5adb933bf3d191bf)
Closes #255 (ae4231d832c00e4cc353d633c1189e4e18ef0382)

# Operator mapping code cleaning

### Main changes:
- rearrange the operator mapping related data in redux store
- adds operator and staking provider addresses as props to the Operator Mapping related modals and `OperatorAddressMappingCard`
- removes unnecessary `shouldDisplayMapOperatorToStakingProviderModal` effect 
- fix some of the review comments from https://github.com/threshold-network/token-dashboard/pull/199
- fix a bug that caused the `MapOperatorToStakingProvider` modal to not be displayed when we changed accounts in MetaMask
- update text in operator mapping modal (Closes: #232)

TODO in the future:
- [ ] display Operator Mapping Card when the operator is mapped - https://github.com/threshold-network/token-dashboard/issues/259
- [ ] display Operator Mapping Card and the modals related to it after submitting a stake (without the need to refresh)

# Change details

### rearrange the operator mapping related data in redux store

Previously we keep the data needed to operator mapping in the redux store in this format:

(1) `rolesOf` property in `connected-account` slice:

```
{
address: string,
  rolesOf: {
    owner: string,
    authorizer: string,
    beneficiary: string, 
}
```

This indicated if the currently logged account is used as a staking provider in other stakes. We fetched the data from the contract.

(2) `mappedOperator` property in `staking-applications` slice:
```
{
  tbtc: {
    parameter: (...),
    stakingProviders: (...),
    mappedOperator: {
      isInitialFetchDone: boolean,
      isFetching: boolean,
      error: string,
      data: string,
  },
  randomBeacon: {
    parameter: (...),
    stakingProviders: (...),
    mappedOperator: {
      isInitialFetchDone: boolean,
      isFetching: boolean,
      error: string,
      data: string,
  }
}
```

This stored the data related to currently logged staking provider and it's mapped operator for each application. This could wrongly indicate that the data is more related to the staking app itself (since it was in the `staking-application` slice) when in reality it was more related to currently connected account.

At first the idea was to not store this data in redux store at all, but unfrotunately this didn't went as smooth as I thogught and I think we must store them mainly to properly display `OperatorAddressMappingCard` in the staking page for all staking providers that don't have the operator mapped. I've encoutered problems with implementing it without using the store when the staking provider !=== owner.

This is why i decided to go back to keeping the data in the redux store (you can see all this in the commit history) but in better (and i think simpler) form.

Right now we will store this data in `account` slice and we ditched the whole `rolesOf` property which was object, to just `isStakingProvider` which is boolean value. After the changes the `connected-account` is renamed to `account` and looks like this:

```
{
  address: string,
  isStakingProvider: boolean,
  operatorMapping: {
    data: {
      tbtc: string,
      randomBeacon: string,
    },
    isFetching: boolean,
    isInitialFetchDone: boolean,
  },
}
```

### adds operator and staking provider addresses as props to the Operator Mapping related modals and `OperatorAddressMappingCard`

For these components we are adding a props that contains operator address and staking provider address:
- `OperatorAddressMappingCard`
- `MapOperatorToStakingProviderModal`
- `MapOperatorToStakingProviderSuccessModal`
- `MapOperatorToStakingProviderConfirmationModal`
- `MapOperatorToStakingProviderForm`

### removes unnecessary shouldDisplayMapOperatorToStakingProviderModal effect

This effect was unnecessary. Instead of using `predicate` (from redux-toolkit) when calling `displayMapOperatorToStakingProviderModalEffect` we now use `actionCreator` (also from redux-toolkit) to call this effect on `operatorMappingInitialFetchDone` action call which in turn is called by `setStakes` action. This way we are sure we have all needed that in redux store and we can use them in ` displayMapOperatorToStakingProviderModalEffect`

### fix some of the review comments from https://github.com/threshold-network/token-dashboard/pull/199

Most comment were either fixed by itself during the refactor process (the part of code that was commented does no longer exist) and other ones I've tried to fix. For some I've added comments there to start a discussion.

### fix a bug that caused the `MapOperatorToStakingProvider` modal to not be displayed when we changed accounts in MetaMask

We are setting the `isSuccessfullLoginModalClosed` flag to false when resetting the store now, because the reset happend when we change the accounts and the login modal is not displayed in such case. We also set `isMappingOperatorToStakingProviderModalClosed` flag back to false in this case.

**Additional note**: The modal queue was also moved in the store to `modal` prop since it's related to modal. In the furue we are planning to create a real modal queue.

### update text in operator mapping modal 

Update text based on Figma designs in `OperatorAddressMappingCard` component and in `MapOperatorToStakingProviderModal`. For more details see #232


src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
```diff
@@ -60,17 +54,27 @@ const OperatorMappingConfirmation: FC<
...
   const { registerMultipleOperators } =
     useRegisterMultipleOperatorsTransaction()
-  const dispatch = useDispatch()
+  const dispatch = useAppDispatch()
+
+  const isOperatorMappedOnlyInTbtc =
+    !isAddressZero(mappedOperatorTbtc) &&
+    isAddressZero(mappedOperatorRandomBeacon)
 
-  const operatorMappedToStakingProviderHelpers =
```
georgeweiler, line 70: why remove this helper hook? seems like it saved some duplicate code. What if we added the tbtc & random beacon addresses as parameters so we don't have to rewrite the logic?

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -69,14 +73,14 @@ const MapOperatorToStakingProviderForm = withFormik<
         }
         if (
           isOperatorMappedOnlyInRandomBeacon &&
-          values.operator !== operatorMappedRandomBeacon
+          values.operator !== mappedOperatorRandomBeacon
         ) {
           validationMsg =
             "The operator address doesn't match the one used in tbtc app"
```
georgeweiler, line 79: let's use tBTC

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -37,43 +37,50 @@ const MapOperatorToStakingProviderSuccessBase: FC<
       body={
         <>
           <List spacing="2" mb={"4rem"}>
```
georgeweiler, line 39: minor suggestion, but can we use `16` instead of `4rem` since we have to touch this file to update the typo below anyway

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -37,43 +37,50 @@ const MapOperatorToStakingProviderSuccessBase: FC<
...
-                <BodySm>Operator</BodySm>
+                <BodySm>Provider Address</BodySm>
                 <BodySm>
-                  {shortenAddress(transactions[0].application.operator)}
+                  {shortenAddress(transactions[0].application.stakingProvider)}
                 </BodySm>
               </HStack>
             </ListItem>
-            <ListItem key="map_operator_succes_modal__staking_provider">
+            <ListItem key="map_operator_succes_modal__operator">
```
georgeweiler, line 48: same as above

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -37,43 +37,50 @@ const MapOperatorToStakingProviderSuccessBase: FC<
...
-              >
-                View{" "}
+          <BodySm align="center" mt={"1"}>
+            {transactions.length === 1 ? (
+              <>
                 <ViewInBlockExplorer
-                  text={text}
-                  id={transaction.txHash}
+                  text="View"
+                  id={`map_operator_transaction_${transactions[0].txHash}`}
```
georgeweiler, line 62: how does this work? Doesn't the `ViewInBlockExplorer` pass this `id` along as the hash to create the link?

src/hooks/staking-applications/useRegisterMultipleOperatorsTransaction.ts
```diff
@@ -23,25 +29,27 @@ export const useRegisterMultipleOperatorsTransaction = () => {
...
+
         if (
-          !isAddressZero(operatorMappedRandomBeacon) &&
-          !isAddressZero(operatorMappedTbtc)
+          !isAddressZero(mappedOperatorRandomBeacon) &&
+          !isAddressZero(mappedOperatorTbtc)
         )
           throw new Error("Both apps already have mapped operator!")
 
+        const isOperatorMappedOnlyInTbtc =
```
georgeweiler, line 45: see my comment about about the helper hook that we removed. I think it could be nice to consolidate

src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -8,19 +8,34 @@ import {
...
 import { LabelSm } from "@threshold-network/components"
+import { FC } from "react"
 import { ModalType } from "../../enums"
-import { useOperatorMappedtoStakingProviderHelpers } from "../../hooks/staking-applications/useOperatorMappedToStakingProviderHelpers"
+// import { useOperatorMappedtoStakingProviderHelpers } from "../../hooks/staking-applications/useOperatorMappedToStakingProviderHelpers"
 import { useModal } from "../../hooks/useModal"
+import { isAddressZero } from "../../web3/utils"
 
-const OperatorAddressMappingCard = () => {
+//TODO: Fix this
```
georgeweiler, line 17: Could we leave a better explanation for what's wrong & how we should fix? Or maybe just fix now? tbh I don't see any issues, so maybe this comment can be deleted?

src/store/connected-account/connectedAccountSlice.ts
```diff
@@ -51,21 +106,29 @@ export const connectedAccountSlice = createSlice({
...
     )
   },
 })
 
-startAppListening({
-  actionCreator: setStakes,
-  effect: getRolesOf,
-})
+export const registerConnectedAccountListeners = () => {
+  if (featureFlags.MULTI_APP_STAKING) {
```
georgeweiler, line 117: maybe we don't need this flag anymore since we launched MAS.

src/store/index.ts
```diff
@@ -53,6 +55,15 @@ const rootReducer: Reducer = (state: RootState, action: AnyAction) => {
         T: { ...state.token.T, balance: 0 },
         TBTC: { ...state.token.TBTC, balance: 0 },
       },
+      // we don't display successful login modal when changin account so we are
+      // setting the isSuccessfullLoginModalClosed flag to true and also
+      // isMappingOperatorToStakingProviderModalClosed flag back to false
+      modal: {
+        modalQueue: {
+          isSuccessfullLoginModalClosed: true,
```
georgeweiler, line 63: small typo should be `isSuccessfulLoginModalClosed`

georgeweiler (changes requested):
Overall the state changes look good! I'm not a fan of the modalQueue as it is implemented right now. I think we should prioritize a real queue so we aren't making specific exceptions in the `closeModal` method.  I think this is non-blocking though because it's a larger update that we can do in a future PR.

I left a few small nits that we should cover before merging

src/store/staking-applications/effects.ts
```diff
@@ -168,114 +170,54 @@ const getKeepStakingAppStakingProvidersData = async (
...
+  if (!modalQueue.isSuccessfullLoginModalClosed) {
+    await listenerApi.condition((action, currentState: RootState) => {
+      return currentState.modal.modalQueue.isSuccessfullLoginModalClosed
+    })
+  }
   const { address } = connectedAccount
-  if (address) {
-    // check if the current connected address is used somewhere as a staking
-    // provider
+  if (address && connectedAccount.operatorMapping.isInitialFetchDone) {
```
r-czajkowski, line 187: Let's remove this nested if for better readability:
```ts
if(!address || !connectedAccount.operatorMapping.isInitialFetchDone) return

...rest here
```

r-czajkowski, line 187: Do we need this condition `connectedAccount.operatorMapping.isInitialFetchDone`? We run this effect on `operatorMappingInitialFetchDone` so we can assume it is set to `true`.

src/store/staking-applications/effects.ts
```diff
@@ -168,114 +170,54 @@ const getKeepStakingAppStakingProvidersData = async (
...
   action: AnyAction,
   listenerApi: AppListenerEffectAPI
 ) => {
-  const { connectedAccount } = listenerApi.getState()
+  const {
+    modal: { modalQueue },
+  } = listenerApi.getState()
+  const { connectedAccount, staking } = listenerApi.getState()
+  if (!modalQueue.isSuccessfullLoginModalClosed) {
+    await listenerApi.condition((action, currentState: RootState) => {
```
r-czajkowski, line 182: Do we need the `RootState` type here? The `listenerApi` has our custom type with our state type so I believe it's unnecessary.

src/store/index.ts
```diff
@@ -17,8 +17,10 @@ import {
...
 import { listenerMiddleware } from "./listener"
-import { connectedAccountSlice } from "./connected-account"
-import { modalQueueSlice } from "./modalQueue"
+import {
+  connectedAccountSlice,
+  registerConnectedAccountListeners,
+} from "./connected-account"
 
 const combinedReducer = combineReducers({
   connectedAccount: connectedAccountSlice.reducer,
```
r-czajkowski, line 26: Can we rename to `account/user/app`?

src/store/connected-account/index.ts
```diff
@@ -1,2 +1 @@
 export * from "./connectedAccountSlice"
```
r-czajkowski, line 1: Same as above.

src/store/connected-account/effects.ts
```diff
@@ -1,37 +1,64 @@
...
+      )
+
+      listenerApi.dispatch(
+        setMappedOperator({
+          appName: "randomBeacon",
+          operator: mappedOperators.randomBeacon,
+        })
+      )
+
+      listenerApi.dispatch(operatorMappingInitialFetchDone())
```
r-czajkowski, line 54: I believe this action is unnecessary. We can set a flag to true in `setMappedOperator` slice.

src/store/connected-account/effects.ts
```diff
@@ -1,37 +1,64 @@
...
   listenerApi: AppListenerEffectAPI
 ) => {
   try {
     const { connectedAccount } = listenerApi.getState()
     const { address } = connectedAccount
-
-    if (!address) return
-    listenerApi.unsubscribe()
     const { owner, authorizer, beneficiary } =
       await listenerApi.extra.threshold.staking.rolesOf(address)
```
r-czajkowski, line 20: This effect will be triggered on `setStakes` action so it means we should have available stakes if the user is an owner. We can check if the stake where `stakingprovider === connected account` exists in the `stakes` array. If that stake exists there is no need to make the `rolesOf` request.

src/store/connected-account/effects.ts
```diff
@@ -1,37 +1,64 @@
...
-      isAddressZero(beneficiary)
-    ) {
-      listenerApi.dispatch(mapOperatorToStakingProviderModalClosed())
+
+    const isUsedAsStakingProvider =
+      !isZeroAddress(owner) &&
+      !isZeroAddress(authorizer) &&
+      !isZeroAddress(beneficiary)
+
+    if (isUsedAsStakingProvider) {
```
r-czajkowski, line 27: Can we `if(!isUsedAsStakingProvider) return`? There is a lot of code inside the `if` statement so IMO it will improve readability.

src/store/connected-account/connectedAccountSlice.ts
```diff
@@ -1,27 +1,42 @@
...
   name: "connected-account",
   initialState: {
     address: "",
-    rolesOf: {
-      owner: "",
-      authorizer: "",
-      beneficiary: "",
+    operatorMapping: {
+      data: {
+        isUsedAsStakingProvider: false,
```
r-czajkowski, line 32: I think we should't keep it in the `operatorMapping` object. It's not only related to the operator mapping it could be useful in other places. So I would keep it outside the `operatorMapping` object.
```
initialState: {
 address: string,
 isStakingProvider: boolean
}
```

src/store/connected-account/connectedAccountSlice.ts
```diff
@@ -31,13 +46,53 @@ export const connectedAccountSlice = createSlice({
...
+      const { isFetching } = action.payload
+      state.operatorMapping.isFetching = isFetching
+    },
+    operatorMappingInitialFetchDone: (
+      state: ConnectedAccountState,
+      action: PayloadAction
+    ) => {
+      state.operatorMapping.isFetching = false
+      state.operatorMapping.isInitialFetchDone = true
+    },
```
r-czajkowski, line 78: Can't we set the `isInitialFetchDone` and `isFetching` in the `setMappedOperator`?

src/store/connected-account/connectedAccountSlice.ts
```diff
@@ -1,27 +1,42 @@
...
+  operatorMapping: FetchingState<{
+    isUsedAsStakingProvider: boolean
+    mappedOperators: {
+      tbtc: string
+      randomBeacon: string
+    }
+  }>
 }
 
 export const connectedAccountSlice = createSlice({
```
r-czajkowski, line 26: Let's rename to `account/user/app` as I mentioned somewhere above.

src/pages/Staking/index.tsx
```diff
@@ -47,32 +46,10 @@ const StakingPage: PageComponent = (props) => {
...
-    beneficiary,
-    operatorMappedRandomBeacon,
-    operatorMappedTbtc,
-    isInitialFetchDone,
-    isEmptyOrZeroAddress,
-  ])
+  const {
+    isInitialFetchDone: isOperatorMappingInitialFetchDone,
+    data: { isUsedAsStakingProvider, mappedOperators },
+  } = useSelector((state: RootState) => state.connectedAccount.operatorMapping)
```
r-czajkowski, line 52: Let's use `useAppSelector`.

src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -8,19 +8,34 @@ import {
   HStack,
 } from "@threshold-network/components"
 import { LabelSm } from "@threshold-network/components"
+import { FC } from "react"
 import { ModalType } from "../../enums"
-import { useOperatorMappedtoStakingProviderHelpers } from "../../hooks/staking-applications/useOperatorMappedToStakingProviderHelpers"
+// import { useOperatorMappedtoStakingProviderHelpers } from "../../hooks/staking-applications/useOperatorMappedToStakingProviderHelpers"
```
r-czajkowski, line 13: Unnecessary import ?

src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -8,19 +8,34 @@ import {
   HStack,
 } from "@threshold-network/components"
 import { LabelSm } from "@threshold-network/components"
```
r-czajkowski, line 10: We already import components from `@threshold-network/components` so let's import from @threshold-network/components` only once.

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -37,43 +37,50 @@ const MapOperatorToStakingProviderSuccessBase: FC<
...
                   type={ExplorerDataType.TRANSACTION}
                 />{" "}
+                transaction on Etherscan
+              </>
+            ) : (
+              <>
+                View{" "}
+                {transactions.map((_, index) => (
+                  <Fragment key={`map_operator_transaction_${index}`}>
```
r-czajkowski, line 71: We could use `_.txHash` as a key.

georgeweiler:
@michalsmiarowski can you take on https://github.com/threshold-network/token-dashboard/issues/232 in this PR as well? Should be a simple commit.

src/store/connected-account/connectedAccountSlice.ts
michalsmiarowski: I think in this case we can't because the `operatorMapping` property contains:
- `isUsedAsStakingProvider` boolean value
- `mappedOperators` object
and `isInitialFetchDone` and `isFetching` is associated with the `operatorMapping` object as a whole, and `setMappedOperator` sets only `mappedOperators` and not fully - only for specific app. So we can't say the initial fetch is fully done.

src/store/connected-account/effects.ts
michalsmiarowski: I think it's not possible that way. See https://github.com/threshold-network/token-dashboard/pull/223#discussion_r990523464

src/store/connected-account/connectedAccountSlice.ts
michalsmiarowski: I think this one is also related to https://github.com/threshold-network/token-dashboard/pull/223#discussion_r990523464

I need to know that `isStakingProvider` is fully fetched (and same with mapped operators). This way I can print the Operator Address Mapping Card (PR for it's refactor here - https://github.com/threshold-network/token-dashboard/pull/261) in Staking Page without if displaying for one sec at the beginning (when `isStakingProvider` is set to `true` and the opreator addresses are still fetching). 

But tbh now when I think of it I think we could just check if the mapping operators are initially fetched and that should also work without any problems 🤔

src/store/index.ts
michalsmiarowski: Fixed in 47639bf858c8fb04be780e3eed636d8e742526b2

src/store/connected-account/index.ts
michalsmiarowski: Fixed in 47639bf858c8fb04be780e3eed636d8e742526b2

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -37,43 +37,50 @@ const MapOperatorToStakingProviderSuccessBase: FC<
       body={
         <>
           <List spacing="2" mb={"4rem"}>
-            <ListItem key="map_operator_succes_modal__operator">
+            <ListItem key="map_operator_succes_modal__staking_provider">
```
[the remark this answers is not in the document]
michalsmiarowski: Fixed in 45a5009ea7eb56c44fdb0e2846bb33c11a955c07

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
michalsmiarowski: Fixed in 0de4912594f5fd6521aa3345b3308a75abcdd607

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
michalsmiarowski: Great catch! I've mistaken it with a `key` prop 🤦‍♂️ Fixed in ae4231d832c00e4cc353d633c1189e4e18ef0382

src/pages/Staking/OperatorAddressMappingCard.tsx
michalsmiarowski: This was fixed earlier and i forgot to remove the TODO. Removed it in 08e6e157803311e44c512bc77cce4c93fedd31dd

src/store/connected-account/connectedAccountSlice.ts
michalsmiarowski: I'll address it in separate PR. Let's keep it for a while after release just in case something breaks after newest changes.

src/store/index.ts
michalsmiarowski: FIxed in d35e967400e6c49cf6c8b8f23ddb6667fb84f61d

src/store/staking-applications/effects.ts
michalsmiarowski: Fixed in 88603134d545e9164e5c2de371435ea44441091d

michalsmiarowski: Fixed in 88603134d545e9164e5c2de371435ea44441091d

src/store/staking-applications/effects.ts
michalsmiarowski: Fixed in 530837e2820eff1546685b441fe3dddd7c680b02

src/store/connected-account/effects.ts
michalsmiarowski: Fixed in d3c637b990b9f530b560e94054ff0dd3261d06f9

I couldn't get the stakes from the state using already created selector because even though the effect is called on `setStakes` action they are not set in state yet. Fortunately all the stakes are transferred via payload so that's how i search i any of the stakes there have `stakingProvider === connected account`.

src/store/connected-account/effects.ts
michalsmiarowski: Fixed in 45c0b72fa274f3314d3c3bd6e471aa5a49a53475

src/store/connected-account/connectedAccountSlice.ts
michalsmiarowski: Fixed in 47639bf858c8fb04be780e3eed636d8e742526b2

src/pages/Staking/index.tsx
michalsmiarowski: Fixed in 4fd84b4c7f1e0a5910ec6a557540305d823d1789

src/pages/Staking/OperatorAddressMappingCard.tsx
michalsmiarowski: Fixed in 369c9c908018a9448a3c5a543197d6d8e9bba994

src/pages/Staking/OperatorAddressMappingCard.tsx
michalsmiarowski: Fixed in 369c9c908018a9448a3c5a543197d6d8e9bba994

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
michalsmiarowski: Fixed in cdf31809bb34c6f8ea30553d4a91ec44644c5010

src/store/index.ts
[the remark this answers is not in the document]
michalsmiarowski: Actually it's not dispatched 🤔 And because of that I'm thinking how to resolve this issue. `resetStoreAction` is aclled only when we change the metamas account while we have the account connected. It is not dispatched when we disconnect and then connect again.

michalsmiarowski:
> @michalsmiarowski can you take on #232 in this PR as well? Should be a simple commit.

Fixed in 64677b39e1b6c5234354bcdb5adb933bf3d191bf

src/store/index.ts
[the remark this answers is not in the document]
r-czajkowski: It's because the injected connector doesn't emit the `Deactivated` event in `deactivate` function https://github.com/Uniswap/web3-react/blob/v6/packages/injected-connector/src/index.ts#L181-L188 but the WalletConnect connector emits this event see https://github.com/Uniswap/web3-react/blob/v6/packages/walletconnect-connector/src/index.ts#L136. So maybe we could override the `deactivate` in InjectedConnector in this way:
```ts
class CustomInjectedConnector extends InjectedConnector {
  deactivate = (): void => {
    super.deactivate()
    this.emitDeactivate()
  }
}

const injectedConnector = new CustomInjectedConnector({})
```

We probably should do the same for `WalletLinkConnector` in `web3/connectors/coinbaseWallet.ts`.

src/store/connected-account/connectedAccountSlice.ts
r-czajkowski: > But tbh now when I think of it I think we could just check if the mapping operators are initially fetched and that should also work without any problems

yeah, I think the same. Also the `isStakingProvider` flag will be `false` by default, so to display operator mapping card we will have a condition like this `isStakingProvider && isInitialFetchDone`.

r-czajkowski: If we move the `isStakingProvider` outside the `operatorMapping` object can we also simplify to:
```
operatorMapping: FetchingState<Record<StakingAppName, string>>
```

I believe the `mappedOperators` is unnecessary because it's inside the `operatorMapping`- `operatorMapping` is clear enough and will be simpler
```
state.operatorMapping.data.tbtc/randomBeacon
vs
state.operatorMapping.data.mappedOperator.tbtc/randomBeacon
```

src/store/account/effects.ts
```diff
@@ -0,0 +1,78 @@
+import { isZeroAddress } from "ethereumjs-util"
```
r-czajkowski, line 1: Let's import from `web3/utils`

src/store/account/slice.ts
```diff
@@ -0,0 +1,131 @@
...
+          tbtc: AddressZero,
+          randomBeacon: AddressZero,
+        },
+      },
+      isFetching: false,
+      isInitialFetchDone: false,
+    },
+  } as AccountState,
+  reducers: {
+    setAccountAddress: (state: AccountState, action: PayloadAction<string>) => {
```
r-czajkowski, line 43: maybe `walletConnected`?

r-czajkowski (changes requested):

src/store/account/slice.ts
```diff
@@ -0,0 +1,131 @@
...
+      state.operatorMapping.data.mappedOperators[appName] = operator
+    },
+  },
+  extraReducers: (builder) => {
+    builder.addMatcher(
+      (action: AnyAction) =>
+        action.type.match(providerStakedForStakingProvider),
+      (state, action: ReturnType<typeof providerStaked>) => {
+        const { owner, beneficiary, authorizer, stakingProvider } =
+          action.payload
```
r-czajkowski, line 101: Let's just `const { stakingProvider } = action.payload`

r-czajkowski (changes requested):

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -69,17 +73,17 @@ const MapOperatorToStakingProviderForm = withFormik<
         }
         if (
           isOperatorMappedOnlyInRandomBeacon &&
-          values.operator !== operatorMappedRandomBeacon
+          values.operator !== mappedOperatorRandomBeacon
```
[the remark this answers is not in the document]
michalsmiarowski: Fixed in 7ed1d17e9e0707d79ad3b49c2ef80816c2c48092

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -45,15 +46,18 @@ const MapOperatorToStakingProviderForm = withFormik<
...
+    const isOperatorMappedOnlyInTbtc =
+      !isAddressZero(mappedOperatorTbtc) &&
+      isAddressZero(mappedOperatorRandomBeacon)
+
+    const isOperatorMappedOnlyInRandomBeacon =
+      isAddressZero(mappedOperatorTbtc) &&
+      !isAddressZero(mappedOperatorRandomBeacon)
     const errors: FormikErrors<MapOperatorToStakingProviderFormValues> = {}
 
     errors.operator = validateETHAddress(values.operator)
```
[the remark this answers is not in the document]
michalsmiarowski: Fixed in 7ed1d17e9e0707d79ad3b49c2ef80816c2c48092

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -36,44 +29,51 @@ const MapOperatorToStakingProviderSuccessBase: FC<
       subTitle="You successfully mapped your Operator Address."
       body={
         <>
-          <List spacing="2" mb={"4rem"}>
-            <ListItem key="map_operator_succes_modal__operator">
+          <List spacing="2" mb={"16"}>
+            <ListItem key="map_operator_success_modal__staking_provider">
```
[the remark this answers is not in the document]
michalsmiarowski: Fixed in 676058f0e90ae34a6d3350b34261fcc5fb1524be

src/components/Modal/MapOperatorToStakingProviderSuccessModal/index.tsx
```diff
@@ -36,44 +29,51 @@ const MapOperatorToStakingProviderSuccessBase: FC<
...
-                <BodySm>Operator</BodySm>
+                <BodySm>Provider Address</BodySm>
                 <BodySm>
-                  {shortenAddress(transactions[0].application.operator)}
+                  {shortenAddress(transactions[0].application.stakingProvider)}
                 </BodySm>
               </HStack>
             </ListItem>
-            <ListItem key="map_operator_succes_modal__staking_provider">
+            <ListItem key="map_operator_success_modal__operator">
```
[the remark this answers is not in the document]
michalsmiarowski: Fixed in 676058f0e90ae34a6d3350b34261fcc5fb1524be

src/store/account/effects.ts
michalsmiarowski: Fixed in 009344de756a6593c914dec921f15bdd2ce13b55

src/store/account/slice.ts
michalsmiarowski: Fixed in c5453fdba5eb88c51259e8c60a66bc010648f832

src/store/account/slice.ts
michalsmiarowski: Fixed in 2fd71544f69bea87569f98eb62149aa93f710ee2

src/store/connected-account/connectedAccountSlice.ts
michalsmiarowski: Fixed in f754bc4f0a0f1f626255c1279d8f7ccdfca703c8 and 390eaea8ac74313b2a48458ab44c8b234971ef15

src/store/connected-account/connectedAccountSlice.ts
michalsmiarowski: Fixed in 7cf4736f217ed8dfb9a84c0fdc262a242f064d7d

src/store/index.ts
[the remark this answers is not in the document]
michalsmiarowski: This does not work for me. Even though the `deactivate` event is emited it is not catched in the `AppBody`'s `useEffect`. It's probably because the events are cleared/removed before the event is fired 🤔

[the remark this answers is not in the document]
r-czajkowski: Let's address in a separate PR.

r-czajkowski (approved):
LGTM! :shipit:

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
michalsmiarowski: This will be addressed in https://github.com/threshold-network/token-dashboard/pull/278

r-czajkowski (approved):

PR #225, opened 2022-10-03
Fix displaying new apps to authorize modal

michalsmiarowski:
The New Apps To Authorize modal was not displayed for the owner that was not a staking provider. This commit fixes that by dispatching the event that set the mapOperatorToStakingPprovitedModal as closed so the condition that displays the authorize new apps modal is met.


michalsmiarowski:
Let's make sure this bug is not happening in this PR - https://github.com/threshold-network/token-dashboard/pull/223

r-czajkowski (approved):

PR #233, opened 2022-10-03
Staking app card updates

r-czajkowski:
This PR fixes the `FilterTab` component in the staking application card. We should reset the form if we switch between tabs. The `Increase` tab should be disabled if there is a pending deauthorization. The `Decrease` tab should be disabled if the authorized stake is `0`-when a user has not yet authorized the staking app.


michalsmiarowski (approved):
LGTM 🦝

PR #260, opened 2022-10-05
Posthog

r-czajkowski:
This PR adds posthog data tracker. Posthog captures product's usage data to see which users are doing what in your application. We want to run the posthog only in production.


src/posthog/index.ts
```diff
@@ -0,0 +1,30 @@
+import posthog from "posthog-js"
+
+const POSTHOG_API_KEY = "phc_5XMnBAeMVxZQhmBm6G0Kee5ujBOiUhGYIakUKfDhgs5"
+const POSTHOG_API_HOST = "https://app.posthog.com"
+
+export const init = () => {
+  posthog.init(POSTHOG_API_KEY, {
+    api_host: POSTHOG_API_HOST,
+    // T dapp is a single page app so we need to make a `pageview` call
+    // manually. It also prevents sending the `pageview` event on postohog
```
georgeweiler, line 10: I think we should use `persistence: "memory"` configuration here because we don't want to store cookies on the user's browser. See: https://posthog.com/tutorials/cookieless-tracking

r-czajkowski: Good catch! Thanks George! 🦅 👁️

georgeweiler (approved):
lgtm

src/posthog/index.ts
```diff
@@ -0,0 +1,31 @@
+import posthog from "posthog-js"
+
+const POSTHOG_API_KEY = "phc_5XMnBAeMVxZQhmBm6G0Kee5ujBOiUhGYIakUKfDhgs5"
+const POSTHOG_API_HOST = "https://app.posthog.com"
```
michalsmiarowski, line 4: Shouldn't we move these to `.env` file?

src/hooks/posthog/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./usePosthog"
```
michalsmiarowski, line 1: Non-blocking but I think it would be benefical:

Posthog can be seen as a pretty separated feature which is not really dependent on any other features in our dashboard. So I think that the best thing would be to keep all the code in one folder - `stc/posthog`. I would move all the hooks to the `/src/posthog/hooks` directory instead of keeping them in `src/hooks/posthog`. 

Anothre argument for this option is that we don not keep all the hooks in `src/hooks` file - for `web3` we keep them in `src/web3/hooks`.

src/posthog/index.ts
```diff
@@ -0,0 +1,31 @@
...
+// one-page app, this means it will only send a `pageview` once, when the app
+// loads. To make sure any navigating a user does within your app gets captured,
+// you must make a `pageview` call manually.
+export const capturePageview = () => {
+  posthog.capture("$pageview")
+}
+
+export const reset = () => {
+  posthog.reset()
+}
```
michalsmiarowski, line 31: Non-blocking:

Coudl we create a class for that maybe? Instead of separating it to just functions  and the `import * as posthog`?

src/hooks/posthog/useCapturePageview.ts
```diff
@@ -0,0 +1,12 @@
...
+import { useLocation } from "react-router-dom"
+import * as posthog from "../../posthog"
+import { featureFlags } from "../../constants"
+
+export const useCapturePageview = () => {
+  const location = useLocation()
+
+  useEffect(() => {
+    if (featureFlags.POSTHOG) posthog.capturePageview()
+  }, [location])
```
michalsmiarowski, line 11: I think `capturePageView()` does not work as intended in some of our cases. For example - when we click `Staking` page on the left menu, this `useEffect` fires two times for two pathnames:

pathname 1: `/staking/how-it-works`
pathname 2: `/staking/how-it-works/overview`

Then in the posthog dashboard it is showed as:

image

So it looks that user went into the same page two times, when in reality he visited it once. We should either try to reduce this event to one in such cases or make sure it catches the url correcttly.

michalsmiarowski, line 11: I think we should specify the property of the object in the dependency array instead of passing the whole object. We only need pathname so I think that should work:

```suggestion
  useEffect(() => {
    if (featureFlags.POSTHOG) posthog.capturePageview()
  }, [location.pathname])
```

michalsmiarowski (changes requested):
Left some more comments to look at before the merge:

.env.production
```diff
@@ -4,3 +4,4 @@ REACT_APP_ETH_HOSTNAME_WS=$ETH_HOSTNAME_WS
 
 REACT_APP_FEATURE_FLAG_TBTC_V2=false
 REACT_APP_FEATURE_FLAG_MULTI_APP_STAKING=true
+REACT_APP_FEATURE_FLAG_POSTHOG=$POSTHOG_SUPPORT
```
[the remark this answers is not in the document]
r-czajkowski: I was thinking about it as additional protection for running the dapp locally. I mean we only want to run the T dapp with posthog on prodaction.

src/hooks/posthog/useIdentify.ts
```diff
@@ -0,0 +1,50 @@
...
+      const account = await connector?.getAccount()
+      if (account) {
+        posthog.identify(getAddress(account))
+      }
+    }
+    onLogin()
+  }, [connector])
+
+  useEffect(() => {
+    if (!featureFlags.POSTHOG) return
```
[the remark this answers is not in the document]
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/260#discussion_r990979213

src/hooks/posthog/useIdentify.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ConnectorEvent, ConnectorUpdate } from "@web3-react/types"
+import * as posthog from "../../posthog"
+import { getAddress, isSameETHAddress } from "../../web3/utils"
+import { featureFlags } from "../../constants"
+
+export const useIdentify = () => {
+  const { connector, account } = useWeb3React()
+
+  useEffect(() => {
+    if (!featureFlags.POSTHOG) return
```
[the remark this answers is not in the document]
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/260#discussion_r990979213

src/App.tsx
```diff
@@ -137,6 +138,7 @@ const AppBody = () => {
     dispatch(fetchETHPriceUSD())
   }, [dispatch])
 
+  usePosthog()
```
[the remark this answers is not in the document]
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/260#discussion_r990979213

src/posthog/index.ts
r-czajkowski: Yeah, we can move to the `.env` file.

src/hooks/posthog/index.ts
r-czajkowski: It actually depends on other features like web3 react and react-router. Posthog hooks can't exist w/o other libs. TBH I don't have a strong opinion on that, both places work for me.

src/hooks/posthog/useCapturePageview.ts
r-czajkowski: Yeah I was thinking about it but it only returns the url w/o `?` and `#` see https://github.com/remix-run/history/blob/main/docs/api-reference.md#location. So if a path will be `<url>#something` or `<url>?query` this hook skip it and does not register the url change. Currently, we don't use the `#` and `?` in our dapp but maybe someday we will and probably we forget to update this hook. So not sure in which direction we should go 🤷

michalsmiarowski (dismissed):
LGTM 🚀

.env.production
```diff
@@ -4,3 +4,6 @@ REACT_APP_ETH_HOSTNAME_WS=$ETH_HOSTNAME_WS
...
+REACT_APP_POSTHOG_API_KEY=phc_5XMnBAeMVxZQhmBm6G0Kee5ujBOiUhGYIakUKfDhgs5
+REACT_APP_POSTHOG_API_HOST=https://app.posthog.com
```
pdyraga, line 9: Shouldn't we take it from repository secrets instead of hardcoding?

r-czajkowski: d532b76

michalsmiarowski (approved):
LGTM

pdyraga (approved):

Issue #259, opened 2022-10-04
Display Operator Mapping Card when operator is mapped


michalsmiarowski:
We should keep the Operator Mapping Card displayed after the operator is mapped. In that case, instead of displaying the button that opens the operator mapping modal, we should display staking provider <> operator pair that is mapped.

<img width="545" alt="image" src="https://user-images.githubusercontent.com/40306834/193849782-b649a296-ac49-405b-9907-e71a0c36626d.png">

Figma desings: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6179%3A147494

EDIT:

The designs were updated:

![image](https://user-images.githubusercontent.com/40306834/194352110-309225ae-8a69-47fe-bfd5-943c54f1008f.png)

![image](https://user-images.githubusercontent.com/40306834/194352139-e424d6ea-5199-493c-ac59-4bb38b602536.png)

Figma desings: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6658%3A180912

PR #261, opened 2022-10-05
Add success state to OperatorAddressMappingCard

michalsmiarowski:
Closes: #259 
Depends on: #223 

`OperatorAddressMappingCard` should be displayed in staking page also for the providers that already mapped an operator. We will keep this card with staking provider <> operator pair displayed.

image

There are also a tooltip for each to help user know which address is which.

**Note:** The OperatorAddressMappingCard with a success state should be displayed at the top on left column when staking provider doesnt own any stakes, and at the bottom of the left column if he owns some stakes.

TODO in the future:
- [ ] add info about mapped operator in PRE application - this will be addressed in a sperate PR - https://github.com/threshold-network/token-dashboard/pull/277

Link to Figma: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=6179%3A147494


src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -46,22 +56,49 @@ const OperatorAddressMappingCard: FC<{
...
+          <Alert
+            status={isOneOfTheAppsNotMapped ? "error" : "warning"}
+            alignItems={"center"}
+            mt={5}
+            p={"8px 10px"}
+          >
+            <AlertIcon h={"14px"} as={"div"} alignSelf="auto" />
+            <BodyXs>
+              {isOneOfTheAppsNotMapped
+                ? "One application from the tBTC + Random Beacon Rewards Bundle Suite requires the Operator Address mapped."
```
georgeweiler, line 93: I know that this matches designs, but I think it might be more clear to tell the user which app needs to be mapped. wdyt? worth bringing up?

georgeweiler:
@michalsmiarowski It looks like the designs have been re-updated. Do you want to implement the newest changnes here before we merge?

michalsmiarowski:
@r-czajkowski @georgeweiler This one is ready to be reviewed.

I will address the info about PRE app in a separate PR, because it will require a lot of changes.

src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -1,22 +1,45 @@
...
+  stakingProvider: string
+  mappedOperators: MappedOperatorsForStakingProvider
+}> = ({ stakingProvider, mappedOperators }) => {
+  const { tbtc: mappedOperatorTbtc, randomBeacon: mappedOperatorRandomBeacon } =
+    mappedOperators
+
+  const appLabels: AppLabels = {
+    tbtc: "tBTC",
+    randomBeacon: "Random Beacon",
+  }
```
r-czajkowski, line 41: Let's use a function from https://github.com/threshold-network/token-dashboard/commit/ea991e2232121d13ff6cbc86f1f08de64753f96d.

src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -44,22 +71,66 @@ const OperatorAddressMappingCard: FC<{
...
-            : "This section is for Staking Providers and self-operating stakers only. Map your Operator Address to your Provider Address for a better support for your hardware wallet in the node setup."}
-        </BodyXs>
-      </Alert>
-      <Button size="lg" w="100%" mt="5" onClick={onStartMappingClick}>
-        Start Mapping
-      </Button>
+      {shoudlDisplaySuccessState ? (
+        Object.entries(mappedOperators).map(([appName, operator]) => {
+          return (
+            <Box key={`mapped_operator_${appName}_${operator}`}>
```
r-czajkowski, line 77: I think we can get rid of the `Box` component and use `Hstack` directly. Also adding the `divider` prop to the `HStack` we can get rid of the logic for rendering the line divider, I mean this code snippet: 
```jsx
 {Object.keys(mappedOperators)[
                Object.keys(mappedOperators).length - 1
              ] !== appName && <LineDivider />}
```

r-czajkowski (changes requested):
Left two minor comments. Overall looks good 👍

src/pages/Staking/OperatorAddressMappingCard.tsx
michalsmiarowski: Fixed in d818bfae1f4bd8a655c2e27cfb31000d7a7c13e3

r-czajkowski (approved):

PR #257, opened 2022-10-03
Fixes visual bug on Stake Card

georgeweiler:
Resolves #227 by updating the component from Badge to BoxLabel.

PR #262, opened 2022-10-05
Fix warning caused by AuthorizeApplciationRow

michalsmiarowski:
After we made some changes in the AuthorizeApplicationRow (https://github.com/threshold-network/token-dashboard/pull/257) two warnings in the console appeared that we missed during review process:

`<p> cannot appear as a descendant of <p>`

and

`<div> cannot appear as a descendant of <p>`

Turne out we used BoxLabel which is wrapped with TypographyElement. Because of that we shouldn't use `Body` or other typography components as children of BoxLabel.


georgeweiler:
LGTM  I noticed this too.

georgeweiler (approved):

r-czajkowski (approved):

Issue #237, opened 2022-10-03
Card gap should be 16px in all cases


georgeweiler:
![Screen Shot 2022-10-03 at 11 25 28 AM](https://user-images.githubusercontent.com/15269424/193640384-43834fb3-613a-4a06-854f-ebbe7d453402.png)

PR #265, opened 2022-10-06
Updates spacing between several components

georgeweiler:
Resolves https://github.com/threshold-network/token-dashboard/issues/237

There were also a few places where I think `SimpleGrid` worked better than making our own grid with Stacks and widths.


src/pages/Upgrade/UpgradeToken.tsx
```diff
@@ -19,18 +19,13 @@ const UpgradeToken: FC<RouteProps & { token: UpgredableToken }> = ({
   }
 
   return (
-    <Stack
-      direction={{ base: "column-reverse", md: "row" }}
-      w="100%"
-      align="flex-start"
-      spacing="1rem"
-    >
+    <SimpleGrid columns={{ base: 1, xl: 2 }} spacing={4}>
```
r-czajkowski, line 22: Now we have two columns 50/50 are we sure it's ok?

PR impl
obraz

Current prod impl:
obraz

src/pages/Upgrade/UpgradeToken.tsx
```diff
@@ -19,18 +19,13 @@ const UpgradeToken: FC<RouteProps & { token: UpgredableToken }> = ({
   }
 
   return (
-    <Stack
-      direction={{ base: "column-reverse", md: "row" }}
```
r-czajkowski, line 23: We set the `column-reverse` to see the token balances on the top on smaller screens instead of upgrade form which IMO makes more sense. Are we sure this order is ok?

PR impl
obraz

Current prod impl:
obraz

r-czajkowski (changes requested):

r-czajkowski (approved):

src/pages/Upgrade/UpgradeToken.tsx
r-czajkowski: 4463ce5

src/pages/Upgrade/UpgradeToken.tsx
r-czajkowski: 4463ce5

michalsmiarowski (approved):
LGTM

Issue #244, opened 2022-10-03
Remove "Authorize Selected Apps" button if all apps have been authorized


georgeweiler:
![Screen Shot 2022-10-03 at 12 06 41 PM](https://user-images.githubusercontent.com/15269424/193647641-1981be3e-3d3c-4e88-9bae-f6bc2ed4c622.png)

PR #268, opened 2022-10-10
Fix multi-app staking flow

r-czajkowski:
Closes: #244

This PR removes `Authorize Selected Apps` button if all apps have been authorized because there are no longer any more apps to select. It also fixes an issue with selected apps after authorization transaction- the authorized app was not removed from the `selectedApps` array so the user was able to click the `Authorize Selected Apps` again.

### DIFF
--- a/src/pages/Staking/AuthorizeStakingApps/index.tsx
+++ b/src/pages/Staking/AuthorizeStakingApps/index.tsx
@@ -83,10 +83,6 @@ const AuthorizeStakingAppsPage: FC = () => {
     dispatch(stakingApplicationsSlice.actions.getSupportedApps({}))
   }, [dispatch, account])
 
-  const tbtcMinAuthAmount = useStakingAppMinAuthorizationAmount("tbtc")
-  const randomBeaconMinAuthAmount =
-    useStakingAppMinAuthorizationAmount("randomBeacon")
-
   const tbtcApp = useStakingAppDataByStakingProvider(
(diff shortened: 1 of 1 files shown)


src/pages/Staking/AuthorizeStakingApps/index.tsx
```diff
@@ -130,6 +115,33 @@ const AuthorizeStakingAppsPage: FC = () => {
...
+  useEffect(() => {
+    if (tbtcApp.isAuthorized) {
+      setSelectedApps((selectedApps) =>
+        selectedApps.filter(({ stakingAppId }) => stakingAppId !== "tbtc")
+      )
+    }
+
+    if (randomBeaconApp.isAuthorized) {
+      selectedApps.filter(({ stakingAppId }) => stakingAppId !== "randomBeacon")
+    }
```
michalsmiarowski, line 127: Shouldn't there be `setSelectedApp` like in the condition above? 🤔

michalsmiarowski (changes requested):
Left one small comment to look at before merge.

src/pages/Staking/AuthorizeStakingApps/index.tsx
r-czajkowski: 🦅 👁️ fixed in cd3636a.

michalsmiarowski (approved):
LGTM

Issue #240, opened 2022-10-03
Staking Info Card needs updates


georgeweiler:
- [ ] Remove “Set PRE” button
- [ ] Change the button style for Top Up to be secondary button style
- [ ] Remove “Missing PRE” tag
- [ ] Add alert for Missing Node instead that links to docs.threshold.network
- [ ]  Badge style adjustment in Dark Mode (noted previously)
- [ ] PRE App needs to be shown in Applications - “PRE APP” with 100%

Card should match designs:

![Screen Shot 2022-10-03 at 11 29 27 AM](https://user-images.githubusercontent.com/15269424/193641383-cd0ed282-a13d-4d9e-8ed0-8cebc4dd048a.png)

r-czajkowski:
Regarding the:
> Add alert for Missing Node instead that links to docs.threshold.network

Currently, we are not able to show this alert so let's create a separate ticket for this and add this feature once we are able to get node metrics.

PR #269, opened 2022-10-11
Update stake card component

r-czajkowski:
~Depends on: https://github.com/threshold-network/components/pull/28 https://github.com/threshold-network/components/pull/27~
Closes: #240

Main changes:
- removes `Set PRE` button at the bottom of the stake card component,
- changes the button style for `Top Up/Unstake` to be secondary button style,
- removes `Missing PRE` tag,
- adjusts the icon color in the dark mode under the `Applications` section,
- lists the PRE App under the `Applications` with 100% if the PRE node is running.


src/pages/Staking/StakeCard/StakeApplications/index.tsx
```diff
@@ -22,43 +22,37 @@ const StakeApplications: FC<{ stakingProvider: string }> = ({
...
+          percentage={randomBeaconApp.percentage}
+          stakingProvider={stakingProvider}
+        />
+        <AuthorizeApplicationRow
+          as="li"
+          label="PRE"
+          isAuthorized={isPRESet}
+          percentage={100}
+          stakingProvider={stakingProvider}
+        />
```
michalsmiarowski, line 54: Since we remove the 'set up PRE' button I think we should always display `PRE` app as authorized, so I would remove the `isPRESet` variable and just set `isAuthorized` to true.

The reson for that is that there is not link to the docs now where the user can read how to set up PRE (because we removed `set up PRE` button) so it might getn confused when he have `Authorize Application` button next to PRE but he doesn't have to authorize it 🤔 

image

michalsmiarowski (changes requested):
Left one small comment. Also let's resolve conflicts

michalsmiarowski (approved):
LGTM

PR #271, opened 2022-10-13
Update MAS success modal

r-czajkowski:
Display copy and buttons based on the number of authorized applications. When two applications have been successfully authorized, we should change the `Authorize Other Apps` button to be `Node Setup Doc` and the secondary button to be `Dismiss` since there aren't any more apps to authorize.

### DIFF
--- a/src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
+++ b/src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
@@ -17,6 +17,7 @@ import {
   Divider,
   Link,
   FlowStepStatus,
+  ButtonProps,
 } from "@threshold-network/components"
 import InfoBox from "../../InfoBox"
 import ExternalLink from "../../ExternalLink"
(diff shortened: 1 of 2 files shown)


src/components/Modal/StakingApplications/StakingApplicationsAuthorized.tsx
```diff
@@ -57,6 +58,8 @@ const StakingApplicationsAuthorizedBase: FC<
     navigate(`/staking/${stakingProvider}/authorize`)
   }
 
+  const numberOfApps = authorizedStakingApplications.length
```
michalsmiarowski, line 61: Non-blocking:

I would change the name of `numberOfApps` to `numberOfAuthorizedApps`

michalsmiarowski (approved):
LGTM 🚀 

Left one small, non-blocking comment. Let's resolve conflicts and we can merge

michalsmiarowski (approved):
LGTM

Issue #250, opened 2022-10-03
Stake Flow modal - confirm staking params


georgeweiler:
- [ ] Do not display the beneficiary, authorizer, etc addresses. Only the stake amount
- [ ] Helper text below the input needs to be updated to match designs

![Screen Shot 2022-10-03 at 12 28 14 PM](https://user-images.githubusercontent.com/15269424/193651529-1d14343f-b234-45bb-89cc-61c057f6f0aa.png)

PR #258, opened 2022-10-03
Link Component Refactor

georgeweiler:
- Removes `InternalLink` and `ExternalLink` components
- Adds a `Link` component with correct styling and `isExternalProp`
- Refactors all uses of `ExternalLink` to use new `Link` component. Replaces `text` prop with children
- Adds `ButtonLink` component with `isExternal` prop, and updates usages
- Removes unnecessary of `as={RouterLink}`

Fixes: https://github.com/threshold-network/token-dashboard/issues/228
Fixes: https://github.com/threshold-network/token-dashboard/issues/230

PR #272, opened 2022-10-13
Update the confirm staking params modal

r-czajkowski:
~Depends on: #258~
Closes: #250 

Update helper texts below the input to match designs. Also get rid of the addresses from the list of confirmation params to match designs.


michalsmiarowski (approved):
LGTM 🚀

Issue #249, opened 2022-10-03
Copy update: tBTC


georgeweiler:
The convention should be "tBTC" in every case
![Screen Shot 2022-10-03 at 12 22 28 PM](https://user-images.githubusercontent.com/15269424/193650486-45b813c7-8ce4-4d12-b3fa-c9638780cb98.png)

PR #273, opened 2022-10-13
Update copy: tBTC

r-czajkowski:
Closes: #249

The convention should be `tBTC` in every case.


src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
```diff
@@ -25,6 +25,11 @@ import TokenBalance from "../../TokenBalance"
 import { StakingAppName } from "../../../store/staking-applications"
 import { useInitiateDeauthorization } from "../../../hooks/staking-applications"
 
+const stakingAppNameToAppLabel: Record<StakingAppName, string> = {
+  tbtc: "tBTC",
+  randomBeacon: "Random Beacon",
+}
```
michalsmiarowski, line 31: I've created something similar in #261 

https://github.com/threshold-network/token-dashboard/pull/261/files#diff-3fa8603c049ef8353a7dd98f5e20aa5a41453162f5100edee663a8b186ea152bR38-R41

Since this is used in many places I think we should export it and move it to some other place - maybe to `src/types`?

michalsmiarowski (changes requested):
Left one small comment

src/components/Modal/DeauthorizeApplicationModal/InititateDeauthorization.tsx
r-czajkowski: ea991e2

michalsmiarowski (approved):
LGTM 🏒

Issue #256, opened 2022-10-03
Tooltips need updating


georgeweiler:
- [ ] ensure copy is correct and exists for all tooltips
- [ ] verify with designers that tooltips should have consistent styling (outline vs filled) and size

example:
 
![Screen Shot 2022-10-03 at 1 44 39 PM](https://user-images.githubusercontent.com/15269424/193665185-47bade59-b4a7-4d17-b3f9-bb7f7504ee33.png)

PR #276, opened 2022-10-17
Update staking application tooltips

r-czajkowski:
Closes: #256
~Depends on: #258~

This PR updates the tooltips for staking apps to match the TDS.


src/components/TooltipIcon.tsx
```diff
@@ -1,23 +1,51 @@
...
-      <span>
-        <Icon
-          color="gray.500"
-          boxSize="16px"
-          as={IoInformationCircleSharp}
-          {...iconProps}
-        />
-      </span>
-    </Tooltip>
+    <Popover trigger="hover" placement="top">
```
michalsmiarowski, line 26: Non-blocking:

To make it more reusable we could add the `placement` to the props with a defailt `top` value

michalsmiarowski (approved):
LGTM! 🦜 

Left one non-blocking comment

PR #277, opened 2022-10-17
Add PRE Application to the thershold-ts lib

michalsmiarowski:
Depends on: #261 

Adds PRE application to the Operator Mapping Success Card

### DIFF
--- a/src/store/account/effects.ts
+++ b/src/store/account/effects.ts
@@ -65,6 +65,13 @@ export const getStakingProviderOperatorInfo = async (
       })
     )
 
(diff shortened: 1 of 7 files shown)


vlc823:
Will anyone help me. I put all my savings into the Threshold staking page and I can not withdraw my stake. I was not able to figure out staking so I only was able to put step 1 and 2 up but was not able to get a node or choose the application to run the node and mapping address. i tried to unstake and take my funds back and it will not let me. It says error, can not estimate gas, etc etc

PR #278, opened 2022-10-20
Simplify mapped operators data check

michalsmiarowski:
The main goal is to reduce the repetition of the code for `isOperatorMappedOnlyInRandomBeacon`, `isOperatorMappedOnlyInTbtc` etc. For `useRegisterMultipleOperator` hook I've created the reducer that will get the operator addresses from the state and returns all the needed data.


r-czajkowski (approved):
LGTM!

Issue #287, opened 2022-11-02
Generate QR code for a deposit address


michalsmiarowski:
<img width="461" alt="image" src="https://user-images.githubusercontent.com/40306834/199466722-9e20c805-53dc-442f-8751-3d0d8e8b7fa9.png">

PR #297, opened 2022-11-08
Generate QR code for the deposit address

michalsmiarowski:
Closes: #287 

Generates proper QR Code for the deposit address.

For that purpose we will use [react-qr-code](https://www.npmjs.com/package/react-qr-code) lib.


src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
```diff
@@ -22,10 +21,9 @@ import CopyToClipboard from "../../../../components/CopyToClipboard"
 import { useTbtcState } from "../../../../hooks/useTbtcState"
 import shortenAddress from "../../../../utils/shortenAddress"
 import { MintingStep } from "../../../../types/tbtc"
-
-import { Divider, useColorModeValue } from "@chakra-ui/react"
 import ViewInBlockExplorer from "../../../../components/ViewInBlockExplorer"
 import { ExplorerDataType } from "../../../../utils/createEtherscanLink"
+import QRCode from "react-qr-code"
```
r-czajkowski, line 26: Can we create our wrapper for this external component? If we changed the lib for generating QR code we would change only our wrapper component and keep the same interface.

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
```diff
@@ -77,23 +75,33 @@ export const MakeDeposit: FC = () => {
...
-        <Image
-          my={6}
-          mx="auto"
-          bg="white"
-          borderRadius="lg"
-          maxW="145px"
-          maxH="145px"
-          src={btcQrTmp}
-        />
+        <Flex direction={"column"} alignItems={"center"}>
```
r-czajkowski, line 78: Do we need flex wrapper?

r-czajkowski (changes requested):
Left two minor comments.

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
michalsmiarowski: Fixed in 0bbe4fbc41e7953f1b58180c6e87c946bce03e9f

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
michalsmiarowski: Fixed in 5ae4f2febbc0c63b29f243f60d66a8a90da59b39

r-czajkowski (approved):

PR #299, opened 2022-11-09
Threshold lib updates

r-czajkowski:
This PR moves the contract integration part from hooks to the Threshold-ts library. Thanks to this layer we can separate work between UI and contract integration. Also, it helps to unit test created services and move the "business logic" from the React hooks.

### Vending Machine
An interface of the wrapper for fetching on-chain data from the `VendingMachine` contract- the interface of update protocol to enable KEEP/NU token holders to wrap their tokens and obtain T tokens according to a fixed ratio. The Threshold lib creates a separate instance of the vending machine for KEEP holders and a separate instance of the vending machine for NU holders. Service covered with unit tests.

### Staking
Adds function `getOwnerStakes` that fetches all stakes for a given owner address. Here we add legacy Keep and Nu staking contracts as private fields to the `Staking` service to calculate possible KEEP/NU top-up in T. Also pass Keep and Nu vending machines to convert KEEP/NU amount to T.

### Utils functions
Move some contract-related utils functions from dashboard to the Threshold-ts lib.

### DIFF
--- a/src/threshold-ts/vending-machine/index.ts
+++ b/src/threshold-ts/vending-machine/index.ts
@@ -1,0 +1,117 @@
+import VendingMachineKeep from "@threshold-network/solidity-contracts/artifacts/VendingMachineKeep.json"
+import VendingMachineNuCypher from "@threshold-network/solidity-contracts/artifacts/VendingMachineNuCypher.json"
+import { BigNumber, BigNumberish, Contract } from "ethers"
+import { EthereumConfig } from "../types"
+import { getContract, STANDARD_ERC20_DECIMALS } from "../utils"
+
+export interface ConversionToT<NumberType extends BigNumberish = BigNumber> {
+  tAmount: NumberType
+  wrappedRemainder: NumberType
+}
+
+/**
+ * Interface of update protocol to enable KEEP/NU token holders to wrap their
+ * tokens and obtain T tokens according to a fixed ratio.
+ */
+export interface IVendingMachine {
+  /**
+   * Number of decimal places of precision in conversion to/from wrapped tokens
+   * (assuming typical ERC20 token with 18 decimals). This implies that amounts
+   * of wrapped tokens below this precision won't take part in the conversion.
+   * E.g., for a value of 3, then for a conversion of 1.123456789 wrapped
+   * tokens, only 1.123 is convertible (i.e., 3 decimal places), and 0.000456789
+   * is left.
+   */
+  WRAPPED_TOKEN_CONVERSION_PRECISION: number
+
+  /**
+   * Divisor for precision purposes, used to represent fractions.
+   */
+  FLOATING_POINT_DIVISOR: BigNumber
+
+  /**
+   * Ethers contract instance of the `VendingMachine` contract.
+   */
+  contract: Contract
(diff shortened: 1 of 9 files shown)


src/threshold-ts/vending-machine/index.ts
```diff
@@ -0,0 +1,117 @@
...
+   * Ethers contract instance of the `VendingMachine` contract.
+   */
+  contract: Contract
+
+  /**
+   * Returns the T token amount that's obtained from `amount` wrapped tokens,
+   * and the remainder that can't be upgraded.
+   * @param amount Amount of the wrapped token to convert to T.
+   * @returns Object representing the T token amount and the reminder that can't
+   * be upgraded @see {@link ConversionToT}
```
michalsmiarowski, line 42: I think the tAmount returned here is in a smalles possible T unit (like wei for ether - do we have an equivalent name for T "cents" 😅 ) and not really in T. I think we should write it in a comment to avoid any potential confusion since `reprresenting the T token amount` might mean that it return the proper T token amount and we don't have to convert it to T. What you think?

src/threshold-ts/vending-machine/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Ethers contract instance of the `VendingMachine` contract.
+   */
+  contract: Contract
+
+  /**
+   * Returns the T token amount that's obtained from `amount` wrapped tokens,
+   * and the remainder that can't be upgraded.
+   * @param amount Amount of the wrapped token to convert to T.
```
michalsmiarowski, line 40: We could also specify here that the `amount` parameter unit is the lowest possible T unit

michalsmiarowski (commented):
Looking good! Left some small comments

src/threshold-ts/vending-machine/index.ts
r-czajkowski: I was thinking about it but repeating that any amount one every place is in the smallest unit could be annoying, especially for writing docs. So I think we could mention about it as a general rule for this lib, that all amounts are in the smallest unit like in solidity, in the readme file in `threshold-ts` dir or a ts-docs for `Threshold` class. If you agree with me could we address it in a separate PR?

src/threshold-ts/vending-machine/index.ts
```diff
@@ -0,0 +1,117 @@
+import VendingMachineKeep from "@threshold-network/solidity-contracts/artifacts/VendingMachineKeep.json"
+import VendingMachineNuCypher from "@threshold-network/solidity-contracts/artifacts/VendingMachineNuCypher.json"
+import { BigNumber, BigNumberish, Contract } from "ethers"
+import { EthereumConfig } from "../types"
+import { getContract, STANDARD_ERC20_DECIMALS } from "../utils"
+
+export interface ConversionToT<NumberType extends BigNumberish = BigNumber> {
+  tAmount: NumberType
+  wrappedRemainder: NumberType
+}
```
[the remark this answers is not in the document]
r-czajkowski: 528884c

michalsmiarowski (approved):
LGTM

PR #306, opened 2022-11-14
Fix `non-serializable value in the state` error

michalsmiarowski:
There was an error in the console when clicking `Generate deposit address` on tbtc page:

`A non-serializable value was detected in the state`

Pasted Graphic 3

This happened because we passed non-serializable values (in this case functions - `handleJsonDownload` and `handleDoubleReject`) as a parameters through a redux action.

This PR moves those two function straight into the `TbtcRecoveryJson` modal.


r-czajkowski:
Fixed in #359.

PR #313, opened 2022-11-17
Installs posthog capture event on close modal events

georgeweiler:
I noticed that our posthog event list had many click events such as "Clicked Path" or "Clicked SVG" when the user closes a modal.

 This PR fixes the problem by:

- Adding `className="ph-no-capture"` to prevent those useless events
- Adding a capture event to the `closeModal` method that logs the modal type


src/hooks/useModal.ts
```diff
@@ -18,6 +19,7 @@ export const useModal: UseModal = () => {
     dispatch(openModalAction({ modalType, props }))
 
   const closeModal = () => {
+    capture(`Closed Modal`, { modalType })
```
r-czajkowski, line 22: Maybe we should create type in `src/posthog/index.ts` for our custom events? eg. `export type PosthogEvents = "modal-closed" | "other-event-here" ....`

src/posthog/index.ts
```diff
@@ -35,3 +35,7 @@ export const capturePageview = () => {
 export const reset = () => {
   posthog.reset()
 }
+
+export const capture = (event: string, params?: any) => {
+  posthog.capture(event, params)
```
r-czajkowski, line 40: shouldn't we check here if the feature flag is enabled? On the other hand, would be great if we could keep a wrapper for posthog independent of our feature flags. So maybe we should create a hook eg:
```tsx
import * as posthog from "../../posthog"
import { featureFlags } from "../../constants"

const useCapture(eventName: PosthogEvents) => {
 return useCallback((params) => {
  if (!featureFlags.POSTHOG) return
  posthog.capture(eventName, params)
 }, [eventName])
}
``` 

and then in a component:

```tsx
const capture = useCapture("modal-closed")

const closeModal = () => {
 capture({ modalType } )
}

```

r-czajkowski (changes requested):

georgeweiler:
All very good suggestions @r-czajkowski . I've implemented your change requests

src/hooks/useModal.ts
georgeweiler: I decided to use an enum in `src/types/posthog`. let me know if that works for you

src/types/types.ts
```diff
@@ -0,0 +1,3 @@
+export enum posthogEvent {
```
r-czajkowski, line 1: Shouldn't be `PosthogEvent`(starting with a capital letter)?

r-czajkowski, line 1: Maybe let's rename the file to `src/types/posthog.ts`.

src/posthog/index.ts
```diff
@@ -21,6 +22,7 @@ export const init = () => {
 }
 
 export const identify = (ethAddress: string) => {
+  console.log("identifying the eth address: ", ethAddress)
```
r-czajkowski, line 25: Do we need this console log?

src/hooks/posthog/useCapture.ts
```diff
@@ -0,0 +1,14 @@
+import { useCallback } from "react"
+import { posthogEvent } from "../../types/types"
+import { featureFlags } from "../../constants"
+import * as posthog from "../../posthog"
+
+export const useCapture = (eventName: posthogEvent) => {
```
r-czajkowski, line 6: Can we export this hook from `src/hooks/posthog/index.ts` ?

`src/hooks/posthog/index.ts`
```
export * from "./usePosthog"
export * from "./useCapture"
```

src/hooks/useModal.ts
```diff
@@ -8,16 +8,20 @@ import {
 } from "../store/modal"
 import { RootState } from "../store"
 import { ModalType } from "../enums"
+import { posthogEvent } from "../types/types"
+import { useCapture } from "./posthog/useCapture"
```
r-czajkowski, line 12: If you agree with a comment from above let's import from `./posthog`.

r-czajkowski (changes requested):

src/types/types.ts
georgeweiler: whoops 🙈

r-czajkowski (approved):

Issue #311, opened 2022-11-16
Check if there are any incoming transactions for the given deposit address


michalsmiarowski:
Blocked by: #310 

On the step 2 of the minting flow we should check periodically if the given deposit address has any incoming transactions. We need to be sure that the transaction is there, because we need to exploit the data from it to use it to reveal the deposit in Step 3. If there are not any incoming transactions for the given address we should disable the `I sent the BTC` button.

Besides that we should also make sure that the reveal was not done for the given deposit.

Figma: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=5807%3A139248

Issue #305, opened 2022-11-14
Convert P2(W)PKH BTC address provided by a user to a public key hash


michalsmiarowski:
The deposit script needs PKH (public key hash) to generate a deposit BTC address. That's why we have to decode BTC Address given by the user into a public key hash.

Function that decodes the BTC address was placed in this PR: https://github.com/keep-network/tbtc-v2/pull/419

Issue #304, opened 2022-11-14
Get proper wallet public key for calculating a deposit address


michalsmiarowski:
Right now deposit address is calculated using a same wallet public key in each case for testing purposes.

We should get the newest wallet using tbtc-v2.ts lib

Issue #288, opened 2022-11-02
Reveal deposit functionality


michalsmiarowski:
Blocked by: #310 

Add reveal deposit functionality that should happen after clicking "Confirm deposit & mint" button in the step 3 of the tbtc-v2 deposit flow.

PR #314, opened 2022-11-18
Create TBTC wrapper in threshold lib

michalsmiarowski:
~Depends on: https://github.com/keep-network/tbtc-v2/pull/419~
~Depends on: https://github.com/keep-network/tbtc-v2/pull/425~
~Depends on: https://github.com/keep-network/tbtc-v2/pull/441~

Closes: #304 
Closes: #311 
Closes: #288 
Closes: #305 

Child branch: #321 

Adds a whole reveal deposit functionality.

Main changes:
- refactor TBTC wrapper
- finds UTXOs of the correct connected user by his deposit address in step 2
- actual functionality to reveal a deposit or multiple deposits
- functionality that saves the deposit data in local storage, which makes it possible to get the actual data after refresh.


src/threshold-ts/tbtc/mock-bitcoin-client.ts
```diff
@@ -0,0 +1,120 @@
+import {
+  Client,
+  RawTransaction,
+  Transaction,
+  TransactionHash,
+  TransactionMerkleBranch,
+  UnspentTransactionOutput,
+} from "@keep-network/tbtc-v2.ts/dist/bitcoin"
+
+export class MockBitcoinClient implements Client {
```
r-czajkowski, line 10: If you agree with the comment above, I think we can move the mock implementation outside the `threshold-ts` dir because it's related to the dapp and dapp environment.

src/threshold-ts/tbtc/tbtc.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+    network: string
+  ): Promise<string>
+
+  findAllUnspentTransactionOutputs(
+    address: string
+  ): Promise<UnspentTransactionOutput[]>
+
+  revealDeposit(utxo: UnspentTransactionOutput, deposit: Deposit): Promise<void>
+
+  mockDepositTransaction(depositAddress: string): void
```
r-czajkowski, line 26: Let's get rid of this mock function- I believe we should not have mock functions in an interface. We should mock the function by implementation and we can do that by creating a fake deposit transaction in `BitcoinClient.findAllUnspentTransactionOutputs` and return it.

src/threshold-ts/tbtc/mock-tbtc.ts
```diff
@@ -0,0 +1,134 @@
...
+      ...testnetTransaction,
+    }
+
+    const utxos = new Map<string, UnspentTransactionOutput[]>()
+    utxos.set(depositAddress, [testnetUTXO])
+    this._bitcoinClient.unspentTransactionOutputs = utxos
+
+    const rawTransactions = new Map<string, RawTransaction>()
+    rawTransactions.set(testnetTransactionHash, testnetTransaction)
+    this._bitcoinClient.rawTransactions = rawTransactions
```
r-czajkowski, line 132: We can move this to `findAllUnspentTransactionOutputs` function from mock implementation of `BitcoinClient`.

src/threshold-ts/types/index.ts
```diff
@@ -9,5 +9,6 @@ export interface EthereumConfig {
 export interface ThresholdConfig {
   ethereum: EthereumConfig
   // TODO: Define the bitcoin config for the tbtc v2 lib.
+  mockTbtc: boolean
```
r-czajkowski, line 12: I would love to avoid a flag like this one. We can pass the `BitcoinClient` implementation via config if needed.

```ts
import { Client } from "@keep-network/tbtc-v2.ts/dist/bitcoin"

export type BitcoinConfig =
  | {
      client: Client
      url?: never
    }
  | {
      client?: never
      url: string
      // and other necessary credentials to create the BitcoinClient
    }
    
    export interface ThresholdConfig {
  ethereum: EthereumConfig
  bitcoin: BitcoinConfig
}
```

src/web3/utils/address.ts
```diff
@@ -11,7 +11,7 @@ export const isAddress = (address: string): boolean => ethersIsAddress(address)
 export const unprefixedAndUncheckedAddress = (address: string): string => {
   const prefix = address.substring(0, 2)
   if (prefix === "0x") {
-    address.slice(2)
+    address = address.slice(2)
```
r-czajkowski, line 14: Now we have 2 implementations of `unprefixedAndUncheckedAddress` one in dapp and one in `threshold-ts`. Let's keep this in `threshold-ts` and export from `web3/utils/address.ts` to keep the same utils API for T dapp.

`src/threshold-ts/utils/address.ts`
```ts
export const unprefixedAndUncheckedAddress = (address: string): string => { ... }
```
`src/web3/utils/address.ts`
```ts
export { unprefixedAndUncheckedAddress } from "../../threshold-ts/utils" 
```

r-czajkowski (changes requested):

r-czajkowski (approved):

PR #317, opened 2022-11-22
Opt-In analytics + Hashed analytics ID

georgeweiler:


src/utils/hashString.ts
```diff
@@ -0,0 +1,17 @@
+const hashString = async ({
+  algorithm = "SHA-256",
+  string,
+}: {
+  algorithm?: string
```
michalsmiarowski, line 5: I think we should specify either union literal string type or string enum for this one. Reading from the docs the available types are:

- "SHA-1"
- "SHA-256"
- "SHA-384"
- "SHA-512".

docs: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#parameters

src/utils/hashString.ts
```diff
@@ -0,0 +1,17 @@
...
+  string: string
```
michalsmiarowski, line 6: Maybe `value` would be a better name to avoid the duplacte of string word in different context? We can keep the function name as `hashString`

src/utils/hashString.ts
```diff
@@ -0,0 +1,17 @@
+const hashString = async ({
```
michalsmiarowski, line 1: Can we rename the file from `hashString.ts` to `crypto.ts`? We could create all the crypto related utility functions here.

Also let's export this function in `utils/index.ts` to simplify the imports

src/store/analytics/analyticsSlice.ts
```diff
@@ -0,0 +1,22 @@
...
+  } as AnalyticsState,
+  reducers: {
+    optIn: (state: AnalyticsState) => {
+      state.shouldEnableAnalytics = true
+    },
+    optOut: (state: AnalyticsState) => {
+      state.shouldEnableAnalytics = false
+    },
+  },
+})
```
michalsmiarowski, line 20: Do we actually want to save this info in a redux store? 🤔 Wouldn't that mean that after refresh the user would have accept/decline the `AnalyticsConfirmation` modal again?

Maybe the better idea would be to store it in local storage, associated with given account?

src/store/analytics/analyticsSlice.ts
```diff
@@ -0,0 +1,22 @@
...
+  initialState: {
+    shouldEnableAnalytics: false,
+  } as AnalyticsState,
+  reducers: {
+    optIn: (state: AnalyticsState) => {
+      state.shouldEnableAnalytics = true
+    },
+    optOut: (state: AnalyticsState) => {
+      state.shouldEnableAnalytics = false
+    },
```
michalsmiarowski, line 18: I think `optIn` and `optOut` are too generic in my opinion. We often export all the actions from the slice. For example in staking slice: 

```
export const {
  updateState,
  setStakes,
  providerStaked,
  providerStakedForStakingProvider,
  toppedUp,
  unstaked,
  setMinStake,
} = stakingSlice.actions
```

If we just have `optIn` and `optOut` then it doesn't really say anything what those action changes. I would change them to something like `analyticsEnabled` and `analyticsDisabled`

michalsmiarowski (changes requested):
Left some comments too look at before the merge 👇

georgeweiler:
closing because the changes in https://github.com/threshold-network/token-dashboard/pull/323 better implement the local storage functionality.

PR #319, opened 2022-11-28
Adds a fixed button to route to the feedback module pages

georgeweiler:
per designs: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=3800%3A141145&t=oObH989OZYEP0dDa-0

Screenshot 2022-11-28 at 4 17 20 PM

### DIFF
--- a/src/components/FeedbackRoutesButton.tsx
+++ b/src/components/FeedbackRoutesButton.tsx
@@ -47,7 +47,7 @@ const FeedbackRoutesButton: FC = () => (
         </BodyMd>
         {feedbackRoutes.map(({ text, url }) => (
           <Link key={text} to={url} as={RouterLink}>
-            <MenuItem>{text}</MenuItem>
+            <MenuItem as="span">{text}</MenuItem>
           </Link>
         ))}
       </MenuList>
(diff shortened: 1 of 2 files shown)


src/components/FeedbackRoutesButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+        }
+      />
+      <MenuList>
+        <BodyMd fontWeight="bold" ml={3} mb={3} mt={2}>
+          Feedback
+        </BodyMd>
+        {feedbackRoutes.map(({ text, url }) => (
+          <Link key={text} to={url} as={RouterLink}>
+            <MenuItem>{text}</MenuItem>
+          </Link>
```
r-czajkowski, line 51: I believe the `MenuItem` renders `button` tag so we have `button` inside `a` and that's not valid HTML snippet. So maybe `Link as={MenuItem}`?.

r-czajkowski (changes requested):
Overall looks great! Let's take a look at the comment below before the merge.

r-czajkowski (approved):

PR #322, opened 2022-11-30
Bump tbtc v2 dependency

r-czajkowski:
This PR bumps the `@keep-network/tbtc-v2.ts` dependency to the latest one with `development` tag. It also updates the workflows because of the randomly failing post-install script from `@threshold-network/solidity-contracts`. As a workaround we install token dashboard dependencies with `--ignore-scripts` flag. We have not observed any issues with the T-dapp due to not executing the post-install scripts so far. Note that the `--ignore-scripts` flag ignores all post-install scripts even for T-dapp so we have to call it manually. For local development, we should install the dependencies with `--ignore-scripts` and then manually run the post-install script `yarn run postinstall`.

### DIFF
--- a/.github/workflows/dashboard-ci.yml
+++ b/.github/workflows/dashboard-ci.yml
@@ -42,7 +42,10 @@ jobs:
         run: git config --global url."https://".insteadOf git://
 
       - name: Install dependencies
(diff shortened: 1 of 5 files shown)


michalsmiarowski (approved):
LGTM

PR #323, opened 2022-11-30
Refactors analytics succes modals

georgeweiler:
- The analytics success and decline modals were very similar so have been combined into one modal with props.
- They are also both used in the "settings" page, so their location should not have been tied directly to the analytics modal. A new modal was created to be shared.
- The Analytics "info" had the same problem, so it was moved to a new sharable component
- Other feedback forms use the same modal style, so their content was added


src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,22 @@
+import { useLocalStorage } from "@rehooks/local-storage"
```
r-czajkowski, line 1: I'll refactor our custom implementation of `useLocalStorage` and use hook from `@rehooks/local-storage` so we will use the new implementation in all places and we will have a nice wrapper for external lib.

r-czajkowski (changes requested):
Looks good 👍 . Left few minor comments.

src/hooks/useAnalytics.ts
r-czajkowski: #328

src/components/Modal/FeedbackSubmissionModal/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        </Alert>
+        <InfoBox p={6}>
+          <H5 mb={4}>{title}</H5>
+          {typeof body === "string" ? <BodyLg>{body}</BodyLg> : body}
+        </InfoBox>
+      </ModalBody>
+      <ModalFooter>
+        <Button onClick={closeModal}>Dismiss</Button>
+      </ModalFooter>
+    </>
```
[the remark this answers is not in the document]
georgeweiler: I agree. let's tackle this refactor in another PR. happy to take that on.

src/pages/Overview/index.tsx
```diff
@@ -8,15 +8,20 @@ import { PageComponent } from "../../types"
...
+  const { isAnalyticsEnabled, hasUserResponded } = useAnalytics()
+
   return (
     <Container maxW={{ base: "2xl", xl: "6xl" }} my={16}>
       <Image src={thresholdWordMark} mb={4} />
       <H1 mb={12}>Overview</H1>
-      {featureFlags.FEEDBACK_MODULE && <AnalyticsBanner />}
+      {featureFlags.FEEDBACK_MODULE &&
+        !isAnalyticsEnabled &&
+        !hasUserResponded && <AnalyticsBanner />}
```
michalsmiarowski, line 24: Non- blocking, but something to think about:

Because of `!hasUserResponded` condition we won't display the `AnalyticsBanner` after use decline it (unless he removes the proper data from local storage manually). Maybe the better idea would be to display the banner anyway even if `isAnalyticsEnabled` is set to false?

For a quick solution i propose to remove `!hasUserResponded` from this condition. This way the AnalyticsBanner will display always until the user accepts it. He can click "X" on the right corner if he doesn't want to see the banner. What do you think @georgeweiler ?

src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,23 @@
+import { useLocalStorage } from "./useLocalStorage"
+import { useCallback } from "react"
+
+export const useAnalytics = () => {
+  const [isAnalyticsEnabled, setIsAnalyticsEnabled] = useLocalStorage<
+    boolean | undefined
+  >("isAnalyticsEnabled", undefined)
```
michalsmiarowski, line 7: I think it's pretty important but I mark it as non-blocking:

I think we should add the info about analytics enabled for each account separately. The key should have the address in it's name. For example:

`0x888467263A7384D1ce4eFf7eF4623D32d838eF6E-isAnalyticsEnabled`: `true`
`0xE432f97B4EfEc07166150a5F013b53a215ffFc3A-isAnalyticsEnabled`: `false`

I'm marking this as non-blocking since I would like to have your opinion first @r-czajkowski @georgeweiler. We can address it in separate PR if needed.

src/components/Modal/AnalyticsModal/index.tsx
```diff
@@ -1,23 +1,79 @@
...
+        </H5>
+        <BodyLg mb={10} color={subduedTextColor}>
+          Change this at any time in Feedback Settings.
+        </BodyLg>
+        <FeedbackAnalyticsInfo />
+        <VStack mb={4}>
+          <BodyXs color={subduedTextColor}>
+            By clicking Accept, you consent to allow a cookie to record this
+            setting.
+          </BodyXs>
```
michalsmiarowski, line 66: Non-blocking UI issue:

Horizontal line is missing under this text. Based on the figma views:

image

michalsmiarowski (approved):
Looking good 🔥 

I left some comments to look at. They ar non-blocking so I'm giving an approve but let's look at them as I believe some of them might be pretty important.

r-czajkowski (approved):

michalsmiarowski (approved):
LGTM

PR #324, opened 2022-12-01
Rewards allocation for Dec 1st 2022

r-czajkowski:
This PR adds a merkle tree computed in https://github.com/threshold-network/merkle-distribution/pull/53 to Threshold Token Dashboard. It also sets the version of `tbtc-v2.ts` dependency to `1.0.0-dev.11`(from `yarn.lock`), this is the exact version used to build the T-dapp in `v1.3.x`, to fix CI build. The `build-and-test` job fails because upgrades the `tbtc-v2.ts` to the latest version with the development tag and the T-dapp is not up to date with these changes. We do not want to
bump dependencies on the release branch. The mainnet and preview jobs builds the dapp with `--frozen-lockfile` flag and `yarn.lock` points to the `1.0.0-dev.11`. Thanks to this all builds should use the same version `1.0.0-dev.11` of the `tbtc-v2.ts` dependency.


michalsmiarowski (approved):
LGTM 🚀

manumonti (approved):
LGTM! ✌️

PR #318, opened 2022-11-28
Missing path when decreasing authorization

r-czajkowski:
Closes: #266 

This PR adds support for decreasing authorization for non-beta operators. The operator who is not a beta operator could not join the pool. If they are deauthorizing to 0, from the sortition pool's perspective, their status is up to date (see [`RandomBeacon.isOperatorUpToDate`](https://github.com/keep-network/keep-core/blob/4e940f02b8bebeafcae2727016e00ea70355a709/solidity/random-beacon/contracts/libraries/BeaconAuthorization.sol#L551) and [`WalletRegistry.isOperatorUpToDate`](https://github.com/keep-network/keep-core/blob/4049dc0153c33c3c4cbb2e9a1808581bf76144fb/solidity/ecdsa/contracts/libraries/EcdsaAuthorization.sol#L554)). The client will not attempt to update the pool if the status is up to date. The new deauthorization flow is: when the authorization decrease is requested, we check if there is a staking provider <> operator mapping. If such mapping exists, we check if the operator is in the pool. If not, we execute a call to `RandomBeacon.updateOperatorStatus / WalletRegistry.updateOperatorStatus` to activate the authorization decrease delay.

### Threshold-ts lib
1. Updates the `getStakingProviderAppInfo` function- add additional fields to returned object:
   - `operator`- operator address that will operate application node,
   - `isOperatorInPool`- determines if the operator is in the sortition pool. If this is `undefined` it means that the operator for a given staking provider is not set.
2. Adds `updateOperatorStatus` function- updates the status of the operator in the sortition pool. If there was an authorization decrease requested, it is activated by starting the authorization decrease delay. Function reverts if the operator is not known. From the T dapp perspective, it should be called when the authorizer requests the authorization decrease and the operator is not in the sortition pool.

### UI
Updates the application card to render the `deauthorization initiation needed` status- this status means that the operator is not in the sortition pool and the deauthorization request is not active. The user must call `updateOperatorStatus` function to activate the authorization decrease delay.

<details>
  <summary>Screenshots</summary>
  
  
obraz

</details>

PR #327, opened 2022-12-07
[Backport] Missing path when decreasing authorization

r-czajkowski:
Backport of: #318

This PR adds support for decreasing authorization for non-beta operators. The operator who is not a beta operator could not join the pool. If they are deauthorizing to 0, from the sortition pool's perspective, their status is up to date (see [`RandomBeacon.isOperatorUpToDate`](https://github.com/keep-network/keep-core/blob/4e940f02b8bebeafcae2727016e00ea70355a709/solidity/random-beacon/contracts/libraries/BeaconAuthorization.sol#L551) and [`WalletRegistry.isOperatorUpToDate`](https://github.com/keep-network/keep-core/blob/4049dc0153c33c3c4cbb2e9a1808581bf76144fb/solidity/ecdsa/contracts/libraries/EcdsaAuthorization.sol#L554)). The client will not attempt to update the pool if the status is up to date. The new deauthorization flow is: when the authorization decrease is requested, we check if there is a staking provider <> operator mapping. If such mapping exists, we check if the operator is in the pool. If not, we execute a call to `RandomBeacon.updateOperatorStatus / WalletRegistry.updateOperatorStatus` to activate the authorization decrease delay.

### Threshold-ts lib
1. Updates the `getStakingProviderAppInfo` function- add additional fields to returned object:
   - `operator`- operator address that will operate application node,
   - `isOperatorInPool`- determines if the operator is in the sortition pool. If this is `undefined` it means that the operator for a given staking provider is not set.
2. Adds `updateOperatorStatus` function- updates the status of the operator in the sortition pool. If there was an authorization decrease requested, it is activated by starting the authorization decrease delay. Function reverts if the operator is not known. From the T dapp perspective, it should be called when the authorizer requests the authorization decrease and the operator is not in the sortition pool.

### UI
Updates the application card to render the `deauthorization initiation needed` status- this status means that the operator is not in the sortition pool and the deauthorization request is not active. The user must call `updateOperatorStatus` function to activate the authorization decrease delay.

<details>
  
  
obraz

</details>


michalsmiarowski (approved):
LGTM 👍

PR #330, opened 2022-12-09
Adds the suggestion page with file upload to GCP

georgeweiler:
This PR builds the suggestion page according to designs: https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=4251%3A276860&t=ZmBL64H8ytLsbgVf-0

Per UX discussion we've decided to implement an override for posthog capture events so that suggestions can still be submitted to posthog, even if the user has declined the anonymous analytics tracking.

User files are uploaded to the Google cloud bucket which are permission locked for authorized users only. The link to the file is sent to posthog.


src/hooks/useGCloudStorage.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    formData.append("file", file)
+    formData.append("key", file.name)
+
+    return fetch(`https://storage.googleapis.com/${bucket}`, {
+      method: "POST",
+      body: formData,
+    }).then((resp) => {
+      return `https://storage.cloud.google.com/${bucket}/${file.name}`
+    })
```
r-czajkowski, line 17: We already use `axios` as a http client see `utils/exchangeApi.ts`- we could use it, probably simplify code a little bit. Ofc non-blocking.

src/hooks/useGCloudStorage.ts
```diff
@@ -0,0 +1,21 @@
...
+  return useCallback(async (file) => {
+    const formData = new FormData()
+
+    formData.append("file", file)
+    formData.append("key", file.name)
+
+    return fetch(`https://storage.googleapis.com/${bucket}`, {
+      method: "POST",
+      body: formData,
+    }).then((resp) => {
```
r-czajkowski, line 15: maybe let's add `catch` to log any errors and avoid app crashes.

src/hooks/posthog/useCapture.ts
```diff
@@ -4,14 +4,24 @@ import { featureFlags } from "../../constants"
 import * as posthog from "../../posthog"
 import { useAnalytics } from "../useAnalytics"
 
-export const useCapture = (eventName: PosthogEvent) => {
+export const useCapture = (
+  eventName: PosthogEvent,
+  overrideConsent = false
```
r-czajkowski, line 9: maybe `shouldOverrideConsent`?

src/pages/Feedback/Suggestions/index.tsx
```diff
@@ -1,9 +1,118 @@
...
+      fileLink = await uploadToGCP(file)
+    }
+
+    captureSuggestionSubmit({
+      file: fileLink,
+      category: featureCategory,
+      suggestion: suggestionText,
+    })
+
+    setIsLoading(false)
```
r-czajkowski, line 51: We probably should set the loading indicator to false once we catch an error.

r-czajkowski (changes requested):
Left minor comments. Overall LGTM 👍

PR #336, opened 2022-12-22
Update unstake flow

r-czajkowski:
This PR updates the unstake flow. The maximum unstake amount depends on the authorized apps for example suppose the given staking provider has 10T, 20T worth of KEEP and 30 T worth of NU- all staked. The maximum application authorization is 40 T, then the maximum available amount to unstake is:
- 10T from T stake because: `10T(T stake) - max(0, 40T(authorization) - 20T(KEEP stake) - 30T(NU stake)) = 10`,
- 20T from KEEP stake because: `20T(KEEP stake) - max(0, 40T(authorization) - 30T(NU stake) - 10T(T stake)) = 20`,
- 20T from NU stake because: `30T(NU stake) - max(0, 40T(authorization) - 20T(KEEP stake) -10T(T stake)) = 20`,
- An owner can't unstake all stake(KEEP+NU+T) because has authorized applications. To unstake all the staking provider cannot have any authorized apps.

Here we also update a copy in the unstaking and top-up modals and add a tooltip next to the max amount to unstake(in unstake form) that says the given staking provider has authorized apps and must deauthorize first if they want to unstake more T.

<details>

obraz

</details>

### DIFF
--- a/src/pages/Staking/StakeCard/index.tsx
+++ b/src/pages/Staking/StakeCard/index.tsx
@@ -35,6 +35,8 @@ import { StakeCardContext } from "../../../contexts/StakeCardContext"
 import { useStakeCardContext } from "../../../hooks/useStakeCardContext"
 import { isSameETHAddress } from "../../../threshold-ts/utils"
 import { useWeb3React } from "@web3-react/core"
+import { useAppSelector } from "../../../hooks/store"
+import { selectAvailableAmountToUnstakeByStakingProvider } from "../../../store/staking"
 
 const StakeCardProvider: FC<{ stake: StakeData }> = ({ stake }) => {
   const isInactiveStake = BigNumber.from(stake.totalInTStake).isZero()
@@ -68,6 +70,12 @@ const StakeCard: FC<{ stake: StakeData }> = ({ stake }) => {
   const { isInactiveStake, canTopUpKepp, canTopUpNu, isPRESet } =
     useStakeCardContext()
   const { account } = useWeb3React()
+  const availableAmountToUnstake = useAppSelector((state) =>
+    selectAvailableAmountToUnstakeByStakingProvider(
+      state,
+      stake.stakingProvider
+    )
+  )
 
   const isOwner = isSameETHAddress(account ?? AddressZero, stake.owner)
 
@@ -157,7 +165,7 @@ const StakeCard: FC<{ stake: StakeData }> = ({ stake }) => {
           onSubmitForm={onSubmitForm}
           label="Amount"
           submitButtonText={submitButtonText}
-          maxTokenAmount={isStakeAction ? tBalance : stake.tStake}
+          maxTokenAmount={isStakeAction ? tBalance : availableAmountToUnstake.t}
           shouldDisplayMaxAmountInLabel
           isDisabled={!isOwner}
           submitButtonVariant="outline"
--- a/src/pages/Staking/StakeCard/index.tsx
+++ b/src/pages/Staking/StakeCard/index.tsx
@@ -17,7 +17,6 @@ import { useTokenBalance } from "../../../hooks/useTokenBalance"
 import { useModal } from "../../../hooks/useModal"
 import { StakeData } from "../../../types/staking"
 import {
-  ExternalHref,
   ModalType,
   StakeType,
   Token,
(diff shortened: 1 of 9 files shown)


src/store/staking/selectors.ts
```diff
@@ -13,3 +16,68 @@ export const selectStakeByStakingProvider = createSelector(
...
+    )
+
+    return {
+      nuInT: isZeroAuthorization
+        ? stake.nuInTStake
+        : BigNumber.from(stake.nuInTStake).sub(nuInTMinStake).toString(),
+      keepInT:
+        isZeroAuthorization || BigNumber.from(keepInT).isZero()
+          ? stake.keepInTStake
+          : "0",
```
michalsmiarowski, line 74: Why do we put `0` here and not `BigNumber.from(stake.keepInTStake).sub(keepInTMintStake).toString()` like in `nuInT` or `t`? 

Note: (I used `keepInTMintStake` in this example instead of `keepInT`. See )

src/threshold-ts/utils/math.ts
```diff
@@ -0,0 +1,9 @@
+import { BigNumber } from "ethers"
+
+export const min = (a: string | number, b: string | number) => {
+  return BigNumber.from(a).lt(b) ? a.toString() : b.toString()
+}
```
michalsmiarowski, line 5: I think it would be better to return the value with the same type as it was passed. This will make it more reusable:

```suggestion
import { BigNumber, BigNumberish } from "ethers"

export const min = (a: BigNumberish, b: BigNumberish) => {
  return BigNumber.from(a).lt(b) ? a : b
}
```

What do you think? (P.S I did not test it)

src/threshold-ts/utils/math.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export const max = (a: string | number, b: string | number) => {
+  return BigNumber.from(a).gt(b) ? a.toString() : b.toString()
+}
```
michalsmiarowski, line 9: Same as above if you agree

michalsmiarowski (changes requested):
Left some comments to look at before the merge but overall I think it looks good 🔥  I also need to test it with legacy stakes little more, so I might do another round of review soon.

src/store/staking/selectors.ts
r-czajkowski: Because you can unstake only the full KEEP stake, you can't unstake KEEP partially.

src/threshold-ts/utils/math.ts
r-czajkowski: Hmm hard to say because you can pass `a` as `number` and `b` as `string`- in that case, function returns `a` as `number` but `b` as `string`. Maybe let's convert result to `BigNumber`?

src/pages/Staking/StakeCard/index.tsx
```diff
@@ -157,7 +165,7 @@ const StakeCard: FC<{ stake: StakeData }> = ({ stake }) => {
           onSubmitForm={onSubmitForm}
           label="Amount"
           submitButtonText={submitButtonText}
-          maxTokenAmount={isStakeAction ? tBalance : stake.tStake}
+          maxTokenAmount={isStakeAction ? tBalance : availableAmountToUnstake.t}
```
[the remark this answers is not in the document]
michalsmiarowski: Maybe we could add a tooltip next to the maximum amount to unstake?:

image

This is a 45k T stake with 40k T authorized. We could add a questionmark-like icon next to the `5,000` and display a tooltip that says "If you want to unstake more T deauthorize the apps for this stake first".

What do you think and how hard do you think it would be? We should display the tooltip only if any app is authorized with some value.

src/store/staking/selectors.ts
michalsmiarowski: Ahh i see now, makes sense.

What about nu? Can we unstake it partially?

src/store/staking/selectors.ts
```diff
@@ -13,3 +16,66 @@ export const selectStakeByStakingProvider = createSelector(
...
+
+    const tMinStake = max(
+      ZERO,
+      maxAuthorization.sub(stake.nuInTStake).sub(stake.keepInTStake).toString()
+    )
+
+    const keepInTMinStake = max(
+      ZERO,
+      maxAuthorization.sub(stake.nuInTStake).sub(stake.tStake).toString()
+    )
```
michalsmiarowski, line 63: Non-blocking:

To be honest it took me a while to figure out what's going on here. Right now it seems easy but i assume if I ever had to go back to this piece of code after 1+ year it could take a while again to figure this out. Maybe we could write a comment that this calcultions are based on the `getMinStaked` method in `TokenStaking` contract? I think it have a nice explanation in the comment.

michalsmiarowski (changes requested):
Overall looks good. I tested it out with legacy Keep stake and it works wonderful. Left some small comments to look at before the merge

src/store/staking/selectors.ts
r-czajkowski: bb18553

src/pages/Staking/StakeCard/index.tsx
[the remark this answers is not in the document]
r-czajkowski: 8f8bd99

michalsmiarowski (commented):
Forgot to gave an approve but I've tested out the new changes before the merge and they look great 🔥

Issue #289, opened 2022-11-02
Action history for tbtc-v2


michalsmiarowski:
In the final version there should be three action types:
- INITIATED (instead of PENDING in the desings) - this action will show in the history after the initiation of the minting
- MINTED - this acction will show in the history if the sweep happens and the deposit is included in the sweep
- ERROR - this action will show in the history if any error happens during the minting

In the future there should be also another action:
- UNMINTED

PR #321, opened 2022-11-29
Unspent transactions check & deposit reveal

michalsmiarowski:
~Depends on: https://github.com/keep-network/tbtc-v2/pull/419~
~Depends on: https://github.com/keep-network/tbtc-v2/pull/425~
~Depends on: https://github.com/keep-network/tbtc-v2/pull/441~

Closes: #304 
Closes: #311 
Closes: #288 
Closes: #305 

Adds a whole reveal deposit functionality.

Main changes:
- refactor TBTC wrapper
- finds UTXOs of the currect connected user by his deposit address in step 2
- actual functionality to reveal a deposit or multiple deposits
- functionality that saves the deposit data in local storage, which makes it possible to get the actual data after refresh

PR #337, opened 2022-12-30
Bridge tx history

r-czajkowski:
Depends on: #321
Closes: #289 

This PR adds support for fetching bridge transaction history in order from the newest to the oldest revealed deposit.

obraz


src/threshold-ts/tbtc/index.ts
```diff
@@ -223,4 +263,105 @@ export class TBTC implements ITBTC {
...
+      )
+
+      return {
+        amount: "0",
+        walletPublicKeyHash: deposit.args?.walletPubKeyHash,
+        fundingTxHash,
+        fundingOutputIndex,
+        depositKey,
+        txHash: deposit.transactionHash,
+      }
```
michalsmiarowski, line 327: I think we should print the tx history from the newest (at the top) to the oldest. Right now it's displayed from the oldest to the newest.

It looks like it gets the `DepositRevealed` events sorted atuomatically from the oldest to newest so I think we sohuld be good by just reversing the order of the events.

The other way would be to get `refundLocktime` from the script and sort it by this value. The problematic thing might be that this value is a 4-byte little-endian un-prefixed hex string so we would probably have to change it to decimal and then sort.

src/threshold-ts/tbtc/index.ts
```diff
@@ -223,4 +263,105 @@ export class TBTC implements ITBTC {
...
+          interface: this._bridgeContract.interface,
+          method: "deposits",
+          args: [depositKey],
+        }))
+    )
+
+    return parsedDeposits.map((deposit, index) => {
+      const amount = revealedDeposits[index][0]?.amount?.toString()
+      return { ...deposit, amount }
+    })
```
michalsmiarowski, line 344: Is this really needed? 🤔 In 76f2420ae2e974f40c4c5571d59f7170cc82977d commit message you wrote:
```
`DepositRevealed` event doesn't include `amount` param so we need to
call `Brodge.deposits` function to get an `amount` for a given deposit
key.
```
but this doesn't seem to be true. `DepositRevealed` event does include amount param. Unless you meant something else?

src/pages/tBTC/Bridge/index.tsx
```diff
@@ -46,7 +62,7 @@ const TBTCBridge: PageComponent = (props) => {
       </Box>
       <Box gridArea="aside">
         <TbtcBalanceCard gridArea="balance-card" mb="5" />
-        <TransactionHistory gridArea="tx-history" mb="5" />
+        <TransactionHistory data={bridgeTransactionHistory} />
```
michalsmiarowski, line 65: We should update the data after the reveal is done so the user won't have to refresh the page to get the current tx history.

We listen for the `DepositRevealed` event here:
https://github.com/threshold-network/token-dashboard/blob/6c2941e062626160f1d331403ca0297d78de4a20/src/hooks/tbtc/useSubsribeToDepositRevealedEvent.ts#L7
so I think we could update the data there.

michalsmiarowski (changes requested):
Overall looking good, but added some comments to look at  👀

src/threshold-ts/tbtc/index.ts
r-czajkowski: You are right! We can get an amount from `DepositRevealed` event- it's `fundingOutputAmount` param. I missed this param. Thanks!

r-czajkowski: 453961d

src/threshold-ts/tbtc/index.ts
r-czajkowski: 453961d

src/pages/tBTC/Bridge/TransactionHistory/index.tsx
```diff
@@ -94,27 +66,38 @@ export const TransactionHistory: FC<ComponentProps<typeof Card>> = ({
...
             </LabelSm>
           </Tr>
         </Thead>
         <BodySm as={Tbody}>
           {isHistoryEmpty ? (
             <EmptyHistoryTableBody />
           ) : (
-            history.map((tx, index) => (
-              <Tr key={index}>
+            data.map((_) => (
```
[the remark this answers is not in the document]
r-czajkowski: 41c4d44

src/pages/tBTC/Bridge/index.tsx
r-czajkowski: 12cc6ff

michalsmiarowski (approved):
LGTM. We will address the "Expand view" button in a separate PR.

PR #339, opened 2023-01-02
Don't use tag as semver range in `solidity-contracts`

michalinacienciala:
This is a test commit. Description needs to be improved.


michalinacienciala:
This PR was created only for the testing purposes. We didn't end up using this change.

PR #340, opened 2023-01-02
Fix failing CI jobs

michalinacienciala:
We're introducing two changes that are needed to fix our failing CI jobs:

1. Using SemVer range instead of tag when describing allowed versions of
   `@keep-network/keep-ecdsa` dependency.
   Without this change the command
   ```
   yarn upgrade \
       @threshold-network/solidity-contracts \
       @keep-network/keep-core \
       @keep-network/keep-ecdsa \
       @keep-network/tbtc \
       @keep-network/coverage-pools \
       @keep-network/tbtc-v2 \
       @keep-network/tbtc-v2.ts \
       @keep-network/ecdsa \
       @keep-network/random-beacon
   ```
   or just
   ```
   yarn upgrade @keep-network/ecdsa
   ```
   is failing with the `expected manifest` error. Why is this error occuring
   (and why only there) is unclear, but we suspect this may be related to the
   bug in Yarn - https://github.com/yarnpkg/yarn/issues/4731.

2. Using resolution
   (https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) for
   `@tenderly/hardhat-tenderly` dependency.
   Without this, the command
   ```
   yarn upgrade \
       @threshold-network/solidity-contracts@dapp-development-goerli \
       @keep-network/keep-core@1.8.1-goerli.0 \
       @keep-network/keep-ecdsa@goerli \
       @keep-network/tbtc@goerli \
       @keep-network/coverage-pools@goerli \
       @keep-network/ecdsa@dapp-development-goerli \
       @keep-network/random-beacon@dapp-development-goerli \
       @keep-network/tbtc-v2@dapp-development-goerli
   ```
   is failing with the following error:
   ```
   error tslog@4.7.1: The engine "node" is incompatible with this module.
   Expected version ">=16". Got "14.21.1"
   ```
   We know that such error occurs when one of the dependencies uses
   `@tenderly/hardhat-tenderly` dependency in versions `>=1.4.x`, as `1.4.0`
   introduced a dependency to `tslog` module which is incompatible with Node.js
   v14. We updated our `development` packages to not use
   `@tenderly/hardhat-tenderly` in versions `>=1.4.x`, but we haven't done that
   yet for `goerli`/`dapp-development-goerli` packages. This is why we need to
   force usage of the lower versions using the resolutions functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@ethersproject/hardware-wallets": "^5.5.0",
     "@fontsource/inter": "^4.5.10",
     "@keep-network/coverage-pools": "development",
-    "@keep-network/ecdsa": "development",
+    "@keep-network/ecdsa": ">2.1.0-dev <2.1.0-goerli",
     "@keep-network/keep-core": "development",
     "@keep-network/keep-ecdsa": "development",
     "@keep-network/random-beacon": "development",
@@ -113,7 +113,8 @@
     "eslint-plugin-prettier": "^4.0.0",
     "babel-loader": "8.1.0",
     "react-error-overlay": "6.0.9",
-    "@types/react": "17.0.30"
+    "@types/react": "17.0.30",
+    "@tenderly/hardhat-tenderly": ">=1.0.12 <1.1.0"
   },
   "devDependencies": {
     "@chakra-ui/cli": "0.0.0-pr-20211126153854",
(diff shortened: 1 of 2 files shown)


r-czajkowski (approved):
🚀

PR #342, opened 2023-01-03
[Backport] Rewards allocation for Jan 1st 2023

r-czajkowski:
Backport of: #341

This PR adds a merkle tree computed in https://github.com/threshold-network/merkle-distribution/pull/56 to Threshold Token Dashboard. It also updates a copy in the rewards card on the staking page and updates the `useNextRewardsDropDate` hook to return always the first day of the next month because monthly staking rewards are distributed retroactively by the Threshold Council and are claimable on or about the beginning of the subsequent month.


michalsmiarowski (approved):
LGTM

PR #347, opened 2023-01-12
tBTC- Electrum support

r-czajkowski:
~Depends on: #321~

This PR adds support for connecting to the Electrum based on the provided env variables. If the `MOCK_BITCOIN_CLIENT` env variable is set to `true` the dapp will use the mock implementation of the Bitcoin client- which is useful for local development. Otherwise, the dapp will pass the following credentials to connect with the default Electrum-based implementation of the Bitcoin client:
- protocol- used by the Electrum server. Env variable: `ELECTRUM_PROTOCOL`,
- host- pointing to the Electrum server. Env variable: `ELECTRUM_HOST`,
- port- the Electrum server listens on. Env variable `ELECTRUM_PORT`.

It also refactors the way we handle env variables in T dapp- in previous implementation when adding a new environment variable, an update was required in three places in the `.ts` files. Here we improved it by using `const` array instead of an enum. Based on this array we can dynamically create a map `env name -> env value` which is used by `getEnvVariable` function. In order not to break the previous enum API here we add an `EnvVariable` object where key is the name of the env variable so using eg. `EnvVariable.ETH_HOSTNAME_HTTP` is still valid.

### QA
Went through the deposit flow with the following steps:
- Step 1:
  - set the ETH address(dapp uses the current logged account by default),
  - set the BTC recovery address to: `2MuvXjtcT5NqrHgKVopSZWyHMg7dXv5WPrA`,
  - hit the `Generate deposit address` button.
- Step 2
  - sent the BTC to a generated deposit address. Tx link [here](https://blockstream.info/testnet/tx/d6d224707aa490c239518c6284352e7255f6a8cc405693026bcbca51d8a2ff16),
  - hit `I sent the BTC` button,
  - The dapp was able to find the utxo for generated address and I was redirected to Step 3.
- Step 3
  - I confirmed and revealed deposit. Ethereum tx [here](https://goerli.etherscan.io/tx/0x4038308d5ab4916beb4ec7745d3a20af82207c5ea7eb60cbaa0e4880cd3c4ef6#eventlog).
 
The `fundingTxHash` param from [`DepositReveledEvent`](https://goerli.etherscan.io/tx/0x4038308d5ab4916beb4ec7745d3a20af82207c5ea7eb60cbaa0e4880cd3c4ef6#eventlog) is `0x16ffa2d851cacb6b02935640cca8f655722e3584628c5139c290a47a7024d2d6`. After conversion from [little endian](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/Deposit.sol#L227) to big-endian it's equal to `D6D224707AA490C239518C6284352E7255F6A8CC405693026BCBCA51D8A2FF16` which is transaction id- see https://blockstream.info/testnet/tx/D6D224707AA490C239518C6284352E7255F6A8CC405693026BCBCA51D8A2FF16


michalinacienciala (commented):
I reviewed the changes in the three modified files under `.github/workflows/`. Passing of the secrets to environment variables looks good 👍

michalsmiarowski (approved):
Looking great! I tested it out using testnet bitcoin wallet and it works fantastic! The dapp found the utxo and revealed it successfuly!

lukasz-zimnoch (approved):

PR #349, opened 2023-01-16
tBTCv2 UI improvements

r-czajkowski:
Depends on: #321 

This PR fixes UI bugs based on the [feedback from QA](https://coda.io/d/Building-Keep_d-fmEgBNFVH/2023-01-11-QA-tBTC-v2_subq6#_luufw). Besides, it adds a new modal to the tBTC deposit form- this modal should be displayed when the user clicks `back` button in Step 2 of the deposit form. This modal asks users if they want to generate a new deposit address and informs them about the consequences. If a user confirms generating a new deposit address, they will be redirected to Step 1 and the deposit data will be removed from local storage. If a user closes this modal it means they want to use already generated deposit address.

### DIFF
--- a/src/components/Modal/TbtcRecoveryFileModal/index.tsx
+++ b/src/components/Modal/TbtcRecoveryFileModal/index.tsx
@@ -14,13 +14,12 @@ import InfoBox from "../../InfoBox"
 import { BaseModalProps } from "../../../types"
 import btcJsonFile from "../../../static/images/tbtc-json-file.png"
 import withBaseModal from "../withBaseModal"
-import ViewInBlockExplorer from "../../ViewInBlockExplorer"
-import { ExplorerDataType } from "../../../utils/createEtherscanLink"
 import { useTbtcState } from "../../../hooks/useTbtcState"
 import { DepositScriptParameters } from "@keep-network/tbtc-v2.ts/dist/deposit"
 import { MintingStep } from "../../../types/tbtc"
 import { downloadFile } from "../../../web3/utils"
-import { useTBTCBridgeContractAddress } from "../../../hooks/useTBTCBridgeContractAddress"
+import Link from "../../Link"
+import { ExternalHref } from "../../../enums"
 
 const TbtcRecoveryFileModalModal: FC<
   BaseModalProps & {
@@ -29,7 +28,6 @@ const TbtcRecoveryFileModalModal: FC<
 > = ({ closeModal, depositScriptParameters }) => {
   const { isOpen: isOnConfirmStep, onOpen: setIsOnConfirmStep } =
     useDisclosure()
-  const bridgeContractAddress = useTBTCBridgeContractAddress()
   const { updateState } = useTbtcState()
 
   const handleDoubleReject = () => {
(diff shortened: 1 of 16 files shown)


src/pages/tBTC/Bridge/MintingCard/MintingTimeline.tsx
```diff
@@ -25,7 +27,16 @@ export const MintingTimelineStep1: FC<MintingTimelineStepProps> = ({
       stepText="Step 1"
       helperLabelText="OFF-CHAIN ACTION"
       title="Provide Data"
-      description="Provide an ETH address and a BTC Recovery address to generate an unique BTC deposit address."
+      description={
+        <>
+          Provide an ETH address and a BTC Recovery address to generate an
+          unique BTC deposit address. {/* TODO: set correct link here */}
```
michalsmiarowski, line 33: Let's not forget about this TODO

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
```diff
@@ -71,9 +74,16 @@ const MakeDepositComponent: FC = () => {
     }
   }
 
+  const onPreviousStepClick = () => {
+    openModal(ModalType.GenerateNewDepositAddress)
+  }
+
```
michalsmiarowski, line 80: After merging `main` to this branch we should move the opening of the modal to here:

https://github.com/threshold-network/token-dashboard/blob/6c2941e062626160f1d331403ca0297d78de4a20/src/pages/tBTC/Bridge/MintingCard/MintingFlowRouter.tsx#L21-L26

and put it inside the if.

src/components/Modal/TbtcRecoveryFileModal/index.tsx
```diff
@@ -38,11 +43,14 @@ const TbtcRecoveryFileModalModal: FC<
...
-    downloadFile(
-      JSON.stringify(data),
-      "deposit-script-parameters.json",
-      "text/json"
-    )
+    const date = new Date()
+    const fileName = `
+    ${ethAddress}-
+    ${btcDepositAddress}-
+    ${date.getDate()}/${date.getMonth() + 1}/${date.getFullYear()}
```
michalsmiarowski, line 50: New lines makes the file name weird (so many spaces). Let's move it to one line and format the code. It should look something like this:

```suggestion
    const fileName = `${ethAddress}-${btcDepositAddress}-${date.getDate()}/${
      date.getMonth() + 1
    }/${date.getFullYear()}`
```

michalsmiarowski, line 50: The [international standard](https://www.iso.org/iso-8601-date-and-time-format.html) for writing dates is YYYY-MM-DD so i think we should write it like this. For example:

`<address>_<deposit_address>_2023-01-18.json`

Note that there should be `01`, not just `1`.

michalsmiarowski (changes requested):
Left some comments to look at. Let's merge the `main` branch to `tbtc-v2-ui-improvements` before the final testing.

src/pages/tBTC/Bridge/MintingCard/MintingTimeline.tsx
r-czajkowski: We don't have the exact link yet but I think the current one is OK.

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
r-czajkowski: 7838fc2

src/components/Modal/TbtcRecoveryFileModal/index.tsx
r-czajkowski: d79cc94

r-czajkowski: d79cc94

michalsmiarowski (approved):
LGTM 👨‍🎤

PR #351, opened 2023-01-18
Update tbtc imports

michalsmiarowski:
The structure of the files in the `tbtc-v2.ts` package is now a bit different than before. Instead of `keep-network/tbtc-v2.ts/dist/deposit` we now have `@keep-network/tbtc-v2.ts/dist/src/deposit`, so there is additional `src` directory there.

### Main changes:
- update the `tbtc-v2.ts` lib version to `[1.0.0-dev.22]`
- update the imports of the `tbtc-v2.ts` lib


r-czajkowski (approved):

src/threshold-ts/tbtc/index.ts
```diff
@@ -5,7 +5,7 @@ import {
   revealDeposit as tBTCRevealDeposit,
   getRevealedDeposit as tBTCgetRevealedDeposit,
   RevealedDeposit,
-} from "@keep-network/tbtc-v2.ts/dist/deposit"
+} from "@keep-network/tbtc-v2.ts/dist/src/deposit"
```
nkuba, line 8: Would it be possible to import from `@keep-network/tbtc-v2` instead of the `dist` directory?
Most of the things are exported via index:
https://github.com/keep-network/tbtc-v2/blob/main/typescript/src/index.ts

r-czajkowski (approved):

PR #355, opened 2023-01-19
WSS connection workaround for Electrum client

michalsmiarowski:
Electrum client does not work properly with WSS connection. As a workaround we will pass the `options` object to the client with `rejectUnauthorized` flag set to `false`

### DIFF
--- a/src/threshold-ts/types/index.ts
+++ b/src/threshold-ts/types/index.ts
@@ -1,5 +1,8 @@
 import { Client } from "@keep-network/tbtc-v2.ts/dist/src/bitcoin"
-import { Credentials } from "@keep-network/tbtc-v2.ts/dist/src/electrum"
+import {
+  ClientOptions,
+  Credentials,
+} from "@keep-network/tbtc-v2.ts/dist/src/electrum"
 import { providers, Signer } from "ethers"
 
 export enum BitcoinNetwork {
@@ -15,7 +18,7 @@ export interface EthereumConfig {
 
 export type BitcoinClientCredentials = Credentials
 
-export type BitcoinClientOptions = object
+export type BitcoinClientOptions = ClientOptions
 
 export interface BitcoinConfig {
   /**
--- a/src/threshold-ts/types/index.ts
+++ b/src/threshold-ts/types/index.ts
@@ -15,7 +15,7 @@ export interface EthereumConfig {
 
 export type BitcoinClientCredentials = Credentials
 
-export type BitcoinClientConfig = object
+export type BitcoinClientOptions = object
 
 export interface BitcoinConfig {
   /**
@@ -38,7 +38,7 @@ export interface BitcoinConfig {
   /**
    * Additional options that can be passed to bitcoin client
    */
-  clientOptions?: BitcoinClientConfig
+  clientOptions?: BitcoinClientOptions
 }
 
 export interface ThresholdConfig {
--- a/src/threshold-ts/types/index.ts
+++ b/src/threshold-ts/types/index.ts
@@ -15,6 +15,8 @@ export interface EthereumConfig {
 
 export type BitcoinClientCredentials = Credentials
 
+export type BitcoinClientConfig = object
+
 export interface BitcoinConfig {
   /**
    * Indicates for which network the addresses (eg deposit address) should be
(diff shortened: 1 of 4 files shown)


src/threshold-ts/types/index.ts
```diff
@@ -15,6 +15,8 @@ export interface EthereumConfig {
 
 export type BitcoinClientCredentials = Credentials
 
+export type BitcoinClientConfig = object
```
nkuba, line 18: Aren't we duplicating the code with `@keep-network/tbtc-v2.ts` library? Couldn't we just import the `Credentials` and `Options` types from it?

https://github.com/keep-network/tbtc-v2/blob/50c38ec0a7951ac8a3e72af6c9bc2d9dc882a415/typescript/src/electrum.ts#L20-L38

src/utils/getThresholdLib.ts
```diff
@@ -31,6 +31,8 @@ function getBitcoinConfig(): BitcoinConfig {
     client: shouldMockBitcoinClient ? new MockBitcoinClient() : undefined,
     network,
     credentials: !shouldMockBitcoinClient ? credentials : undefined,
+    // FIXME: It's a temporary workaround to get the connection working.
```
nkuba, line 34:
```suggestion
    // FIXME: It's a temporary workaround to get the WSS connection working.
```

src/utils/getThresholdLib.ts
```diff
@@ -31,6 +31,8 @@ function getBitcoinConfig(): BitcoinConfig {
...
+    clientOptions: { rejectUnauthorize: false },
```
nkuba, line 35: We will need to set the options for `wss` only. So maybe we could reflect that in the code?
```suggestion
    clientOptions: credentials.protocol === "wss" ? { rejectUnauthorize: false } : undefined,
```

src/threshold-ts/types/index.ts
michalsmiarowski: I believe that we assign `Credentials` type to our custom `BitcoinClientCredentials` type in case we will change the object that we want to use. For example if we decide to change the name from `Credentials` to `ClientCredentials` in `tbtc-v2.ts` lib we will only have to change it in one place instead of every occurence in the dApp (an we use it in some other places too).

I believe it would be wise to assign `Options` type to the `BitcoinClientConfig` but I actually didn't export it in https://github.com/keep-network/tbtc-v2/pull/486 😢 If you whink it is worth it to create separate PR for that I can do it!

I renamed the type name though. It is now `BicointClientOptions` instead of `BitcoinClientConfig`. You can see the change in 5729e6ff476ed77e2650e1a77e45723c7b312be0

src/utils/getThresholdLib.ts
michalsmiarowski: 71aff73bdb799a9522da89a53df93f87d11f8699

src/utils/getThresholdLib.ts
michalsmiarowski: 71aff73bdb799a9522da89a53df93f87d11f8699

src/threshold-ts/types/index.ts
nkuba: > I believe that we assign Credentials type to our custom BitcoinClientCredentials type in case we will change the object that we want to use. For example if we decide to change the name from Credentials to ClientCredentials in tbtc-v2.ts lib we will only have to change it in one place instead of every occurence in the dApp (an we use it in some other places too).

Couldn't we do something like `import type { Credentials as BitcoinClientCredentials } from "@keep-network/tbtc-v2.ts"` ?

> If you think it is worth it to create separate PR for that then I can do it!

Please do 🙏🏻 I'll do a turbo-review-and-merge for you!

nkuba: > > I believe that we assign Credentials type to our custom BitcoinClientCredentials type in case we will change the object that we want to use. For example if we decide to change the name from Credentials to ClientCredentials in tbtc-v2.ts lib we will only have to change it in one place instead of every occurence in the dApp (an we use it in some other places too).
> 
> Couldn't we do something like `import type { Credentials as BitcoinClientCredentials } from "@keep-network/tbtc-v2.ts"` ?

OK, now I get your point. It makes sense!

michalsmiarowski: https://github.com/keep-network/tbtc-v2/pull/487

michalinacienciala:
The failing checks are likely connected to the GH outage: https://www.githubstatus.com/incidents/n18j5g7wpg0s

src/threshold-ts/types/index.ts
michalsmiarowski: 956fbb25e00134c600091b8c8c65413e6bb815f5

michalinacienciala:
I've re-run the tests and they're passing now

nkuba (approved):

PR #359, opened 2023-01-25
Update tbtc-v2.ts lib to 1.0.0-dev.29

michalsmiarowski:
Update tbtc-v2.ts lib to `1.0.0-dev.29` version. This also requires to update all occurences of `Identifier` type in the dApp, because there was a change in tbtc-v2.ts lib that added `equals` function to the `Identifier`. The change was introduced here: https://github.com/keep-network/tbtc-v2/pull/460

The errors were fixed by creating the Identifier from the given ethereum address using `EthereumAddress.from(<eth_address>)`.

#### Change the way we pass props to filerecovery modal

After the changes there was an error in the web console:

`A non-serializable value was detected in the state, in the path: modal.props.depositScriptParameters.depositor`.

It happens because when we open the modal we pass the parameters through the state and in this case we are opening `TbtcRecoveryFileModalModal`. A non-serializable value in this case is `equals` function that we passed with
`depositScriptParameters` property.

To fix that I've changed the way how we pass props to the modal - we pass each parameter on it's own instead of the whole `DepositScriptParameter` object as one property. So now instead of passing `Identifier` type we pass just an eth
address and we change it to `Identifier` inside the modal using `EthereumAddress.from(ethAddress)`.

#### Create `getChainIdentifier` util function

We use `EthereumAddress.from()` function in many places in the dApp. This is why I've introduced `getChainIdentifier` function that return the `Identifier` from it (or to be more technical, an `EthereumAddress` type that extends the
`Identifier` type). This might be helpful in the future if we change the way we calculate that identifier - we will only need to change the body of the `getChainIdentifier` function.

### DIFF
--- a/src/threshold-ts/utils/bitcoin.ts
+++ b/src/threshold-ts/utils/bitcoin.ts
@@ -1,3 +1,4 @@
+import { EthereumAddress } from "@keep-network/tbtc-v2.ts"
 import { Network, validate } from "bitcoin-address-validation"
 
 export const isValidBtcAddress = (
@@ -5,4 +6,8 @@ export const isValidBtcAddress = (
   network: Network = Network.mainnet
 ): boolean => {
   return validate(address, network)
+}
+
+export const getChainIdentifier = (ethAddress: string): EthereumAddress => {
+  return EthereumAddress.from(ethAddress)
 }
--- a/src/threshold-ts/utils/bitcoin.ts
+++ b/src/threshold-ts/utils/bitcoin.ts
@@ -1,4 +1,3 @@
-import { EthereumAddress } from "@keep-network/tbtc-v2.ts"
 import { Network, validate } from "bitcoin-address-validation"
 
 export const isValidBtcAddress = (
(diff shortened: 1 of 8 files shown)


src/threshold-ts/utils/bitcoin.ts
```diff
@@ -6,3 +7,7 @@ export const isValidBtcAddress = (
 ): boolean => {
   return validate(address, network)
 }
+
+export const getChainIdentifier = (ethAddress: string): EthereumAddress => {
+  return EthereumAddress.from(ethAddress)
+}
```
r-czajkowski, line 13: Shouldn't we call this under the hood in `TBTC` class and pass "raw" format? Also, I'm not sure if `bitcoin.ts` file is the right place for a function related to Ethereum chain.

r-czajkowski (changes requested):

src/threshold-ts/utils/bitcoin.ts
michalsmiarowski: Moved `getChainIdentifier` to the `utils/address.ts` file: d23594ccad322c5e40ee3d4a6ebb0259f668ff8f

I think you are right and we should convert the eth and bitcoin addresses under the hood inside the tbtc wrapper. I'll address it in a separate PR.

r-czajkowski (approved):
👍

Issue #356, opened 2023-01-20
Add telemetry to the deposit flow


lukasz-zimnoch:
We need to add some telemetry that will log each deposit from the dApp to a server of ours to help people here in case a reveal fails. We should also collect errors. A possible solution that can be used is [Sentry](https://sentry.io)

PR #361, opened 2023-01-25
Sentry-based deposit telemetry

lukasz-zimnoch:
Closes: #356

Here we add a telemetry mechanism based on Sentry. Responsibilities of this telemetry are:
- Recording all generated deposit data sets in order to facilitate further reveals that haven't happened automatically as well as possible refunds
- Recording all unhandled errors thrown by the dashboard. This is an out-of-the-box feature of Sentry SDK

We are also integrating Sentry switches with the existing CI workflows, in order to enable it on production and test environments automatically.

**Important note**: Popular ad blockers and privacy-focused browsers (e.g. Brave) may block requests targeting Sentry servers and cause some blind spots in the collected data as a result.


src/hooks/sentry/useSentry.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect } from "react"
+import { featureFlags } from "../../constants"
+import * as sentry from "../../sentry"
+
+export const useSentry = () => {
+  useEffect(() => {
```
mhluongo, line 6: Not necessary for this PR, but it's a little annoying that multiple uses of `useSentry()` throughout the component tree will call `sentry.init()` multiple times.

mhluongo (approved):
Hell yeah. I haven't tested it locally but strong concept ACK

mhluongo:
Leaving the merge to someone more familiar with the rest of the codebase.

src/hooks/sentry/useSentry.ts
lukasz-zimnoch: Looks like `[]` we pass as `useEffect`'s second param prevents that. I checked that during development.

src/sentry/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const captureMessage = (
+  message: string,
+  params: { [key: string]: unknown }
+) => {
+  Sentry.withScope((scope) => {
+    // eslint-disable-next-line guard-for-in
+    for (const key in params) {
+      scope.setExtra(key, params[key])
+    }
```
r-czajkowski, line 24: What do you think about this?
```ts
Object.entries(params).forEach(([key, value]) => scope.setExtra(key, value))
```
This way we can get rid of the`// eslint-disable-next-line guard-for-in`.

r-czajkowski (changes requested):
Left minor comments. Overall 🔥

src/hooks/tbtc/useDepositTelemetry.ts
```diff
@@ -0,0 +1,14 @@
...
+import { DepositScriptParameters } from "@keep-network/tbtc-v2.ts/dist/src/deposit"
+
+export const useDepositTelemetry = () => {
+  const captureMessage = useCaptureMessage()
+
+  return useCallback(
+    (deposit: DepositScriptParameters, depositAddress: string) => {
+      captureMessage(`Generated deposit [${depositAddress}]`, deposit)
+    },
+    []
```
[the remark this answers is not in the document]
lukasz-zimnoch: Hasn't an empty array `[]` the same effect in this case?

src/sentry/index.ts
lukasz-zimnoch: I was thinking about it but the for-loop looks far more readable so decided to go that way.

src/sentry/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getEnvVariable } from "../utils/getEnvVariable"
+import { EnvVariable } from "../enums"
+
+export const init = () => {
+  const dsn = getEnvVariable(EnvVariable.SENTRY_DSN)
+
+  Sentry.init({
+    dsn: dsn,
+    integrations: [new BrowserTracing()],
+    tracesSampleRate: 0.5,
```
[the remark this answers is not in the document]
michalsmiarowski: I tested it out locally and it looks that 100% of the messages are captured anyway. So I'm not sure what this options exactly does 🤔

michalsmiarowski (changes requested):
Looking good! Left some comments to look at before the merge!

I tested out if the sentry catches the message and it works wonderful!

src/sentry/index.ts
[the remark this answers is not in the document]
lukasz-zimnoch: `tracesSampleRate` determines what percentage of transactions is collected for performance tracing. A transaction in Sentry is for example a page load, page navigation, or an asynchronous task. This data is collected out of the box by enabling the `BrowserTracing` plugin. 

The errors and messages (the thing we use to capture deposits) are something different though. The percentage of collected errors/messages is set using the [`sampleRate`](https://docs.sentry.io/platforms/javascript/guides/react/configuration/options/#sample-rate) parameter whose default value is `1.0` so `100%` of events is collected.

src/hooks/sentry/useSentry.ts
mhluongo: I believe the empty array only prevents that at a single call-site with `useEffect`... so this won't get called every render. If it appears multiple times in the component hierarchy, you'll still have multiple calls.

To get full idempotency, I think we'd need to use a context provider or global state. I can share a proof of concept layer, but again, not a blocker right now.

src/sentry/index.ts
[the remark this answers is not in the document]
mhluongo: Trace sampling is for full stacktracesand performance stuff. If we send everything it can get expensive / slow.

I believe explicit events like deposits will be sent regardless.

r-czajkowski (approved):

src/sentry/index.ts
r-czajkowski: I think we can also use `scope.setExtras(params)` but I'm not sure because I can't find `setExtras` in docs. Ofc non-blocking.

michalsmiarowski (approved):
LGTM 🐉

r-czajkowski (approved):

pdyraga (approved):

Issue #362, opened 2023-01-26
(MEDIUM) Wrong links to the `Bridge` contract in minting flow


michalinacienciala:
The link to the `Bridge` contract on the `Recovery .JSON file` and `Take note` pop-ups are incorrect. They link to https://github.com/keep-network/tbtc-v2/blob/853bc8ff0601c82bb02cfaa0c904ba98d7d92fe5/solidity/contracts/bridge/Bridge.sol This revision of the Bridge.sol file is not current, it differs from what we have on the main branch.

The link to the `Bridge` contract is also incorrect on the `Initiate minting tBTC` pop-up. It links to https://goerli.etherscan.io/address/NEED%20BRIDGE%20CONTRACT%20ADDRESS.

In all those places we should link to the deployed Bridge contract on the Etherscan, as we do in a couple of other places.

r-czajkowski:
Figma links redirect to GitHub. We should definitely unify the Bridge contract links, but which link should we choose? I prefer the etherscan link as we do this for other contract links. Also on etherscan we can check the contract code and the etherscan link always points to the current version of the contract on mainnet.

michalinacienciala:
> which link should we choose?  I prefer the etherscan link [...]

Yep, I think linking to deployed contract on Etherscan is the way to go.

PR #374, opened 2023-01-27
Unify Bridge Contract links

r-czajkowski:
Closes: #362 

Each link should redirect to the Etherscan block explorer.


michalsmiarowski (dismissed):
Left two non-blocking comments but overall looks good!

Since they are not blocking I'm giving the approve. 🔥

michalsmiarowski (approved):
LGTM 🔥

Issue #367, opened 2023-01-26
(MAJOR) Error With Button Searching if Funds Have Been Sent - Stuck, But I Already Minted Successfully


liz-shinn:
(( PENDING - this may have been the intended interaction here ))

⛔️ If I refresh, re-connect my wallet, and come back to the Dapp, it gets hung on Step 2 and won't advance.

I have already minted successfully and the Tx History table shows that I have already minted.

<img width="570" alt="Screenshot 2023-01-26 at 5 51 39 PM" src="https://user-images.githubusercontent.com/57226633/214898810-e18f2d54-cd57-404f-b2d1-84850f75ae0e.png">

Even though the Tx History shows correctly that I have successfully minted:

<img width="338" alt="Screenshot 2023-01-26 at 5 56 43 PM" src="https://user-images.githubusercontent.com/57226633/214899165-8614eb28-b607-4f10-af24-d50d2faa1053.png">

michalsmiarowski:
We decided to change how the flow works:

**1.** User generates deposit address, download JSON, disconnects at step 2 and reconnects again (did not reveal the deposit yet)

**Solution:** Show him step 2

**2.** User generates deposit address, download JSON, sends the funds to the deposit address and then disconnects at step 3 and reconnects again (did not reveal the deposit yet)

**Solution:** Show him step 3

**3.** User generates deposit address, download JSON, sends the funds to the deposit address and then reveal the deposit and sees the success modal (step 4). He then disconnects at step 4 and reconnects again

**Solution:** Show him step 1 because he already revealed the deposit

**Additional notes:**

Do not allow the user to reveal multiple deposits that were sent to the same deposit address. The dApp should only reveal the oldest un-revealed deposit transaction (sent to this deposit address) after clicking the minting button at step 3.

**The back button:**

**1.** Clicking the back button on step 2 will display the New Generation address modal and take user to step 1

**2.** Clicking the back button on step 3 will display the New Generation address modal and the user to step 1 (he already send funds at this point so no need to navigate him back to step 2)

**3.** Clicking Dismiss or back button at step 4 will navigate the user to the step 1 without displaying the New Generation address modal

michalsmiarowski:
I will additionaly add here what @sashatanase proposed:

When the dApp recognizes that the funds were send to the deposit address in step 2 we will automatically navigate the user to step 3 without the need to click `I sent BTC` button.

Issue #369, opened 2023-01-26
(MEDIUM) Deposit flow doesn't react to sent funds when there is an earlier pending deposit


michalinacienciala:
The deposit gets stuck in the `Checking if funds have been sent...` state even after the fund have been sent. This happens if there is an earlier deposit connected to the Ethereum wallet which is in the `PENDING` state. Once the earlier deposit changes state to `MINTED`, the newer deposit unblocks (although first dApp needs to be refreshed) and `I sent the BTC` button appears.

Steps:
1. Connect to dApp with a wallet
2. Go to `tBTC` and create & reveal deposit. Deposit gets listed in the `TX HISTORY` panel as `PENDING`.
3. Using the same Ethereum account, soon after deposit is revealed, use the ◀️ to go back and confirm that you want to generate a new deposit. Fill the `Recovery BTC Address` field (can be a new value) and progress further.
4. After you sent the BTC to the deposit address there's no change in the dApp. The dApp is stuck on `Checking if funds have been sent...`
5. Wait until first deposit gets minted
6. Now you can progress with the second deposit - `I sent the BTC` option gets activated.

Expected: We should be able to progress with the second deposit even if first is still `PENDING`.

pdyraga:
Two questions:

1. What do you think about changing the copy from `Checking if funds have been sent...` to `Waiting for funds to be sent...`? 

    I think this is what is actually happening under the hood. Having the spinner next to `Checking ...`, as a user, I have a feeling there is something wrong with server connection that makes it problematic to check the state. cc @sashatanase 

2. How do we currently check if funds have been sent? Based on the issue description, deposits were generated to two different P2SH addresses so there should be no problem in distinguishing them @michalsmiarowski @r-czajkowski

michalsmiarowski:
@michalinacienciala I flagged this task in https://github.com/threshold-network/token-dashboard/pull/377 but to be honest I'm not sure if it resolves the issue you mentioned. I dind't find time to test it properly. If you encounter this bug again during testing please let me know 😉

michalinacienciala:
Ok, I'll keep an eye on it!

michalinacienciala:
@michalsmiarowski, I wasn't able to reproduce it recently, maybe it got fixed in #377 after all. I'll close the issue and if we see it occur in the future, we can reopen.

PR #377, opened 2023-01-27
Improve deposit flow

michalsmiarowski:
Closes: #367 
Ref: #369 

We decided to change how the flow works:

**1.** User generates deposit address, download JSON, disconnects at step 2 and reconnects again (did not reveal the deposit yet)

**Solution:** Show him step 2

**2.** User generates deposit address, download JSON, sends the funds to the deposit address and then disconnects at step 3 and reconnects again (did not reveal the deposit yet)

**Solution:** Show him step 3

**3.** User generates deposit address, download JSON, sends the funds to the deposit address and then reveal the deposit and sees the success modal (step 4). He then disconnects at step 4 and reconnects again

**Solution:** Show him step 1 because he already revealed the deposit

**Additional notes:**

Do not allow the user to reveal multiple deposits that were sent to the same deposit address. The dApp should only reveal the oldest un-revealed deposit transaction (sent to this deposit address) after clicking the minting button at step 3.

**The back button:**

**1.** Clicking the back button on step 2 will display the New Generation address modal and take user to step 1

**2.** Clicking the back button on step 3 will display the New Generation address modal and the user to step 1 (he already send funds at this point so no need to navigate him back to step 2)

**3.** Clicking Dismiss or back button at step 4 will navigate the user to the step 1 without displaying the New Generation address modal

#### 1. Moving most of the logic in `MakeDeposit` component to `MintingFlowRouter`

We want to check if the funds are sent and if the utxo is revealed in the
component that is a parent to the minting steps component when the user
connects. This way we could display the proper step based on the actual data of
the user.

#### 2. Clearing the interval after the utxos are found

We clear the interval that search for the utxos after we found them in a given
deposit address. We won't need that interval anymore since we don't want to
promote sending multiple transactions to one deposit address. After we find one
we go with the flow and decide which step should be displayed to the user - if
it's not revealed then step 3. If it's revealed then it's step 4(TODO: change it
to step 1 + local storage data clean).

#### 3. Operate on only one UTXO instead of array of UTXOs

We will only operate on the oldest UTXO from the array returned from
`findAllUnspentTransactionOutputs`.

#### 5. Remove unnecessary updates of minting steps in the state


src/components/Modal/TbtcMintingConfirmationModal/index.tsx
```diff
@@ -71,19 +68,12 @@ const TbtcMintingConfirmationModal: FC<TbtcMintingConfirmationModalProps> = ({
       refundLocktime,
     }
 
-    const successfulTransactions = await revealMultipleDeposits(
-      utxos,
-      depositScriptParameters
-    )
+    const transaction = await revealDeposit(utxo, depositScriptParameters)
```
r-czajkowski, line 71: We can just `await revealDeposit(utxo, depositScriptParameters)`. We don't use `transaction` anywhere.

src/components/Modal/TbtcMintingConfirmationModal/index.tsx
```diff
@@ -71,19 +68,12 @@ const TbtcMintingConfirmationModal: FC<TbtcMintingConfirmationModalProps> = ({
...
   }
 
   useEffect(() => {
     const getEstimatedFees = async () => {
-      const amount = utxos.reduce(
-        (accumulator, currentValue) =>
-          BigNumber.from(accumulator).add(currentValue.value),
-        BigNumber.from(0)
-      )
+      const amount = BigNumber.from(utxo.value)
```
r-czajkowski, line 76: Let's extract `const amount = BigNumber.from(utxo.value)` outside the `useEffect` and add to dependency array(and other missing dependencies). See https://github.com/threshold-network/token-dashboard/pull/378/commits/4eb391a906b9a97bf5acd15e55a37273e55b219f

src/components/Modal/TbtcRecoveryFileModal/index.tsx
```diff
@@ -45,7 +44,6 @@ const TbtcRecoveryFileModalModal: FC<
   const { updateState } = useTbtcState()
 
   const handleDoubleReject = () => {
-    updateState("mintingStep", MintingStep.Deposit)
```
r-czajkowski, line 48: Unused `const { updateState } = useTbtcState()` in line 44.

src/components/Modal/tBTC/GenerateNewDepositAddress.tsx
```diff
@@ -28,7 +28,6 @@ const GenerateNewDepositAddressBase: FC<BaseModalProps> = ({ closeModal }) => {
 
   const onConfirmClick = () => {
     removeDepositDataFromLocalStorage()
-    resetDepositData()
```
r-czajkowski, line 31: Are we sure? Why don't we reset the deposit data once someone wants to generate a new deposit address?

r-czajkowski: If we are sure then let's remove ` const { resetDepositData } = useTbtcState()` and import `useTbtcState`.

src/components/Modal/TbtcRecoveryFileModal/index.tsx
r-czajkowski: and let's remove import.

src/hooks/tbtc/useTBTCDepositDataFromLocalStorage.ts
```diff
@@ -36,6 +38,7 @@ export const useTBTCDepositDataFromLocalStorage = () => {
     }
     delete newLocalStorageData[`${account}`]
     setTBTCDepositData(newLocalStorageData)
+    resetDepositData() // additionaly remove the deposit data from the store
```
r-czajkowski, line 41: I would wrap those 2 hooks into a single one. IMO looks weird inside this `useTBTCDepositDataFromLocalStorage`. Eg
`useRemoveDepositData.`
```
const useRemoveDepositData() => {
  const { resetDepositData } = useTbtcState()
    const { removeDepositDataFromLocalStorage } =
    useTBTCDepositDataFromLocalStorage()
    
    return useCallback(() => {
      resetDepositData()
      removeDepositDataFromLocalStorage()
    })
}
```

Or remove deposit from local storage inside `useTbtcState` hook.

src/pages/tBTC/Bridge/MintingCard/InitiateMinting.tsx
```diff
@@ -11,20 +11,19 @@ import withOnlyConnectedWallet from "../../../../components/withOnlyConnectedWal
 import { UnspentTransactionOutput } from "@keep-network/tbtc-v2.ts/dist/src/bitcoin"
 
 const InitiateMintingComponent: FC<{
-  utxos: UnspentTransactionOutput[] | undefined
+  utxo: UnspentTransactionOutput | undefined
```
r-czajkowski, line 14: why `undefined` ? I think we should only redirect to step 3 if we find the utxo.

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
```diff
@@ -39,35 +39,10 @@ const AddressRow: FC<{ address: string; text: string }> = ({
 }
 
 const MakeDepositComponent: FC<{
-  utxos: UnspentTransactionOutput[] | undefined
```
r-czajkowski, line 42: There are some unused imports in this file. Let's remove.

src/pages/tBTC/Bridge/MintingCard/MintingFlowRouter.tsx
```diff
@@ -41,7 +47,21 @@ const MintingFlowRouterBase = () => {
...
+          const utxo = utxos.pop()
+          setUtxo(utxo)
+          // If there is at least one utxo we remove the interval, because we
+          // don't want to encourage sending multiple transactions to the same
+          // deposit address.
+          clearInterval(interval)
+        } else {
+          // If ther is no utxo then we display the `MakeDeposit` step to the
+          // user and still run the interval to check if the funds were sent.
+          updateState("mintingStep", MintingStep.Deposit)
```
r-czajkowski, line 63: Let's add `threshold` and `updateState` to the dependency array.

src/pages/tBTC/Bridge/MintingCard/MintingFlowRouter.tsx
```diff
@@ -53,19 +73,35 @@ const MintingFlowRouterBase = () => {
...
+
+      if (isDepositRevealed) {
+        updateState("mintingStep", MintingStep.ProvideData)
+        removeDepositDataFromLocalStorage()
+      } else {
+        updateState("mintingStep", MintingStep.InitiateMinting)
+      }
+    }
+    checkIfUtxoIsRevealed()
+  }, [utxo])
```
r-czajkowski, line 92: Can we do this in the `useEffect` from above? The `utxo` is an object so this effect will be called on every render.

r-czajkowski (changes requested):

src/components/Modal/TbtcMintingConfirmationModal/index.tsx
michalsmiarowski: c848ac024c9ba34761f70c32e32ff9e4c409d14c

src/components/Modal/TbtcMintingConfirmationModal/index.tsx
michalsmiarowski: 0ee520757ac4308dc349d3cfe18b9faae497366f

src/components/Modal/TbtcRecoveryFileModal/index.tsx
michalsmiarowski: 266552ddac5230ac5bf2a3fca9b125dfc7cc33fa

src/hooks/tbtc/useTBTCDepositDataFromLocalStorage.ts
michalsmiarowski: 2c75eb02480776f97bf9abc28f2cf0c78ce41ae5

src/components/Modal/tBTC/GenerateNewDepositAddress.tsx
michalsmiarowski: See https://github.com/threshold-network/token-dashboard/pull/377#discussion_r1089291288

src/pages/tBTC/Bridge/MintingCard/InitiateMinting.tsx
michalsmiarowski: 6a9225e58ed310dca65870f173ea700b0b49e43c

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
michalsmiarowski: 25e3a5a133799f40d7d87a663414e9d6ad20a923

src/pages/tBTC/Bridge/MintingCard/MintingFlowRouter.tsx
michalsmiarowski: e33c995660a2b9ed0be2efbdf09451abc5032bb0

src/pages/tBTC/Bridge/MintingCard/MintingFlowRouter.tsx
michalsmiarowski: See https://github.com/threshold-network/token-dashboard/pull/377#discussion_r1089292055

r-czajkowski (approved):

Issue #373, opened 2023-01-27
(MEDIUM) Change of the labels on the `Initiate minting tBTC` pop-up and in `Step 4`


michalinacienciala:
1. What we call in dApp the `Bitcoin Miner Fee` isn't actually the Bitcoin Miner Fee, it's a fee for optimistic minting. Could we rename it to `Minting Fee`?

2. We present the `Minted Amount` on `Initiate minting tBTC` pop-up  and in `Step 4`. When we show that, no tBTC is minted yet, so I would prefer to change the labels to `Amount To Be Minted` (or something other not suggesting that the minting actually took place)...

@sashatanase, what do you think?

![image](https://user-images.githubusercontent.com/78352137/215045869-73537108-d3d4-4721-bf4c-22217d4228e4.png)

![image](https://user-images.githubusercontent.com/78352137/215045901-eb567d85-6d1c-4d79-a4c5-3c35b3155751.png)

r-czajkowski:
Should't we display fees in `tBTC`? @pdyraga The system doesn't send any BTC to Bitcoin treasury address but sends fees in tBTC to treasury address.

Issue #371, opened 2023-01-26
(MEDIUM) Displayed precision (no. of decimal places) is different  for fees and for amount to be minted


michalinacienciala:
When the `Initiate minting tBTC` pop-up is displayed to the user in Step 3, it shows the values of the tBTC that will be minted and the fees that will be subtracted. The fees are displayed with the precision of 6 decimal places, but it looks to me that the minted amount is displayed with the precision of two decimal places (and rounded down). So despite I sent 0.016 BTC to the deposit address, I can see only 0.01 as a minted amount. As a user, knowing that the fees are presented with the 6 digit precision, I would assume that the minted amount is also presented with the same precision and would be wondering where my ~0.006 BTC has gone.
My suggestion would be to show the same 6 digit precision for the fees and for the minted amount. For the value in the title I think we could leave the two decimal points.

Wondering what's @sashatanase's opinion here.
Also, how should we round up the value in case of `0.099999` minted amount - to `0.10` or `0.09`? Mathematical standard is to round fractions to the nearest number - which would be `0.10`. It looks to me that we always round the deposited amount to the lower value. Maybe there's some preference here from the point of UX?

![image](https://user-images.githubusercontent.com/78352137/214949327-bd6f0b5c-d19b-4350-ae8d-dca21874a997.png)

pdyraga:
@michalinacienciala What do you think about promoting this issue to MEDIUM level? People can be confused/panic when they see fees are not adding up due to rounding in the dApp.

michalinacienciala:
Agree, I was battling with myself when deciding between MINOR/MEDIUM :)

michalinacienciala:
Generally what we want to achieve here is to not confuse a user about the displayed values.

Current design may be confusing because:
- user sees some of the values rounded to two decimal places, some to six
- the values seems to be always rounded down (i.e. `0.019` becomes `0.01`, `0.00000855` becomes `0.00008`), which is not the usual standard for rounding (we should round to the nearest value, i.e. (`0.019` becomes `0.02`, `0.00000855` becomes `0.00009`)

There are couple of ways we can tackle the problem here.
a) we can show the values with full precision (no confusion, but also hard to read)
b) we can show the `Minted Amount`, `Bitcoin Miner Fee` and `Threshold Network Fee` with 6 digit precision, and leave the amount in the title with 2 digit precision; in all those places use the standard rounding rules
[...]

Issue #364, opened 2023-01-26
(MAJOR) Balance Displaying Wrong in Final Step of tBTC Minting (Step 3)


liz-shinn:
**Final Minting tBTC Balance Was Wrong**
@r-czajkowski @michalsmiarowski 

⛔️ It all appeared to work fine until the very end. The displayed balance of tBTC I am to receive displays wrong in the center card. I sent 0.02 BTC and in the modals it said I would receive 0.02 tBTC. Here it says "You should receive **1.2 tBTC** in around 1-3 hours":

<img width="548" alt="Screenshot 2023-01-26 at 5 36 28 PM" src="https://user-images.githubusercontent.com/57226633/214896935-0ad93738-ddac-4b2e-b842-2dfa044fabbd.png">

⛔️ It also appears wrong here in the top lefthand corner, even though balance appears correctly in the History table though:

<img width="311" alt="Screenshot 2023-01-26 at 5 55 34 PM" src="https://user-images.githubusercontent.com/57226633/214900032-4021723c-ea6f-456c-836a-42783b15dcbc.png">

Tracked in QA Doc in Coda here: https://coda.io/d/Building-Keep_d-fmEgBNFVH/Liz-2023-01-26-QA-Template-tBTC-v2_suxe2#_lu121

PR #378, opened 2023-01-27
tBTCv2- token amount format improvements

r-czajkowski:
Closes: #373
Closes: #371
Closes: #364 

This PR improves the token amount format- here we implement the solution which behaves like this:
`0.001` -> `<0.01`
`0.016` -> `~0.02`
and amounts with higher precision are displayed after hovering on the rounded values. Of course we can control the precision. So eg. for precision `4`: `0.001` -> `0.001`; `0.00009` -> `<0.0001` etc. See unit tests for more details.

Also, it updates the `getEstimatedFees` method in TBTC wrapper- return the amount of tBTC token that will be minted. Based on the smart contract code the amount to mint is `<btcDepositAmount> - <treasuryFee> - <optimisticMintFee>`. Return all values in ERC20 standart- 18 decimals.

### DIFF
--- a/src/pages/tBTC/Bridge/TransactionHistory/index.tsx
+++ b/src/pages/tBTC/Bridge/TransactionHistory/index.tsx
@@ -19,7 +19,7 @@ import emptyHistoryImageSrcDark from "../../../../static/images/tBTC-bridge-no-h
 import emptyHistoryImageSrcLight from "../../../../static/images/tBTC-bridge-no-history-light.svg"
 import ViewInBlockExplorer from "../../../../components/ViewInBlockExplorer"
 import { ExplorerDataType } from "../../../../utils/createEtherscanLink"
-import { formatSatoshi } from "../../../../utils/formatAmount"
+import { InlineTokenBalance } from "../../../../components/TokenBalance"
 
 const bridgeTxHistoryStatusToBadgeProps: Record<
   BridgeHistoryStatus,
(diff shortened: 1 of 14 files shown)


src/threshold-ts/tbtc/index.ts
```diff
@@ -115,13 +116,16 @@ export interface ITBTC {
...
+   * Gets estimated fees that will be payed during a reveal and estimated amount
+   * of tBTC token that will be minted.
+   * @param depositAmount Amount that will be revealed in Satoshi.
+   * @returns Treasury fee, optitimistic mint fee and estimated amount of tBTC
+   * token that will be minted in ERC20 standart.
    */
   getEstimatedFees(depositAmount: string): Promise<{
     treasuryFee: string
     optimisticMintFee: string
+    amountToMint: string
```
michalsmiarowski, line 128: Non-blocking:

Since we are returning fees **and**, additionally, amount to mint then I think we should change the function name, because it's not longer returns only fees.

Maybe `getMintingPaymentsData`? I can't think of a better name right now so I'm open to discussion 🤔

src/pages/tBTC/Bridge/TransactionHistory/index.tsx
```diff
@@ -98,7 +98,12 @@ export const TransactionHistoryTable: FC<{
             data.map((_) => (
               <Tr key={_.txHash}>
                 <Td py={4} px={2}>
-                  {formatSatoshi(_.amount)}
+                  <InlineTokenBalance
+                    tokenAmount={_.amount}
+                    // TODO: Remove once we update the `bridgeTxHistory` in
+                    // `TBTC` wrapper in threshold lib.
```
r-czajkowski, line 104: I'll address in a separate PR. We need to take into account fees while fetching bridge tx history.

michalsmiarowski (approved):
LGTM 🔥

src/pages/tBTC/Bridge/TransactionHistory/index.tsx
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/408

Issue #368, opened 2023-01-26
(COSMETIC) The `x` character in the generated deposit address displayed incorrectly (as 'times'/'multiply' char)


michalinacienciala:
I generated new deposit address and noticed that one of the characters (`x`) is displayed in a weird way, resembling symbol of multiplication.
The other `x` character in the address was displayed normally.
When copied to clipboard and pasted to a text editor, the address looks good:`tb1qtx0ytj3sv0v4runu67ndgz45m7m8xarsan6qd3x25qvlv7r4qsrqft7mdq`.

![image](https://user-images.githubusercontent.com/78352137/214901454-90b1d639-7da4-433a-8632-157531572531.png)

michalsmiarowski:
Update on the bug:

I think this happens when we have two numbers between the `x` characters. If there is any letter next to `x` it displays normally.

#383 should fix that but this info may be helpful to test it out.

Issue #365, opened 2023-01-26
(COSMETIC) Unify capitalization of `address/Address` labels in the minting flow


michalinacienciala:
On the `Step 1` of the minting flow one input field label uses capitalized Address, the other label uses non-capitalized address. We should unify this.
![image](https://user-images.githubusercontent.com/78352137/214897651-c563dcb7-54d5-4114-a1db-a368a66ec173.png)

michalinacienciala:
![image](https://user-images.githubusercontent.com/78352137/214897566-abe2d9e6-4533-4ef5-9611-ddf710f7190d.png)

PR #383, opened 2023-01-27
Cosmetic improvements of the minting flow

michalsmiarowski:
Closes: #365 
Closes: #368

# Changes:

### 1. Unify capitalization of address/Address labels in the provide data form. 
We will use `Address` with the capital `A` for eth address and btc recovery address.

### 2. Fix the `x` character

The `x` character was replaced with the multipliction symbol because this is how `Inter` font works. We can disable that with css property:
`font-variant-ligatures: no-contextual;`

Additional helpful links:
https://github.com/react-native-elements/react-native-elements/issues/2546
https://github.com/rsms/inter/issues/222

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -8,6 +8,10 @@
       #root > div {
         min-height: 100vh;
       }
+      body,
+      html {
+        font-variant-ligatures: no-contextual;
+      }
     </style>
     <meta charset="utf-8" />
     <link rel="icon" href="%PUBLIC_URL%/favicon-T.png" />
--- a/public/index.html
+++ b/public/index.html
@@ -9,7 +9,7 @@
         min-height: 100vh;
       }
       body,
-      html {
+      * {
         font-variant-ligatures: no-contextual;
       }
     </style>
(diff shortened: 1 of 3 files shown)


src/pages/tBTC/Bridge/UnmintingCard/ProvideData.tsx
```diff
@@ -32,7 +32,7 @@ const MintingProcessFormBase: FC<ComponentProps & FormikProps<FormValues>> = ({
     <Form id={formId} mb={6}>
       <FormikInput
         name="ethAddress"
-        label="ETH address"
+        label="ETH Address"
```
r-czajkowski, line 35: We should update in `MintingCard/ProvideData.tsx`.

public/index.html
```diff
@@ -8,6 +8,10 @@
       #root > div {
         min-height: 100vh;
       }
+      body,
+      html {
+        font-variant-ligatures: no-contextual;
+      }
```
r-czajkowski, line 14: I still have small `x` in address in some places:

obraz

obraz

But looks ok in bride tx history table

obraz

r-czajkowski (changes requested):

src/pages/tBTC/Bridge/UnmintingCard/ProvideData.tsx
michalsmiarowski: Fixed in 63bead5b250d14852adee433a48e27576e7262da

Looks like the whole `Unminting Card` needs to be refactored (cause there are so much code that repeats) so I don't think it's necessary to remove this change from `UnmintingCard` anyway.

r-czajkowski (approved):

Issue #379, opened 2023-01-27
Adding BTC Recovery Address to a JSON output file after step 1 in the deposit flow


dimpar:
It would improve the recovery process of BTC if we also have a BTC recovery address inside this JSON file.

PR #386, opened 2023-01-27
Integrate second track deposit verification with Sentry telemetry

lukasz-zimnoch:
Here we integrate the second track deposit address verification with deposit telemetry. The verification calls a Cloud Function deployed on `keep-test` and is done right before sending deposit data to Sentry. The verification outcome is attached to the telemetry data as two additional fields:
- `verificationStatus` which can have three different values
  - `valid`: verification service responded with the same address as the generated one
  - `invalid`: verification service responded with a different address
  - `error`: request or verification service error
- `verificationResponse` which holds the complete request outcome

The verification service availability does not impact the dApp functionality. The deposit flow and telemetry works even if something is wrong on service side.

### DIFF
--- a/src/hooks/tbtc/useDepositTelemetry.ts
+++ b/src/hooks/tbtc/useDepositTelemetry.ts
@@ -15,11 +15,17 @@ export const useDepositTelemetry = (network: BitcoinNetwork) => {
         network
       )
 
-      captureMessage(`Generated deposit [${depositAddress}]`, {
-        ...deposit,
-        verificationStatus: status,
-        verificationResponse: response,
-      })
+      captureMessage(
+        `Generated deposit [${depositAddress}]`,
+        {
+          ...deposit,
+          verificationStatus: status,
+          verificationResponse: response,
+        },
+        {
+          "verification.status": status,
+        }
+      )
     },
     [verifyDepositAddress, network, captureMessage]
   )
--- a/src/hooks/tbtc/useDepositTelemetry.ts
+++ b/src/hooks/tbtc/useDepositTelemetry.ts
@@ -1,14 +1,26 @@
 import { useCallback } from "react"
 import { useCaptureMessage } from "../sentry"
 import { DepositScriptParameters } from "@keep-network/tbtc-v2.ts/dist/src/deposit"
+import { verifyDepositAddress } from "../../utils/verifyDepositAddress"
+import { BitcoinNetwork } from "../../threshold-ts/types"
 
-export const useDepositTelemetry = () => {
+export const useDepositTelemetry = (network: BitcoinNetwork) => {
   const captureMessage = useCaptureMessage()
 
   return useCallback(
-    (deposit: DepositScriptParameters, depositAddress: string) => {
-      captureMessage(`Generated deposit [${depositAddress}]`, deposit)
+    async (deposit: DepositScriptParameters, depositAddress: string) => {
+      const { status, response } = await verifyDepositAddress(
+        deposit,
+        depositAddress,
+        network
+      )
+
+      captureMessage(`Generated deposit [${depositAddress}]`, {
+        ...deposit,
+        verificationStatus: status,
+        verificationResponse: response,
+      })
     },
-    []
+    [verifyDepositAddress, network, captureMessage]
   )
 }
--- a/src/utils/verifyDepositAddress.ts
+++ b/src/utils/verifyDepositAddress.ts
@@ -4,7 +4,7 @@ import { BitcoinNetwork } from "../threshold-ts/types"
 
 export interface VerificationOutcome {
   status: "valid" | "invalid" | "error"
-  response: any
+  response: unknown
 }
 
 export const verifyDepositAddress = async (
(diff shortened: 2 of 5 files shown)


src/utils/verifyDepositAddress.ts
```diff
@@ -0,0 +1,38 @@
+import axios from "axios"
+import { DepositScriptParameters } from "@keep-network/tbtc-v2.ts/dist/src/deposit"
+import { BitcoinNetwork } from "../threshold-ts/types"
+
+export interface VerificationOutcome {
+  status: "valid" | "invalid" | "error"
+  response: any
```
mhluongo, line 7: Let's use `unknown` here. `any` is an anti-pattern

mhluongo: Or if we're confident `response.address` will be a non-empty hex string or `response` will be an `Error`, we could just type it that way

src/utils/verifyDepositAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const verifyDepositAddress = async (
+  deposit: DepositScriptParameters,
+  depositAddress: string,
```
mhluongo, line 12: Do we have a hex string type in this codebase? Not a blocker or anything, but it'd be nice

mhluongo (commented):
Couple quick thoughts

src/utils/verifyDepositAddress.ts
lukasz-zimnoch: That's not a hex. It's a base58-encoded Bitcoin address.

src/utils/verifyDepositAddress.ts
lukasz-zimnoch: The verification endpoint returns more things than just the address. We attach it here to facilitate debugging in case the verification resolves to `invalid`. That said, we either have a JSON or error here so `unknown` seems to be more flexible. See: f41ae41e

src/hooks/tbtc/useDepositTelemetry.ts
```diff
@@ -1,14 +1,26 @@
...
+        deposit,
+        depositAddress,
+        network
+      )
+
+      captureMessage(`Generated deposit [${depositAddress}]`, {
+        ...deposit,
+        verificationStatus: status,
+        verificationResponse: response,
+      })
```
michalsmiarowski: Some time ago Dymitr proposed to include `BTC recovery address` (alongside with `refundPublicKeyHash`) in the JSON file as it would improve the recovery process (see #379).

Maybe we could also include it here when we capture the message so that it displays in the Sentry task?

src/utils/verifyDepositAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+      `${endpoint}/json/${network}/latest/${depositor.identifierHex}/${blindingFactor}/${refundPublicKeyHash}/${refundLocktime}`,
+      { timeout: 10000 } // 10s
+    )
+
+    const match = response.data.address === depositAddress
+
+    return {
+      status: match ? "valid" : "invalid",
+      response: response.data,
+    }
```
michalsmiarowski, line 31: Hmmm, something is wrong either with the endpoint or the code in the dApp. Two different walletPubKeyHashes are used and the status is `invalid` when I test it locally.

I consoled log the `deposit` argument of `verifyDepositAddress` and `response.data`:

`deposit`:
```
{
    "depositor": {
        "identifierHex": "f119557ac33585405467135ec9a343dcdb047517"
    },
    "blindingFactor": "823cb96f4ce67f38",
    "walletPublicKeyHash": "56988a974575d42db330193acd7a8d9efc67f830",
    "refundPublicKeyHash": "7ac2d9378a1c47e589dfb8095ca95ed2140d2726",
    "refundLocktime": "e48c3b65"
}
```

`response.data`:
```
{
    "network": "testnet",
    "block": "latest",
    "sender": "f119557ac33585405467135ec9a343dcdb047517",
    "blindingFactor": "823cb96f4ce67f38",
    "walletPubKeyHash": "0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e",
    "refundPubKeyHash": "7ac2d9378a1c47e589dfb8095ca95ed2140d2726",
    "lockTime": "e48c3b65",
    "address": "tb1qc7wnr704wyrzn4a30g9wnzduaf8x2w4dkl8nnphl5zwd07epj2aq57fp49"
}
```

As you can see two different wallet public key hashes are used. Maybe we do it wrongly in the dApp? We are getting the wallet public key here:
https://github.com/threshold-network/token-dashboard/blob/e888db948a1a47107010b0b6a1399c585c0488ca/src/threshold-ts/tbtc/index.ts#L262-L264

and then convert it to public key hash here:
https://github.com/threshold-network/token-dashboard/blob/e888db948a1a47107010b0b6a1399c585c0488ca/src/threshold-ts/tbtc/index.ts#L287

michalsmiarowski: Neverming. It happened because i've used `dapp-development-goerli` locally and we have `goerli` contracts addresses in the `https://us-central1-keep-test-f3e0.cloudfunctions.net/verify-deposit-address` endpoint. I've tested it out with `goerli` contracts and it works ok.

Gonna keep this comment just in case as maybe somene find this helpful.

src/hooks/tbtc/useDepositTelemetry.ts
lukasz-zimnoch: The `refundPublicKeyHash` is included. We can easily derive the address from it if needed.

michalsmiarowski (approved):
Looking good to me ✌️ 

Teted it out locally and it works wonderful!

Issue #403, opened 2023-01-28
(MAJOR) Don't allow the user to use eth address other than his own


michalsmiarowski:
The dApp allows user to create a deposit address using address other than his own. We shouldn't allow this because after user sends his funds to the deposit address he won't be able to reveal the deposit. It will throw an error `Wrong 32-byte script hash`.

We should add some kind of validation to the form that checks if given eth address is equal to the one that is connected.

pdyraga:
![image](https://user-images.githubusercontent.com/4712360/215289281-b9b9a3ac-49df-460f-813f-33c667f0f1eb.png)

sashatanase:
We should make the ETH Address Read-only, as @pdyraga suggested.

PR #404, opened 2023-01-28
Make ETH address field readonly

### LINKED ISSUE #407: (MEDIUM) The `ETH Address` does not get updated when user switches between accounts
When user switches between accounts, the `ETH Address` does not get updated:

Let's consider two accounts that I have in my MetaMask:

* `Acc1: 0xC31D206f48EC365bc786563dE7cFb67fDb7e74e4`
* `Acc2: 0x58C6A45AcFCc1fD0E5A103Cab2caE00b0B188EC5`

Steps:

1. I open a fresh tab in the browser and go to https://dashboard.test.threshold.network/tBTC
2. I make sure that in my MM plugin I have `Acc1` selected
3. I click `Connect Wallet` -> `MetaMask` -> `View Dashboard`. I can see correct `ETH address`
image
4. I open MM plugin, click on `Connected` next to account name and then I choose `Switch to this account` under `Acc2`. I can still see ETH address of `Acc1`:
image.
5. If I progress with further steps of the minting, I can still see the wrong ETH account.

michalsmiarowski:
Closes: #403 
Closes: #407 

The dApp allows user to create a deposit address using address other than his own. We shouldn't allow this because after user sends his funds to the deposit address he won't be able to reveal the deposit. It will throw an error `Wrong 32-byte script hash`.

We are adding two things:
- making eth address readonly,
- add a `throw error` in onSubmit function just in case someone manage to submit the form with a different address.

**Additional notes:**
We also add `enableReinitialize` option to the formik form so that it is reinitialized with a new initial value (which is eth address) when the address change. (this closes #407)

### DIFF
--- a/src/pages/tBTC/Bridge/MintingCard/ProvideData.tsx
+++ b/src/pages/tBTC/Bridge/MintingCard/ProvideData.tsx
@@ -19,6 +19,7 @@ import { BitcoinNetwork } from "../../../../threshold-ts/types"
 import { useTBTCDepositDataFromLocalStorage } from "../../../../hooks/tbtc"
 import withOnlyConnectedWallet from "../../../../components/withOnlyConnectedWallet"
 import { useDepositTelemetry } from "../../../../hooks/tbtc/useDepositTelemetry"
+import { isSameETHAddress } from "../../../../web3/utils"
 
 export interface FormValues {
   ethAddress: string
@@ -40,6 +41,7 @@ const MintingProcessFormBase: FC<ComponentProps & FormikProps<FormValues>> = ({
         label="ETH address"
         tooltip="ETH address is prepopulated with your wallet address. This is the address where you’ll receive your tBTC."
         mb={6}
+        isReadOnly={true}
       />
       <FormikInput
         name="btcRecoveryAddress"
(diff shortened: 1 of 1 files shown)


src/pages/tBTC/Bridge/MintingCard/ProvideData.tsx
```diff
@@ -40,6 +41,7 @@ const MintingProcessFormBase: FC<ComponentProps & FormikProps<FormValues>> = ({
         label="ETH address"
         tooltip="ETH address is prepopulated with your wallet address. This is the address where you’ll receive your tBTC."
         mb={6}
+        isReadOnly={true}
```
r-czajkowski, line 44: Non-blocking

Or maybe `isDisbaled`? In my opinion, looks like a bug now because I can't select and change the field. Note if we agree on `isDisabled` we need to update styles because the tooltip for this field will be unreadable.

r-czajkowski (approved):

src/pages/tBTC/Bridge/MintingCard/ProvideData.tsx
r-czajkowski: Or just change the cursor to `not-allowed` on hover to indicate this field is read-only.

Issue #390, opened 2023-01-28
(MAJOR) No deposits under 0.01 BTC are able to be minted


mhluongo:
The dust threshold enforced by the minting contracts isn't displayed in the app, allowing users to mistakenly send ~$230 at today's prices, leaving it stuck for up to 9 months.

We need to
* Inform users of the dust limit at deposit time
* Give them a human-readable error if they sent too little BTC, and let them know about the refund instructions.

Below is when I realized I'd sent too little BTC.

![Screenshot from 2023-01-27 22-20-50](https://user-images.githubusercontent.com/427505/215240114-53e2f007-f7c5-4cca-8698-5ddcd3727a0a.png)

pdyraga:
What about this @mhluongo @sashatanase, is this enough? 

![image](https://user-images.githubusercontent.com/4712360/215274614-69f23e88-9107-456d-b17d-7cd9b31b72e8.png)

sashatanase:
So for this, we have it handled in the design here https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=7760%3A199526&t=KYjqvDq92FqwH61h-1 and here https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=7793%3A204627&t=KYjqvDq92FqwH61h-1

Meaning we have this information in the How it works page and in Step 2 

![image](https://user-images.githubusercontent.com/40768736/215288812-6636c22f-d7d7-42b7-b1d7-1140ee86cda9.png)
![image](https://user-images.githubusercontent.com/40768736/215288832-c78fd798-76be-4d90-85a4-e997eaa2e83b.png)

michalinacienciala:
@r-czajkowski, I checked this and I still can't see the new copy. Are you sure it got fixed in #406?

michalinacienciala:
Sorry for confusion, I made a mistake. The changes are there.

PR #406, opened 2023-01-28
Add info about minimum deposit value

michalsmiarowski:
Closes: #390 

The dust threshold enforced by the minting contracts isn't displayed in the app, allowing users to mistakenly send ~$230 at today's prices, leaving it stuck for up to 9 months.

We are adding info about minimum deposit value (0.01 BTC) in step 2 of the minting flow and in minting timeline.

I've also added a non-breaking space (`&nbsp;`) between the value and `BTC` (`0.01&nbsp;BTC`) so that those two words alway stay in the same line.

Todo: 
- [ ] Make the error when doing a reveal more readable if too little BTC were sent. Also let the user know about the refund instructions.

### DIFF
--- a/src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
+++ b/src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
@@ -53,8 +53,8 @@ const MakeDepositComponent: FC<{
         subTitle="Make your BTC deposit"
       />
       <BodyMd color="gray.500" mb={6}>
-        Use this generated address to send any amount of BTC you want to mint as
-        tBTC.
+        Use this generated address to send any amount larger than 0.01&nbsp;BTC,
+        to mint as tBTC.
       </BodyMd>
       <BodyMd color="gray.500" mb={6}>
         This address is an unique generated address based on the data you
--- a/src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
+++ b/src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
@@ -53,8 +53,8 @@ const MakeDepositComponent: FC<{
         subTitle="Make your BTC deposit"
       />
       <BodyMd color="gray.500" mb={6}>
-        Use this generated address to send any amount larger than 0.01&nbsp;BTC,
-        to mint as tBTC.
+        Use this generated address to send minimum 0.01&nbsp;BTC, to mint as
+        tBTC.
       </BodyMd>
       <BodyMd color="gray.500" mb={6}>
         This address is an unique generated address based on the data you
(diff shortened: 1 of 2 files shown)


src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
```diff
@@ -53,8 +53,8 @@ const MakeDepositComponent: FC<{
         subTitle="Make your BTC deposit"
       />
       <BodyMd color="gray.500" mb={6}>
-        Use this generated address to send any amount of BTC you want to mint as
-        tBTC.
+        Use this generated address to send any amount larger than 0.01&nbsp;BTC,
```
r-czajkowski, line 56: The amount may be larger than or equal to deposit dust threshold.

https://github.com/keep-network/tbtc-v2/blob/a511c31bcd44c44d4fc28dbe07f98b1648954da0/solidity/contracts/bridge/Deposit.sol#L248-L251

src/pages/tBTC/Bridge/MintingCard/MintingTimeline.tsx
```diff
@@ -55,7 +55,7 @@ export const MintingTimelineStep2: FC<MintingTimelineStepProps> = ({
       stepText="Step 2"
       helperLabelText="ACTION ON BITCOIN NETWORK"
       title="Make a BTC deposit"
-      description="Send any amount of BTC to this unique BTC Deposit Address. The amount you send is the amount will be minted as tBTC."
+      description="Send any amount larger than 0.01&nbsp;BTC to this unique BTC Deposit Address. The amount you send will be minted as tBTC."
```
r-czajkowski, line 58: Same as above.

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
r-czajkowski, line 56: The `deposit dust threshold` is a governable param so we should fetch this value from the contract. Ofc for release we are good to go here.

michalsmiarowski: c6794cd989efa030d3f9fb2185d04737a660e388

src/pages/tBTC/Bridge/MintingCard/MintingTimeline.tsx
michalsmiarowski: c6794cd989efa030d3f9fb2185d04737a660e388

src/pages/tBTC/Bridge/MintingCard/MakeDeposit.tsx
michalsmiarowski: Let's address it in a separate PR after the release

r-czajkowski (approved):

Issue #388, opened 2023-01-27
(MAJOR) Balance shown in TH HISTORY shows value of deposited BTC, not minted tBTC


michalinacienciala:
I deposited `0,01567891 BTC` and requested minting.
The success page shows:

```
You should receive:    ~0.02 (0.01564)
Amount To Be Minted:   ~0.02 (0.01564)
Minting Fee:           ~0.000031 (0.00003134)
Threshold Network Fee: ~0.000008 (0.00000783)
Balance in Tx History: ~0.02 (0.15679)
tBTC Balance:          0 (0)
```
The values are not in line (~~the balance shouldn't be `0` and~~ we should see the same value for the amount to be minted and the one seen in tx history).

![image](https://user-images.githubusercontent.com/78352137/215221305-2cdd70f2-c076-45fa-a1d2-841af5ee8487.png)

ETH wallet: 0xC31D206f48EC365bc786563dE7cFb67fDb7e74e4
Deposit address: tb1qtlqvxg64tt25h77q7tvgg20mvcpqsfytyug7ytpamclngrj6jnjq7398tn
Reveal tx: https://goerli.etherscan.io/tx/0x39cd742c001376846b2d13346079fbab69a150acb2dcfc2ce13d2d9d6699229a#eventlog

michalinacienciala:
After transaction got minted, the balances got updated accordingly:
```
You should receive:    ~0.02 (0.01564)
Amount To Be Minted:   ~0.02 (0.01564)
Minting Fee:           ~0.000031 (0.00003134)
Threshold Network Fee: ~0.000008 (0.00000783)
Balance in Tx History: ~0.02 (0.15679)
tBTC Balance:          ~0.02 (0.01564)
```

Minting tx: https://goerli.etherscan.io/tx/0x52838e710adcc321fc510bb93cdb3226930a57efcc94c21d2305d839a3677326

michalinacienciala:
**Looks that the balance in the `TX HISTORY` shows value of deposited BTC, not TBTC minted for depositor.**

PR #408, opened 2023-01-29
Update `bridgeTxHistory` method in TBTC wrapper

r-czajkowski:
Closes: #388

Take into account fees while calculating an amount. If it's a revealed deposit we need to get `treasuryFee` param from onchain `DepositRequest` struct because the `treausuryFee` depends on the deposited amount and `depositTreasuryFeeDivisor` param which is governable so it means it may change over time.

If it's an already minted deposit the only way to get the exact amount of minted `tBTC` is:
1. Get all logs from transaction in which the optimistic minting was completed.
2. Find `Transfer` events from tBTC token contract based on logs.
3. Find `Transfer` event where `to` param is equal to the depositor address. There is only one transfer to the depositor account.

### DIFF
--- a/src/hooks/tbtc/useSubsribeToDepositRevealedEvent.ts
+++ b/src/hooks/tbtc/useSubsribeToDepositRevealedEvent.ts
@@ -31,11 +31,15 @@ export const useSubscribeToDepositRevealedEvent = () => {
     ) => {
       if (!account || !isSameETHAddress(depositor, account)) return
 
+      const { amountToMint } = await threshold.tbtc.getEstimatedFees(
+        amount.toString()
+      )
+
       dispatch(
         tbtcSlice.actions.depositRevealed({
           fundingOutputIndex,
           fundingTxHash,
-          amount: amount.toString(),
+          amount: amountToMint,
           txHash: event.transactionHash,
           depositor: depositor,
           depositKey: threshold.tbtc.buildDepositKey(
--- a/src/threshold-ts/tbtc/index.ts
+++ b/src/threshold-ts/tbtc/index.ts
@@ -426,11 +426,7 @@ export class TBTC implements ITBTC {
     )
 
     const estimatedAmountToMintByDepositKey =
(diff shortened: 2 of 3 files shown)


src/threshold-ts/tbtc/index.ts
```diff
@@ -384,33 +420,55 @@ export class TBTC implements ITBTC {
...
+      depositor,
+      depositKeys
+    )
+
+    const estimatedAmountToMintByDepositKey =
+      await this._calculateEstimatedAmountToMintForRevealedDeposits(
+        mintedDepositEvents.map((event) =>
+          (event.args?.depositKey as BigNumber).toHexString()
+        )
+      )
```
michalsmiarowski, line 433: We should pass `depositKeys` array that we got from `revealedDeposits` not `mintedDepositEvents. We need `estimatedAmountToMintByDepositKey` to calculate the amount for all revealed deposits (that can be also in pending state).

michalsmiarowski (changes requested):
Left some comments to look at before the merge:

src/hooks/tbtc/useSubsribeToDepositRevealedEvent.ts
```diff
@@ -31,11 +31,15 @@ export const useSubscribeToDepositRevealedEvent = () => {
...
+      const { amountToMint } = await threshold.tbtc.getEstimatedFees(
+        amount.toString()
+      )
+
       dispatch(
         tbtcSlice.actions.depositRevealed({
           fundingOutputIndex,
           fundingTxHash,
-          amount: amount.toString(),
+          amount: amountToMint,
```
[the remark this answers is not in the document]
michalsmiarowski: What about `amountWithoutFees`? 😃 

If not then maybe we could at least add a coment about it in the `BridgeTxHistory` interface?

src/threshold-ts/tbtc/index.ts
r-czajkowski: 50a511f

michalsmiarowski (approved):
LGTM

PR #409, opened 2023-01-29
Fix deposit flow

r-czajkowski:
~Depends on: #392~

Previous implementation didn't work well due to unnecessary state updates caused by `setInterval` inside the `useEffect` hook. Here we move the redirection logic to the correct step of the deposit flow to the listener middleware for better handling async work.

### DIFF
--- a/src/store/tbtc/effects.ts
+++ b/src/store/tbtc/effects.ts
@@ -4,6 +4,11 @@ import { isAddress, isAddressZero } from "../../web3/utils"
 import { MintingStep } from "../../types/tbtc"
 import { ONE_SEC_IN_MILISECONDS } from "../../utils/date"
 import { TaskAbortError } from "@reduxjs/toolkit"
+import {
+  TBTCLocalStorageDepositData,
+  key,
+  removeDataForAccount,
+} from "../../utils/tbtcLocalStorageData"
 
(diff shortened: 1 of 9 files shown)


src/store/tbtc/effects.ts
```diff
@@ -30,3 +38,104 @@ export const fetchBridgeTxHitoryEffect = async (
...
+        const isDepositRevealed = deposit.revealedAt !== 0
+
+        if (isDepositRevealed) {
+          // Deposit already revealed, force start from step 1 and remove deposit data.
+          removeDataForAccount(
+            depositor,
+            JSON.parse(
+              localStorage.getItem(key) || "{}"
+            ) as TBTCLocalStorageDepositData
+          )
```
michalsmiarowski, line 101: Non-blocking:

I think we could also remove the deposit data after user reveals the deposit (but keep the data in the store). This way we will only remove the data from the store if he click `Dismiss` or back button in step 4, and if he refresh (while being on step 4) we won't have to wait for the utxo to be found and the check if it's revealed. We will navigate him to step 1 (provide data) immedietaly.

michalsmiarowski (approved):
Tested it out locally and it looks good 🧞‍♂️ Approved!

PR #391, opened 2023-01-28
Prefer tBTC over TBTC in all user-facing copy

mhluongo:
The codebase still mixes `tBTC`, `TBTC`, and `Tbtc` in variables names. I'm considering adding an explicit linter rule for those, but will save it for later 🤪

PR #410, opened 2023-01-30
Prefer tBTC over TBTC in all user-facing copy

r-czajkowski:
Updates made by @mhluongo  🙏.

Ref: #391
The codebase still mixes tBTC, TBTC, and Tbtc in variables names. I'm considering adding an explicit linter rule for those, but will save it for later 🤪.

Force the first letter(tBTC) to be lowercase. The `LabelSm`
component sets the text to uppercase by default.


michalsmiarowski (approved):
LGTM 🔥

Issue #393, opened 2023-01-28
Check the validation logic for non-witness multisig address


pdyraga:
From @lukasz-zimnoch:

A witness multisig address will be rejected by the validation as its hash has 32 bytes. However non-witness multisig address is a 20 byte hash underneath, just like single sig P2PKH and P2WPKH. That means it may pass the static validation but the refund will not be possible as the Bitcoin script processor will use PKH rules to validate it. We need to check this path and if that turns out to be possible, improve the static validation.

PR #412, opened 2023-01-30
Enforce BTC recovery address to be always PKH-based

lukasz-zimnoch:
Closes: #393

The underlying Bitcoin deposit script that is used to lock the deposit UTXO supports only PKH-based addresses as recovery. Here we enforce that requirement on the dApp level.


michalsmiarowski (approved):
LGTM 🔥

r-czajkowski (approved):

PR #440, opened 2023-02-08
Fix fetching legacy Keep stake

r-czajkowski:
The `Multicall` service calls `decodeFunctionResult` under the hood which returns the result as a `Result` type from `ethers` lib. A Result is an array, so each value can be accessed as a positional argument. Additionally, if values are named, the identical object as its positional value can be accessed by its name. So if we call `eligibleStake` function using `Multicall` we can get the value as `result[0]` or `result.balance`. In the previous implementation, we passed the amount as `Result` type to the `convertToT` which caused an unexpected error. Also, we should pass amount as a string to the `convertToT` method from `VendingMachine` service.


michalsmiarowski (approved):
LGTM

PR #443, opened 2023-02-10
Fix fetching data with Coinbase Wallet

r-czajkowski:
We need to force the Coinbase Wallet provider to use our RPC url because we can't fetch past events from block range greater than `1000` when using the default RPC url provided by Coinbase Wallet extension. So the dapp is unable to fetch stakes for user. Here we force the provider to use our RPC url by triggering the `updateProviderInfo` function from `CoinbaseWalletProvider`. See
(https://github.com/coinbase/coinbase-wallet-sdk/blob/v3.0.4/src/provider/CoinbaseWalletProvider.ts#L209). The `updateProviderInfo` is a private method but as we know private and protected are only enforced during type checking. This means that JavaScript runtime constructs like `in` or simple property lookup can still access a `private` or `protected` member.


michalsmiarowski (approved):
LGTM 🔥 Testeid it out locally on mainnet contracts and on goerli contracts and looks like it's fixed.

Issue #429, opened 2023-02-02
Make it easier to track confirmations of a deposit


pdyraga:
Right now, the user who deposited Bitcoin does not see the number of Bitcoin confirmations on their deposit transaction and does not know how long it is going to take for the optimistic minting to be requested.

This requires design and implementation.

sashatanase:
Following our discussion and decision today, we will add the BTC Confirmations loader on the Success Page like here. 


![Image](https://user-images.githubusercontent.com/40768736/217020884-5b3e6646-5b26-4d43-8f96-17dc56004bbb.png)

BTC Confirmations loader

The Loader should be dynamic and fill each BTC confirmation. In the end, when it reaches 6/6, the checkmark will turn from grey to purple.

Behaviour Prototype Animation 

https://www.figma.com/proto/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?page-id=7126%3A198919&node-id=7919%3A208962&viewport=-1672%2C-20428%2C1.12&scaling=min-zoom&starting-point-node-id=7918%3A208894&show-proto-sidebar=1


![Image](https://user-images.githubusercontent.com/40768736/217024396-d1e519ef-f66c-4a7b-b647-32b4ff957d9d.png)



Figma Files

https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=7846%3A206304&t=DPkqsh0XagqqVP3k-1

sashatanase:
Please review the following copy. 

```
The tBTC minting process will start only after 6 Bitcoin network confirmations.
```

pdyraga:
Here are the rules:
- If the amount is less than 0.1 BTC, it should have at least 1 confirmation. 
- If the tx is less than 1 BTC, it should have at least 3 confirmations.
- Otherwise, we need to wait for 6 confirmations.

mhluongo:
Note that I want to start the minting request even earlier, but that won't change for at least one maintainer release. Until then, @pdyraga 's rules are correct.

https://github.com/keep-network/optimistic-minting/issues/65

mhluongo:
Closed by #444

PR #444, opened 2023-02-10
🔎 📄  Track transaction confirmations 

michalsmiarowski:
Closes: #429 

Adds the bitcoin confirmation loader to the Minting Success page.

image

# Main changes:
- add `getTransactionConfirmations` method to the TBTC class
- add `minimumNumberOfConfirmationsNeeded` method to the TBTC class
- add new effect - `fetchUtxoConfirmationsEffect`
- mock transaction confirmations in the `MockBitcoinClient`

# Details of the changes

### - add `getTransactionConfirmations` method to the TBTC class

Added a new method to the TBTC class that gets the number of confirmations that a given transaction has accumulated
so far. It take hash of the transaction as an argument. It calls the `getTransactionConfirmations` method defined inside the `bitcoinClient`.

### - add `minimumNumberOfConfirmationsNeeded` method to the TBTC class

Added a new method to the TBTC class that gets the minimum number of confirmations needed for the minter to start the
minting process. The minimum number of confirmations is based on the amount
that was sent to the deposit address:
The rules are:
- If the amount is less than 0.1 BTC, it should have at least 1 confirmation.
- If the tx is less than 1 BTC, it should have at least 3 confirmations.
- Otherwise, we need to wait for 6 confirmations.

### - add new effect - `fetchUtxoConfirmationsEffect`

Added a `txConfirmations` variable to the redux store. It will store the current number of transactions (as long as it's lower or equal than `minimumNumberOfConfirmationsNeeded`. It stops updating after it's above that value). The new effect - `fetchUtxoConfirmationsEffect` is called when the user lands on the minting success page and it fetches the actual number fo the confirmation for the given utxo.

The effect works similar to `findUtxoEffect`. It calls the `getTransactionConfirmations` method every 10 seconds.

### - mock transaction confirmations in the `MockBitcoinClient`

Created a new private method in `MockBitcoinClient` called `_mockConfirmationsForTransaction` which can mock the confirmations for the given transaction. We could also specify how many confirmations we want to mock (default is 10). Every confirmation is mocked every 8 seconds.


src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -19,18 +19,30 @@ import TransactionDetailsTable from "../components/TransactionDetailsTable"
...
   const onDismissButtonClick = () => {
     onPreviousStepClick(MintingStep.ProvideData)
   }
 
+  useEffect(() => {
+    // TODO: This useEffect is fired when we go from Success Modal to another
+    // subpage (for example `Staking`) and go back here. We probalby shouldn't
+    // fire it again if we have at least 6 confirmations in the state.
+    dispatch(tbtcSlice.actions.fetchUtxoConfirmations({ utxo }))
+  }, [dispatch, JSON.stringify(utxo)])
```
mhluongo, line 44: Why the `stringify` here? Not familiar with the codebase, just curious!

src/threshold-ts/tbtc/index.ts
```diff
@@ -414,6 +425,25 @@ export class TBTC implements ITBTC {
...
+  getTransactionConfirmations = async (
+    transactionHash: TransactionHash
+  ): Promise<number> => {
+    return await this._bitcoinClient.getTransactionConfirmations(
+      transactionHash
+    )
+  }
+
+  minimumNumberOfConfirmationsNeeded = (amount: BigNumberish): number => {
+    let minConfrimations = 6
```
r-czajkowski, line 443: Non-blocking: looks like we can return number directly because we're not doing anything else with the `minConfrimations` variable.
```
return 1
...
return 3
...
return 6
```

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -19,18 +23,47 @@ import TransactionDetailsTable from "../components/TransactionDetailsTable"
...
+  useEffect(() => {
+    dispatch(tbtcSlice.actions.fetchUtxoConfirmations({ utxo }))
+  }, [dispatch, JSON.stringify(utxo)])
+
+  useEffect(() => {
+    const minConfrimations = threshold.tbtc.minimumNumberOfConfirmationsNeeded(
+      utxo.value
+    )
+    setMinConfirmationsNeeded(minConfrimations)
+  }, [utxo.value])
```
r-czajkowski, line 58: Looks like the `utxo.value` is a `BigNumber`. Let's extract value from utxo `const value = utxo.value.toString()`and use it in the dependency array.

r-czajkowski (changes requested):

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
r-czajkowski, line 58: I wonder if we really need this useEffect. The `minimumNumberOfConfirmationsNeeded` is not an async method so we can just:
```ts
const minimumNumberOfConfirmationsNeeded = threshold.tbtc.minimumNumberOfConfirmationsNeeded(
      utxo.value
    )
```

If we want to memoized value we can just use `useMemo` hook but IMO it's unnecessary because the `minimumNumberOfConfirmationsNeeded` returns `number` and doesn't do any complicated operations.

r-czajkowski (changes requested):

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
michalsmiarowski: In `fetchUtxoConfirmations` we use the utxo data to fetch the confirmations so we want it to re-fetch if the utxo changes. Dependency array in `useEffect` uses referential equality to compare values. This means that if we just put object in a dependency array it will still unnecessary re-render.  To avoid that I've put `JSON.stringify` there to compare if the content of the object changes and only then dispatch the `fetchUtxoConfrimations`. 

I think Rafał proposed awesome solution though here: https://github.com/threshold-network/token-dashboard/pull/444#discussion_r1105772794. So I will refactor this. Thanks for pointing that out!

src/threshold-ts/tbtc/index.ts
```diff
@@ -148,6 +148,17 @@ export interface ITBTC {
...
+  /**
+   * Gets the number of confirmations that a given transaction has accumulated
+   * so far.
+   * @param transactionHash - Hash of the transaction.
+   * @returns The number of confirmations.
+   */
+  getTransactionConfirmations(transactionHash: TransactionHash): Promise<number>
+
+  // TODO: add comment
+  minimumNumberOfConfirmationsNeeded(amount: BigNumberish): number
```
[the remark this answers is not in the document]
michalsmiarowski: 6a9f7104b885509730dd463e64a88a48b70ed130

src/threshold-ts/tbtc/index.ts
```diff
@@ -414,6 +425,25 @@ export class TBTC implements ITBTC {
...
+  ): Promise<number> => {
+    return await this._bitcoinClient.getTransactionConfirmations(
+      transactionHash
+    )
+  }
+
+  minimumNumberOfConfirmationsNeeded = (amount: BigNumberish): number => {
+    let minConfrimations = 6
+    const amountInBN = BigNumber.from(amount)
+    if (amountInBN.lt(10000000)) {
```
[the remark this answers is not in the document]
michalsmiarowski: 6a9f7104b885509730dd463e64a88a48b70ed130

src/threshold-ts/tbtc/index.ts
```diff
@@ -414,6 +425,25 @@ export class TBTC implements ITBTC {
...
+      minConfrimations = 1
+    } else if (amountInBN.lt(100000000)) {
```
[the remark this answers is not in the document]
michalsmiarowski: 6a9f7104b885509730dd463e64a88a48b70ed130

src/threshold-ts/tbtc/index.ts
michalsmiarowski: bd4d0f0b12e806d93dc06da420c434315eb04198

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -19,18 +23,47 @@ import TransactionDetailsTable from "../components/TransactionDetailsTable"
...
 
   const tbtcTokenAddress = useTBTCTokenAddress()
 
   const onDismissButtonClick = () => {
     onPreviousStepClick(MintingStep.ProvideData)
   }
 
+  useEffect(() => {
+    dispatch(tbtcSlice.actions.fetchUtxoConfirmations({ utxo }))
+  }, [dispatch, JSON.stringify(utxo)])
```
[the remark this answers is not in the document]
michalsmiarowski: 7a54f8bb4a82a3e082a462f7465b71feb05d159e

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
michalsmiarowski: That's a good point! Fixed it in f56040d51186b00c97e6f5a3a13de8ed8aeef7d5

michalsmiarowski: See https://github.com/threshold-network/token-dashboard/pull/444#discussion_r1108513134

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -66,6 +130,10 @@ const MintingSuccessComponent: FC<{
             text="token address"
           />{" "}
           to your Ethererum wallet.
+          <Text>
+            The tBTC minting process will start only after 6 Bitcoin Network
+            confirmations.
+          </Text>
```
r-czajkowski, line 136: As I remember it's only true for deposits greater than 1BTC. Shouldn't we use `minConfirmationsNeeded` instead of `6` here?

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -44,6 +77,37 @@ const MintingSuccessComponent: FC<{
       <InfoBox>
         <Image src={tbtcSuccess} />
       </InfoBox>
+      <Stack my={2}>
+        <Skeleton isLoaded={txConfirmations !== undefined}>
```
r-czajkowski, line 81: Maybe let's extract this condition to a variable because we use the same condition in other places. Eg. `const areConfirmationsLoaded = txConfirmations !== undefined`.

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -44,6 +77,37 @@ const MintingSuccessComponent: FC<{
...
+                ? 100
+                : (txConfirmations / minConfirmationsNeeded) * 100
+            }
+          />
+        </Skeleton>
+        <HStack mt={1} alignSelf="flex-end">
+          <CheckCircleIcon w={4} h={4} color={checkmarkColor} />{" "}
+          <BodySm color={"gray.500"}>
+            <Skeleton
+              isLoaded={txConfirmations !== undefined}
```
r-czajkowski, line 97: Same here if you agree with the comment above.

src/tbtc/mock-bitcoin-client.ts
```diff
@@ -194,9 +196,33 @@ export class MockBitcoinClient implements Client {
     rawTransactions.set(transactionHash2.toString(), transaction2)
     this.rawTransactions = rawTransactions
 
+    this._mockConfirmationsForTransaction(depositUtxo.transactionHash)
+    this._mockConfirmationsForTransaction(depositUtxo2.transactionHash)
+
```
r-czajkowski, line 201: Why can't we call this function in `getTransactionConfirmations` method and just increment the confirmation for a given transaction hash after the delay/timeout? The transaction is not confirmed immediately so I think we should increment the confirmation inside the `getTransactionConfirmations` to better reflect the "real" flow. It works ok because we do not use `await` here. Ofc non blocking because it's mock implementation.

src/threshold-ts/tbtc/index.ts
```diff
@@ -420,6 +441,24 @@ export class TBTC implements ITBTC {
...
+  }
+
+  minimumNumberOfConfirmationsNeeded = (amount: BigNumberish): number => {
+    const amountInBN = BigNumber.from(amount)
+    if (amountInBN.lt(10000000) /* 0.1 BTC */) {
+      return 1
+    } else if (amountInBN.lt(100000000) /* 1 BTC */) {
+      return 3
+    }
+    return 0
```
r-czajkowski, line 459: Shouldn't be `6` here ?

r-czajkowski (changes requested):

src/threshold-ts/tbtc/index.ts
r-czajkowski: See https://github.com/threshold-network/token-dashboard/pull/444#discussion_r1109595820

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
michalsmiarowski: c96b83238c874efb25d1f9a0e8dc8d7a99ccc5da

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
michalsmiarowski: 702958e6228608574a647b11435686deafcc0bff

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
michalsmiarowski: 702958e6228608574a647b11435686deafcc0bff

src/threshold-ts/tbtc/index.ts
michalsmiarowski: a5f661c0315f608562a704d611640dea54ab8125

src/tbtc/mock-bitcoin-client.ts
michalsmiarowski: 065787bba782c359162920e4efaf70c1745ed047

r-czajkowski (approved):

Issue #430, opened 2023-02-02
Show links to all ETH and BTC transactions (3 total on success) 


pdyraga:
We should show links to all transactions made by the user during the deposit flow: Etherscan and Blockstream.info (or another BTC block explorer).

sashatanase:
![Image](https://user-images.githubusercontent.com/40768736/217048489-9655ae3a-ccd1-4bb2-bb6f-2c4f0bd32525.png)

https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=7907%3A209216&t=59lNq8DAGZokZBhL-1

sashatanase:
@pdyraga I have two txs on success; which one is the third?

pdyraga:
There are three transactions:
- Bitcoin deposited (transaction on Bitcoin)
- Optimistic minting requested (transaction on Ethereum)
- Optimistic minting finalized (transaction on Ethereum)

mhluongo:
Note that including all of these in a detailed mint view was my original intent. There was an interim idea because I was looking for a way to roll it out in pieces to deliver value to users sooner.

r-czajkowski:
> Note that including all of these in a detailed mint view was my original intent. There was an interim idea because I was looking for a way to roll it out in pieces to deliver value to users sooner.

I see. So displaying links to these transactions:
- Bitcoin deposited (transaction on Bitcoin),
- Deposit revealed (transaction on Ethereum).
on the `Sucess` page is OK? Or do we want to display optimistic minting requested, optimistic minting finalized as well? I wonder how many users leave the app open for at least 3 hours and wait for finalization 🤔. I'm definitely not that kind of user 😆.

mhluongo:
> I wonder how many users leave the app open for at least 3 hours and wait for finalization on the Success page 

I did, because I was afraid to lose my money 🤷 It's new tech, and people will be anxious using it with large amounts of money.

I'm fine with displaying all of them in the success page, or working on a mint details page first. As long as we're addressing the pressing user need of being able to see the system state, and we can make some sort of useful change this sprint.

mhluongo:
It's also not clear to the user whether the dApp will successfully keep the state — another reason for people to leave it open. Shipping to a mint details page with a persistent URL is the end goal as it addresses all of this, just want to iterate to that.

r-czajkowski:
[...]

PR #446, opened 2023-02-14
tBTC- show links to all ETH and BTC transactions on sucess page

r-czajkowski:
Closes: #430 

This PR adds the following links to the deposit success page:
- Bitcoin deposited (transaction on Bitcoin),
- Deposit revealed  (transaction on Ethereum),
- Optimistic minting requested (transaction on Ethereum),
- Optimistic minting finalized (transaction on Ethereum).

The last 2 of them are not displayed immediately - it takes time for the minter to request the optimistic minting and about 3 hours to finalize the optimistic minting. To display these links we subscribe to `OptimisticMintingRequested` and `OptimisticMintingFinalized` events and save the transaction hash in redux store. When the dapp catches event the link will be displayed.


src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
```diff
@@ -69,6 +78,47 @@ const MintingSuccessComponent: FC<{
...
+        {optimisticMintingFinalizedTxHash && (
+          <BodySm>
+            View Optimistic Minting Finalized transaction on{" "}
+            <ViewInBlockExplorer
+              id={optimisticMintingFinalizedTxHash}
+              type={ExplorerDataType.TRANSACTION}
+              text="Etherscan"
+            />
+          </BodySm>
+        )}
```
michalsmiarowski, line 120: Non-blocking:

I think that from the UX perspective it would be nice to print those links at the beginning, but grayed out (with some questionmar-like icon next to it and a tooltip describing why it's grayed out and when the transactions will be available to see on block explorer). It's a non-blocking issue, just a thought.

michalsmiarowski (changes requested):
Looking great! Tested it locally and it works fantastic 🔥 

Left some minor comments to look at before the merge

src/pages/tBTC/Bridge/MintingCard/MintingSuccess.tsx
r-czajkowski: Since we are going to implement #448 I think we can leave it as is now

src/types/tbtc.ts
```diff
@@ -1,21 +1,38 @@
...
+  walletPublicKeyHash: string
+  utxo: UnspentTransactionOutput
+  nextBridgeCrossingInUnix?: number
+  depositRevealedTxHash?: string
+  optimisticMintingRequestedTxHash?: string
+  optimisticMintingFinalizedTxHash?: string
+
+  // TODO: These may be incorrect types
+  tBTCMintAmount: string
+  ethGasCost: number
```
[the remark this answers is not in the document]
r-czajkowski: 6363a0f

michalsmiarowski (approved):
LGTM 🔥

PR #451, opened 2023-02-20
[Backport] Switch 2nd track deposit verifier endpoint to `keep-prd` instance

r-czajkowski:
Backport of: #450 

Refs: https://github.com/keep-network/pm/issues/33

We deployed the 2nd track verifier function in the `keep-prd` GCP project. This is now the default production instance so we are switching the dashboard to point it.


michalsmiarowski (approved):
LGTM 🧞‍♂️

PR #454, opened 2023-02-21
[Backport] Track transaction confirmations

michalsmiarowski:
Backport of #444 
Closes: #429 

Adds the bitcoin confirmation loader to the Minting Success page.

image

# Main changes:
- add `getTransactionConfirmations` method to the TBTC class
- add `minimumNumberOfConfirmationsNeeded` method to the TBTC class
- add new effect - `fetchUtxoConfirmationsEffect`
- mock transaction confirmations in the `MockBitcoinClient`

# Details of the changes

### - add `getTransactionConfirmations` method to the TBTC class

Added a new method to the TBTC class that gets the number of confirmations that a given transaction has accumulated
so far. It take hash of the transaction as an argument. It calls the `getTransactionConfirmations` method defined inside the `bitcoinClient`.

### - add `minimumNumberOfConfirmationsNeeded` method to the TBTC class

Added a new method to the TBTC class that gets the minimum number of confirmations needed for the minter to start the
minting process. The minimum number of confirmations is based on the amount
that was sent to the deposit address:
The rules are:
- If the amount is less than 0.1 BTC, it should have at least 1 confirmation.
- If the tx is less than 1 BTC, it should have at least 3 confirmations.
- Otherwise, we need to wait for 6 confirmations.

### - add new effect - `fetchUtxoConfirmationsEffect`

Added a `txConfirmations` variable to the redux store. It will store the current number of transactions (as long as it's lower or equal than `minimumNumberOfConfirmationsNeeded`. It stops updating after it's above that value). The new effect - `fetchUtxoConfirmationsEffect` is called when the user lands on the minting success page and it fetches the actual number fo the confirmation for the given utxo.

The effect works similar to `findUtxoEffect`. It calls the `getTransactionConfirmations` method every 10 seconds.

### - mock transaction confirmations in the `MockBitcoinClient`

Created a new private method in `MockBitcoinClient` called `_mockConfirmationsForTransaction` which can mock the confirmations for the given transaction. We could also specify how many confirmations we want to mock (default is 10). Every confirmation is mocked every 8 seconds.


michalsmiarowski:
@r-czajkowski I've resolved the conflicts. It's ready for review & merge.

r-czajkowski (approved):
LGTM

Issue #401, opened 2023-01-28
(COSMETIC) Display full addresses in the deposit flow


michalinacienciala:
It would be nice to display full BTC Deposit Address in the Step 2 of the deposit flow. So that the user who scanned the QR code and wants to verify if the decoded address matches the one displayed on the dashboard wouldn't have to copy it to some text editor to display it in full length.

![image](https://user-images.githubusercontent.com/78352137/215283051-686e0c3f-ee8c-4d55-ae34-b3d32d43fb11.png)

EDIT: Also other addresses should be shown in full length, not only on Step 2.

sashatanase:
We need to do this for sure. 

Some ideas open for discussion:

- To display the addresses, we need to introduce a new typeface. We could use a monospace typeface that we do not currently have in the design system only for this purpose.
- We need to increase the QR code size
- We need to add Etherscan links to the ETH address and BLOCKSTREAM links BTC Recovery Address

sashatanase:
Changes are made in the design here.

**What changed:**

- Addresses displayed in a monospaced font. 
- Changed the BTC Deposit Address Card
- made the BTC Deposit Address area bigger
- made the QR code bigger
- we need to add links towards Etherscan for the ETH address provided and towards Blockstream for the Recovery Address


![Image](https://user-images.githubusercontent.com/40768736/217043966-c65e035e-2e28-41a8-862c-a1ad96f335d3.png)


https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=7846%3A205809&t=59lNq8DAGZokZBhL-1

pdyraga:
@r-czajkowski Can we close this issue now?

r-czajkowski:
Addressed in #449

PR #449, opened 2023-02-16
Display full BTC deposit address in deposit flow

r-czajkowski:
Closes: #401 

Here we display the full BTC deposit address in the deposit flow. This PR also refactors the `CopyToClipboard` component to be more flexible so we can use the same copy logic but with different UI components.

PR #464, opened 2023-02-22
[Backport] Display full BTC deposit address in deposit flow

r-czajkowski:
Backport of: #449 
Closes: #401 

Here we display the full BTC deposit address in the deposit flow. This PR also refactors the `CopyToClipboard` component to be more flexible so we can use the same copy logic but with different UI components.


michalsmiarowski (approved):
LGTM 🔥

Issue #455, opened 2023-02-21
New Success Page + Pools Cross Promo


pdyraga:
Figma file here https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=8019%3A214048&t=2jBOz5njnaJT1oxO-1

![image](https://user-images.githubusercontent.com/4712360/220324735-2039b22b-7da9-4acb-836d-ba7c3fa99eae.png)

Issue #448, opened 2023-02-14
Show the minting progress


sashatanase:
I started from this here. 

I have this proposal for our lack of visibility of system status, and I took a stab at solving the issue.

Context: we received feedback that users need to learn what happens with their deposited tokens after they hit initialized minting. Our dApp doesn't provide any information about what is happening under the hood.

Proposal: Introduce a modal between Step 3 and Success that will showcase and sequence what is happening with the funds and the process under the hood.

Bear with me. I got down and dirty, so I might have breaks in alignments, padding, and other messy things.

https://github.com/threshold-network/token-dashboard/issues/430#issuecomment-1426045772

sashatanase:
**Feedback received**

**Matt**

I like it! A couple of thoughts

* Generally, this feels like an extensive improvement / better exposure of the system status
* Instead of a modal, something that's more friendly to mints happening in parallel would be nice... especially if each mint gets its own dedicated, persistent URL. It's already unclear if it's safe to leave the dApp during a mint
* We'll be collapsing the "Bitcoin checkpoint" and "Minting initialized" steps over the next few releases to shave time off, so be ready for that. I don't think it's wrong though

Details on the last point - https://github.com/keep-network/optimistic-minting/issues/65

sashatanase:
**Changes in version 2**

- I killed the timeline on the right-hand side since the user doesn't need to take any action anymore. they are just waiting for the dApp to do its magic. 
- used the new space to display the suite of transactions that happen during the minting process
- changed the timeline for the system status visually 
- added an animation that shows the metamorphosis from BTC to tBTC and killed the idea of a bridge
- changed the loading bar for the Bitcoin Network confirmations - there were too many linear elements and it was competing with the main timeline. 
- changed the success page. I removed some information from it - fees and stuff, and kept only the success message, the amount minted, and the transactions suite - BTC and ETH. 
- added on the success pool cards to entice the user into depositing.
- changed the Secondary nav title from _bridge_ to _dApp_

sashatanase:
**Prototype**
https://www.figma.com/proto/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?page-id=7126%3A198919&node-id=8019%3A214048&viewport=-17975%2C-10463%2C0.6&scaling=min-zoom&starting-point-node-id=8006%3A212058

**Figma file**
[...]

PR #473, opened 2023-03-02
External pool component

michalsmiarowski:
Closes: #455 

Creates `ExternalPool` component that will be used to display tBTC/wBTC/SBTC curve pool data.

# Main changes:
- fetching the curve pool data
- saving the curve pool data in redux store
- displaying the data in `ExternalPool` component

# The details

### fetching the curve pool data
The data fetch is done similar to how we fetch the USD price for ETH. I've added a new case in `extraReducers` of `tbtcSlice` which fires when `tbtc/fetchFactoryPool` (from which we created the action creator funcion using `createAsyncThunk`) is dispatched and the operations inside succeeds (`fulfilled`). 

Inside the async think we fetch the data using [curve pool api](https://github.com/curvefi/curve-api/blob/main/endpoints.md). For fetching the specific pool we use the `CurveFactoryPoolId`enum that defined all needed pool ids. For now we only need `tBTC/wBTC/SBTC` pool.

### saving the curve pool data in redux store
We are saving the curve pool data in a redux store after the operation inside the async full is fulfilled. For that I've added a new property in tbtc state - `curveTBTCPool`. This property has the type specified below:
```
export type CurvePoolData = {
  url: string
  address: string
  apy: number[]
  tvl: number
}
```
Explanation of the fields:
- **url** - urls that reirect to the deposit action of specific curve pool. In the case of `tBTC/wBTC/SBTC` it's https://curve.fi/#/ethereum/pools/factory-v2-247/deposit
- **address** - address of the pool contract 
- **apy** - apy of the given pool. There are two values in this pool returned from api that are saved as an array. We display it as `${value1} -> %{value2}%`
- **tvl** - tvl of the given pool. In the API it is placed under `usdTotal` field.

### displaying the data in `ExternalPool` component
We display the curve pool data from redux store in newly created `ExternalPool` component. The component is not displayed anywhere in this PR since it's related to the #448 and it should be displayed there.

For the showcase purposes I've displayed it in `MintingCard` component adding it under a `<Card></Card>` as follows:
```
return (
    <>
      <Card {...props} minW="0">
        (...)
      </Card>
      <ExternalPool
        title={"TBTC Curve Pool"}
        externalPoolData={{ poolName: "tBTC/WBTC/sBTC", ...curveTBTCPool }}
        mt={4}
      />
    </>
  )
```
This might be moved to a better place if needed and probably require additional code so that it displays only on the last step of success page.


mhluongo:
This looks awesome. Could we get screenshots in the description going forward? Makes it easier for those of us who aren't as deep in the repo.

mhluongo:
@r-czajkowski can you give this a look? Hoping to get a meaningful dashboard update out this week :pray:

src/store/tbtc/tbtcSlice.ts
```diff
@@ -15,6 +15,29 @@ import {
...
 } from "./effects"
+import { curveAPI } from "../../utils/curveAPI"
+import { CurveFactoryPoolId } from "../../enums"
+
+export type CurvePoolData = {
+  url: string
+  address: string
+  apy: number[]
+  tvl: number
+}
```
r-czajkowski, line 26: Can we rename it to be more abstract eg. `ExternalPoolData`? And maybe let's move this type to `types` dir?

src/types/tbtc.ts
```diff
@@ -26,6 +27,8 @@ export interface TbtcState {
   mintingFee: string
 
   transactionsHistory: FetchingState<BridgeTxHistory[]>
+
+  curveTBTCPool: CurvePoolData
```
r-czajkowski, line 31: IMO it's unnecessary to store this data in redux. I think we can create a hook that will fetch the pool data based on the Pool type(Curve, Uniswap) and by pool ID(eg. `tBTC/WBTC/SBTC`, `tBTC/USD`).

src/store/tbtc/tbtcSlice.ts
```diff
@@ -121,6 +144,13 @@ export const tbtcSlice = createSlice({
       }>
     ) => {},
   },
+  extraReducers: (builder) => {
+    builder.addCase(fetchCurveFactoryPoolData.fulfilled, (state, action) => {
+      const curveFactoryPool = action.payload
+
+      state.curveTBTCPool = curveFactoryPool
+    })
+  },
```
r-czajkowski, line 153: Same as above.

src/store/tbtc/tbtcSlice.ts
```diff
@@ -15,6 +15,29 @@ import {
...
+      CurveFactoryPoolId.TBTC_WBTC_SBTC
+    )
+
+    return {
+      address: factoryPool.address,
+      url: factoryPool.poolUrls.deposit[0],
+      apy: factoryPool.gaugeCrvApy,
+      tvl: factoryPool.usdTotal,
+    }
+  })
```
r-czajkowski, line 40: Same as above.

r-czajkowski (changes requested):
Left a few minor comments. Overall LGTM.

src/store/tbtc/tbtcSlice.ts
michalsmiarowski: Fixed in 099344f36b929dd2a62428c41c55f72349558123 and b70f471f9bcb8e0fe2841ed4bcde5aa6c37ed933

src/store/tbtc/tbtcSlice.ts
michalsmiarowski: 2f46605e6593a4b511d934ce392b79de71b110fc

src/store/tbtc/tbtcSlice.ts
michalsmiarowski: 2f46605e6593a4b511d934ce392b79de71b110fc

src/components/tBTC/ExternalPool.tsx
```diff
@@ -0,0 +1,110 @@
...
+                      )} -> ${formatPercentage(
+                        apy[1],
+                        2,
+                        false,
+                        true
+                      )} CRV`}</BodyXs>
+                    )}
+                  </Td>
+                  <Td {...commonCellProps}>
+                    {!!tvl && <BodyXs>{formatNumeral(tvl)}</BodyXs>}
```
mhluongo, line 95: If TVL is in USD, we should show that. Seeing the raw number is confusing as it could be in BTC for example

src/hooks/useFetchExternalPoolData.ts
```diff
@@ -0,0 +1,54 @@
...
+          apy: factoryPool.gaugeCrvApy,
+          tvl: factoryPool.usdTotal,
+        }
+        setData(finalData)
+        return finalData
+      } else {
+        throw new Error(
+          `Error while fetching external pool data: fetching of ${type} external pool data is not implemented.`
+        )
+      }
```
r-czajkowski, line 50: Non-blocking

Maybe we could do it in `useEffect` hook?

src/pages/tBTC/Bridge/MintingCard/index.tsx
```diff
@@ -7,11 +7,23 @@ import { useTbtcState } from "../../../../hooks/useTbtcState"
...
   const { account } = useWeb3React()
+  const [tBTCWBTCSBTCPoolData, fetchTBTCWBTCSBTCPoolData] =
+    useFetchExternalPoolData(
+      ExternalPoolType.CURVE_POOL,
+      CurveFactoryPoolId.TBTC_WBTC_SBTC
+    )
+
+  useEffect(() => {
+    fetchTBTCWBTCSBTCPoolData()
+  }, [fetchTBTCWBTCSBTCPoolData])
```
r-czajkowski, line 26: If you agree with the comment above, it will be unnecessary.

src/components/tBTC/ExternalPool.tsx
michalsmiarowski: Fixed in 8a659d7d72779bf88a77444b5a54dcdf24779ca7

I've added a dollar sign (`$`) at the beginning similarly as it is in curve:
image

Implementation:
image

Let me know if that's enough

src/hooks/useFetchExternalPoolData.ts
```diff
@@ -0,0 +1,54 @@
...
+  poolName: "",
+  address: "",
+  url: "",
+  apy: [],
+  tvl: 0,
+}
+
+export const useFetchExternalPoolData = (
+  type: ExternalPoolType,
+  poolId: CurveFactoryPoolId
```
r-czajkowski, line 16: Let's say the `ExternalPoolType` will be `UNISWAP` so then the `CurveFactoryPoolId` is too specific. It should be `ExternalPoolId` and it should depend on the `type` argument. Eg:

```ts
const pools = {
  curve: ["id1", "id2"],
  uniswap: ["id3", "id4"],
} as const

type ExternalPool = keyof typeof pools
type ExternalPoolId<T extends ExternalPool> = typeof pools[T][number]

const useFetchExternalPoolData = <T extends ExternalPool>(type: T, poolId: ExternalPoolId<T>) => {
  ...
}

useFetchExternalPoolData("uniswap", "id3")
```

r-czajkowski (changes requested):

src/hooks/useFetchExternalPoolData.ts
r-czajkowski, line 50: If we add more pools there will be lots of if statements. If think we could implement a kind of strategy pattern something like this:

```ts
type ExternalPoolData = {}

const fetchCurvePool: (poolId: ExternalPoolId<"curve">) => ExternalPoolData = (
  poolId
) => {
  return {}
}

const fetchUniswapPool: (
  poolId: ExternalPoolId<"uniswap">
) => ExternalPoolData = (poolId) => {
  return {}
}

const fetchPoolDataStrategy: {
  [key in ExternalPool]: (poolId: ExternalPoolId<key>) => ExternalPoolData
} = {
  curve: fetchCurvePool,
  uniswap: fetchUniswapPool,
}
```

and then in hook:
```ts
const fetchPoolData = fetchPoolDataStrategy[type]
const data = fetchPoolData(poolId)
```

r-czajkowski (changes requested):

src/pages/tBTC/Bridge/MintingCard/index.tsx
michalsmiarowski: e8d2579cb6547177500a2024fe6209d3403fe6f8

src/hooks/useFetchExternalPoolData.ts
michalsmiarowski: 64318b6b52fe18fa9027b14fef59d58aa5941b60

src/hooks/useFetchExternalPoolData.ts
michalsmiarowski: f448fb6c2d8761a231297bc45c57df4647e723cf

src/hooks/useFetchExternalPoolData.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  useEffect(() => {
+    const fetchExternalPoolData = async () => {
+      const fetchPoolData = fetchPoolDataStrategy[type]
+      const data = await fetchPoolData(poolId)
+      setData(data)
+    }
+
+    fetchExternalPoolData()
+  }, [type, poolId, fetchPoolDataStrategy])
```
r-czajkowski, line 71: `fetchPoolDataStrategy` is unnecessary because it's hoisted outside of the hook scope.

r-czajkowski (approved):
Left a minor comment. LGTM! 🚀

src/hooks/useFetchExternalPoolData.ts
michalsmiarowski: 7e9e8526559927498daa9e38cb1e2120ecb50dbb

r-czajkowski (approved):
:shipit:

Issue #456, opened 2023-02-21
Add a persistent link to the deposit flow so people can share them with support 


pdyraga:
Generate a unique link after deposit data are generated and update it in the browser. Redirect to the right step after the link is opened.

michalsmiarowski:
@pdyraga @sashatanase I would like specify the issue a little bit more here so that I'm sure we all agree on how it suppose to look like.

We have 4 steps:
Step 1 - Provide data (which is eth address and btc recovery address)
Step 2 - Make Deposit (display of the deposit address, user did not send any fund there yet)
Step 3 - Initiate Minting (user send funds to the deposit address)
Step 4 - Success step with the display of the whole minting process status (see #448)

My understanding is that we should create a unique link at the step 2 (so i assume it will probably has a generated deposit address in url param) and after pasting it in the url it will display the correct step based on the deposit status. There are some question that arises on my end:
1. How should we display it for the user that did not generated that address? Should we  display it the same way as for the author but without the possibility to execute any transactions? So display the step in which the author has finished his deposit flow?
2. How should we display it for the user that generated that address but is connecting on another browser/machine where he doesn't have any deposit data saved in the local storage? Should we display it the same way as in the point 1? Or should we allow him to do the transactions related to that deposit on another browser (i think this will require to put all deposit data in the url params)?.

sashatanase:
I may not fully understand your question, so please bear with me. 
> How should we display it for the user not generating that deposit address? Should we display it the same way as for the author but without the possibility of executing any transactions? So display the step in which the author has finished his deposit flow?

If the user did not generate the deposit address, he is still on Step 1 and has barely started the flow. `Generate Deposit Address` is the CTA of Step 1, Step 2 is the outcome of Step 1, and it shows the unique P2KH address for the user. In Step 2, the dApp waits for the user to make the deposit, continuously checking if the user has sent the funds. 

I hope this answer is helpful @michalsmiarowski 

> How should we display it for the user that generated that address but is connecting on another browser/machine where he doesn't have any deposit data saved in the local storage? Should we display it the same way as in the point 1? Or should we allow him to do the transactions related to that deposit on another browser (i think this will require to put all deposit data in the url params)?.

This is a @pdyraga question.

michalsmiarowski:
[...]

PR #475, opened 2023-03-10
Show the minting progress 

r-czajkowski:
Closes: #448 
Ref: #456 

We received feedback that users need to learn what happens with their deposited tokens after they hit initialized minting. Here we add support for tracking the minting process for a given deposit. After initializing the minting the dapp redirects users to the deposit details page `tBTC/mint/deposit/<deposit_key>` where can track the current status of the minting process.

There are 4 steps, and each step redirects to the next one after 10 seconds once it is completed:
1. **Bitcoin confirmations**- there is a rule that requires a number of transaction confirmations to be met for an optimistic minting to be requested. The step is completed once the Bitcoin deposit transaction has the required number of confirmations.
2. **Minting Initialized**- at this step the dapp waits for the `OptimisticMintingRequested` event. Once it catches this event and the `depositKey` event param matches this one from an URL the step is completed.
3. **Guardian Check**- at this step the dapp waits for the `OptimisticMintingFinalized` event. The guardian doesn't submit any transaction if everything is OK with a deposit so we can say this step is completed once the dapp catches the `OptimisticMintingFinalized` event and the `depositKey` event param matches this one from an URL.
4. **Minting in progress**- this step is marked as complete by default because the minting happens in the same transaction where the `OptimisticMintingFinalized` event is emitted. After 10 seconds the user will be redirected to the success page

Once all steps are completed user is redirected to the success page where can see all links to the transactions, start a new mint or go to the external pool where can deposit minted tBTC tokens.

If a user goes to the page `tBTC/mint/deposit/<deposit_key>` directly the current step of the minting process should be displayed.

<details>
  
obraz

obraz

obraz

obraz

obraz

obraz

obraz

obraz

</details>


src/hooks/tbtc/useFetchDepositDetails.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  useEffect(() => {
+    const fetch = async () => {
+      setIsFetching(true)
+      try {
+        const { depositor } = (await threshold.tbtc.bridgeContract.deposits(
+          depositKey
+        )) as {
+          depositor: string
+        }
```
michalsmiarowski, line 30: Non-blocking:

Maybe it would be worth it to create separate method that returns `Promise<string>` in our tbtc interface for `deposits`? This would get rid of that type cast.

src/hooks/tbtc/useFetchDepositDetails.ts
```diff
@@ -0,0 +1,97 @@
...
+        const optimisticMintingFinalizedEvents = await getContractPastEvents(
+          threshold.tbtc.vaultContract,
+          {
+            eventName: "OptimisticMintingFinalized",
+            fromBlock: deposit.blockNumber,
+            filterParams: [undefined, depositKey, depositor],
+          }
+        )
+
+        const btcTxHash = TransactionHash.from(deposit.fundingTxHash).reverse()
```
michalsmiarowski, line 67: Non-blocking:

After a while I forgot why we reverse the transaction hash in some cases and in some not. If that's the issue only for the `fundingTxHash` that is returned from tbtc-related events then maybe it would be worth creating a separate methods calle `extractBtcTransactionHashFromEvent` or `getTransactionHashFromFundingTx` and do the whole reversing inside it?

src/hooks/tbtc/useSubscribeToOptimisticMintingFinalizedEvent.ts
```diff
@@ -8,17 +8,38 @@ import { useTbtcState } from "../useTbtcState"
...
+  )
+}
+
+export const useSubscribeToOptimisticMintingFinalizedEvent = () => {
+  const threshold = useThreshold()
+  const { updateState, utxo, mintingStep } = useTbtcState()
+  const dispatch = useAppDispatch()
+  const { account } = useWeb3React()
+
+  useSubscribeToOptimisticMintingFinalizedEventBase(
```
michalsmiarowski, line 42: Non-blocking:

There are some unused variables below. If not needed then let's remove `miner`, `depositor` and `optimisticMintingDebt`

src/hooks/tbtc/useSubscribeToOptimisticMintingRequestedEvent.ts
```diff
@@ -6,16 +6,39 @@ import { useTbtcState } from "../useTbtcState"
...
+    shouldSubscribeIfUserNotConnected
+  )
+}
+
+export const useSubscribeToOptimisticMintingRequestedEvent = () => {
+  const threshold = useThreshold()
+  const { updateState, utxo, mintingStep } = useTbtcState()
+  const { account } = useWeb3React()
+
+  useSubscribeToOptimisticMintingRequestedEventBase(
```
michalsmiarowski, line 41: Non-blocking:

Same as abov we have unused variables here. Let's remove `minter`, `depositor`, `amount`, `fundingTxHash` and`fundingOutputIndex` if they are not needed.

michalsmiarowski (commented):
First part of the review done! As of now only non-blocking comments. More to come... (probably)

src/hooks/tbtc/useSubscribeToOptimisticMintingFinalizedEvent.ts
r-czajkowski: I think we need to "extract" all values because we need the last param `event`. I don't know a better solution for that other than that:
```
(...args) => {
      const [, depositKey, , , event] = args
      ...
}
```
but IMO the readability is worse.

src/hooks/tbtc/useSubscribeToOptimisticMintingRequestedEvent.ts
r-czajkowski: See https://github.com/threshold-network/token-dashboard/pull/475#discussion_r1132831477

src/hooks/tbtc/useFetchDepositDetails.ts
r-czajkowski: We actually have the `getRevealedDeposit` method but it accepts only the utxo- in this case, we only have the deposit key. I would like to keep the tbtc interface simple(but actually is not as simple as I wish 😆) so adding a new method like `getDepositorByDepositKey` doesn't sound like a good idea. We use the `depositor` to determine if a given deposit exists if not depositor is a zero address. We could also use other values from the [DepositRequest](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/Deposit.sol#L81) struct eg. `amount` and check is equal to `0`.

src/pages/tBTC/Bridge/DepositDetails.tsx
```diff
@@ -0,0 +1,564 @@
...
+      </Card>
+    </DepositDetailsPageContext.Provider>
+  )
+}
+
+DepositDetails.route = {
+  path: "deposit/:depositKey",
+  index: false,
+  isPageEnabled: true,
+}
```
michalsmiarowski, line 272: Non-blocking and open to discussion (and it is probably to do in future PRs, not this one):

The routes seems hidden everywhere. I wouldn't know where to look at them (but maybe thats just me) and not every component in `pages` directory have it. I would think of a way to keep it simple and in one file, maybe in `index` of each directory? For example all links for tbtc would be in `pages/tBTC/index.tsx` or even in a separate files. Not sure if it's possible to extract them to a separate file tho and didn't think of a way to do so yet, but it's just an idea to think about.

src/pages/tBTC/Bridge/MintUnmintNav.tsx
```diff
@@ -1,22 +1,27 @@
...
+    tabId={index.toString()}
+  >
+    {page.route.title}
+  </FilterTab>
+)
 
+export const MintUnmintNav: FC<
+  ComponentProps<typeof Card> & { pages: PageComponent[] }
+> = ({ pages, ...props }) => {
+  // TODO: set the proper selectedTabId based on the current location.
```
michalsmiarowski, line 20: Let's not forget about this TODO

src/pages/tBTC/Bridge/DepositDetails.tsx
```diff
@@ -0,0 +1,564 @@
...
+      )
+    case "guardian-check":
+      return (
+        <Step3
+          txHash={optimisticMintingFinalizedTxHash}
+          onComplete={onComplete}
+        />
+      )
+    case "minting-completed":
+      return <Step4 onComplete={onComplete} />
```
michalsmiarowski, line 482: I think we should improve the way those steps are shown. When we have a deposit with minting process completed then it would print a completed `Step2` for 10 secs, then completed `Step3` for 10secs, and then complete `Step4`. I believe it should immidietaly show step4 in such case.

Similarly if the minting was requested but not finished yet then we should immidietaly show `Step3` when going into it from a link. No need to display completed `Step2` in such case.

Bear in mind that i meant it for someone who goes to the page from the url. For someone who is doing the minting process we should keep displaying the completed steps for x seoncds before moving to the next one.

src/theme/DotsLoadingIndicator.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  return "gray"
+}
+
+const baseDotStyles: SystemStyleObject = {
+  borderRadius: "50%",
+}
+
+const baseStyle: SystemStyleFunction = (props) => {
```
michalsmiarowski, line 33: If you agree with , then we could move this theme to `@threshold-network/components` too.

michalsmiarowski (changes requested):
Left some additional review comments

src/pages/tBTC/Bridge/components/DepositDetailsStep.tsx
```diff
@@ -0,0 +1,213 @@
...
+      </BodySm>
+    </HStack>
+  )
+}
+
+type CommonStepProps = Pick<StepTemplateProps, "onComplete"> & {
+  txHash?: string
+}
+
+export const Step1: FC<
```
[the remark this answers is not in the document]
r-czajkowski: We are in the `DepositDetailsStep` file so I think it's necessary. Also when we import components the path is `../components/DepositDetailsStep` so I think it's clear. We can also use aliases if needed eg. `import { Step1 as DepositDetailsStep1 } from "../components/DepositDetailsStep"`.

src/pages/tBTC/Bridge/components/DepositDetailsStep.tsx
```diff
@@ -0,0 +1,213 @@
...
+        confirmations &&
+          requiredConfirmations &&
+          confirmations >= requiredConfirmations
+      )}
+      onComplete={onComplete}
+    />
+  )
+}
+
+export const Step2: FC<CommonStepProps> = ({ txHash, onComplete }) => {
```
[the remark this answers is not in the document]
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/475#discussion_r1136693841

src/pages/tBTC/Bridge/components/DepositDetailsStep.tsx
```diff
@@ -0,0 +1,213 @@
...
+      txHash={txHash}
+      progressBarColor="yellow.500"
+      isCompleted={!!txHash}
+      onComplete={onComplete}
+      isIndeterminate
+    />
+  )
+}
+
+export const Step3: FC<CommonStepProps> = ({ txHash, onComplete }) => {
```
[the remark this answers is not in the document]
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/475#discussion_r1136693841

src/pages/tBTC/Bridge/components/DepositDetailsStep.tsx
```diff
@@ -0,0 +1,213 @@
...
+      chain="ethereum"
+      progressBarColor="green.500"
+      isCompleted={!!txHash}
+      onComplete={onComplete}
+      isIndeterminate
+    />
+  )
+}
+
+export const Step4: FC<CommonStepProps> = ({ txHash, onComplete }) => {
```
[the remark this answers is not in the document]
r-czajkowski: https://github.com/threshold-network/token-dashboard/pull/475#discussion_r1136693841

src/pages/tBTC/Bridge/components/DepositDetailsStep.tsx
```diff
@@ -0,0 +1,213 @@
...
+export const Step1: FC<
+  { confirmations?: number; requiredConfirmations?: number } & Pick<
+    StepTemplateProps,
+    "txHash" | "onComplete"
+  >
+> = ({ confirmations, requiredConfirmations, txHash, onComplete }) => {
+  return (
+    <StepTemplate
+      title="Waiting for the Bitcoin Network Confirmations..."
+      subtitle={`The Bitcoin Deposit transaction needs to get ${requiredConfirmations} confirmations on the Bitcoin Network before the minting is initialised.`}
```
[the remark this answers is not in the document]
r-czajkowski: 1313558

src/components/DotsLoadingIndicator/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  Box,
+  HStack,
+  StackProps,
+  ThemingProps,
+  useStyleConfig,
+} from "@threshold-network/components"
+
+type DotsLoadingIndicatorProps = StackProps & Omit<ThemingProps, "orientation">
+
+export const DotsLoadingIndicator: FC<DotsLoadingIndicatorProps> = ({
```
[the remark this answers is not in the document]
r-czajkowski: Yep definitely. I'll address it in a separate PR.

src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,82 @@
...
+  useStyles,
+  BoxProps,
+} from "@threshold-network/components"
+
+export type TimelineItemStatus = "active" | "semi-active" | "inactive"
+type TimelineItemProps = {
+  status: TimelineItemStatus
+}
+
+export const Timeline: FC = ({ children, ...props }) => {
```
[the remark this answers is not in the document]
r-czajkowski: `Timline` is a common component for displaying a list of events in chronological order so I don't think so it's too generic. The Timeline can be `horizontal` or `vertical` so we can use `orientation` prop from the Chakra `ThemingProps` type to define orientation of timeline instead of a specific name. The current implementation doesn't support it but once we move it to the https://github.com/threshold-network/components repo we can add it.

We should reverse your thinking 😆- if we should rename it should be a timeline component from tBTC card because this component's name is too generic.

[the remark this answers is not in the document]
r-czajkowski: Actually, the timeline in tbtc card is `MintingTimeline` so in my opinion the name is OK and describes the purpose of the component well.

src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <StylesProvider value={styles}>
+      <OrderedList className="timeline" sx={styles.container}>
+        {children}
+      </OrderedList>
+    </StylesProvider>
+  )
+}
+
+export const TimelineItem: FC<TimelineItemProps> = ({
```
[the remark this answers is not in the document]
r-czajkowski: See https://github.com/threshold-network/token-dashboard/pull/475#discussion_r1136749816 If I had to rename something, it would be the `TimelineItem` from `src/pages/tBTC/Bridge/components/TimelineItem`.

Regarding the timeline in tBTC card- I believe we should build the tBTC timeline using this component(from `src/components/Timeline`) with `orientation="vertical"` prop. But it's out of scope of this PR so let's address it in a separate PR.

src/pages/tBTC/Bridge/DepositDetails.tsx
r-czajkowski: Probably you tested it out with the mocked bitcoin client. It's because we save the confirmations by `txHash` in the `MockBitcoinClient` class field so when you refresh the page this class is recreated and loses its state. Then `useFetchDepositDetails` hook returns `0` in `confirmations` field so the flow will start from step 1. It should work ok with the "real" Electrum client.

src/hooks/tbtc/useFetchDepositDetails.ts
r-czajkowski: 31e47a6 here is a hook that wraps the whole reversing.

src/pages/tBTC/Bridge/MintUnmintNav.tsx
r-czajkowski: 66427ff

src/pages/tBTC/Bridge/DepositDetails.tsx
```diff
@@ -0,0 +1,564 @@
...
+      }
+    },
+    undefined,
+    true
+  )
+
+  return { mintingRequestedTxHash, mintingFinalizedTxHash }
+}
+
+// TODO: Update link and copy for subtitle!
```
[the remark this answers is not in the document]
r-czajkowski: 48f933b

src/pages/tBTC/Bridge/DepositDetails.tsx
r-czajkowski: You are right! Thanks for catching this. Fixed in 5cf4978.

src/hooks/tbtc/useSubscribeToOptimisticMintingFinalizedEvent.ts
michalsmiarowski: Got it! Let's keep it as it is now then

src/components/Timeline/Timeline.tsx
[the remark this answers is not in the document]
michalsmiarowski: Ah i see! Yes, you definitely right about this, now i see it 😅  Gonna resolve this conv.

src/components/Timeline/Timeline.tsx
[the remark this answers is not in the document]
michalsmiarowski: > Regarding the timeline in tBTC card- I believe we should build the tBTC timeline using this component(from src/components/Timeline) with orientation="vertical" prop. But it's out of scope of this PR so let's address it in a separate PR.

Gonna keep this convi as unresolved  then (even though we resolved the initial issue) to not forget about that. It's probably very low prority though

src/theme/index.ts
```diff
@@ -11,9 +11,18 @@ import { Checkbox } from "./Checkbox"
...
 import { AnnouncementBanner } from "./AnnouncementBanner"
 import { SecondaryAnnouncementBanner } from "./SecondaryAnnouncementBanner"
+import { Timeline } from "./Timeline"
+import { DotsLoadingIndicator } from "./DotsLoadingIndicator"
 
 const index = extendTheme({
   ...defaultTheme,
+  colors: {
+    ...defaultTheme.colors,
+    // TODO: move color to components repo.
```
michalsmiarowski, line 21: Non-blocking:

Let's not forget to address this in the future

michalsmiarowski (approved):
LGTM 👍

PR #484, opened 2023-03-30
Display fees on the success page

r-czajkowski:
Depends on: #483 

Here we display the minting fees on the success page. We should also take into account the fees while displaying the minted amount. The minted tBTC amount is slightly lower than requested due to fees.


michalsmiarowski (approved):
LGTM

Issue #447, opened 2023-02-14
Skip step 1&2 and jump to step 3 in the process of minting tBTC


dimpar:
Due to a browser and wallet incompatibility or simply by rejecting a transaction by a user in step 3, it might happen that the local storage will be lost. The problem is that a user doesn't know how to skip steps 1&2  and go straight to step 3 to resume minting. One of the options is to recreate local storage and paste JSON with the following format:
```
{
    "<eth_address>":
    {
        "ethAddress": "<eth_address>",
        "blindingFactor": "<blinding_factor>",
        "btcRecoveryAddress": "<btc_recovery_address>",
        "walletPublicKeyHash": "<wallet_publick_key_hash>",
        "refundLocktime": "<lock_time>",
        "btcDepositAddress": "<generated_deposit_address>"
    }
}
```
but this is not a friendly user experience.
![image](https://user-images.githubusercontent.com/3156420/218767870-07fb957c-1708-46c1-adb9-5315d656029b.png)

sashatanase:
Watching this.

Issue #457, opened 2023-02-21
Resume a deposit


pdyraga:
#456 will open the deposit flow at the right step but it will not handle revealing the deposit. When someone downloaded JSON but did not reveal it, we need a UI to upload that JSON and resume the deposit.

For this issue we need to:
- [x] Design the resume from the JSON.
- [x] Implement it!

sashatanase:
It was mentioned at some point, minting in parallel from the same address. Is this still a thing? @mhluongo said it yesterday, and I was wondering how someone can mint in parallel with the same wallet address. 

I might've misheard it, but I want to ensure I have valid data.

pdyraga:
For now, this is just about having a unique URL for a deposit. Allowing for multiple mints simultaneously requires more work - we need to check what data we store in the browser's storage. We'll have a separate issue for that.

mhluongo:
Confirming though that it is possible @sashatanase — I can stagger 10 deposits at once if I want. Our dApp won't support it, but I could do it with multiple browser profiles or by using EtherScan directly.

There's not much reason to do this as a user if you plan ahead, but I already did it once on accident to save time :see_no_evil:

sashatanase:
**Context** 

We need to provide the users with a way of resuming their minting. I assume that most cases with the revealing issues will be due to gas errors triggered by low funds or gas set too low. 


**Solution**

In the event of a transaction error as soon as the user hits `Dismiss` they will be shown the Mint Resuming page inside of the dApp



![Image](https://user-images.githubusercontent.com/40768736/222213557-92a752bf-9657-4080-ae75-9dfdf0376aee.png)



![Image](https://user-images.githubusercontent.com/40768736/222213654-522ca75f-7506-4939-83d9-d81685136330.png)

As soon as they Upload  and Click Resume this is what they'll see


![Image](https://user-images.githubusercontent.com/40768736/222213819-1b3d999c-1f6b-4964-b03f-cac3b028ae2f.png)



![Image](https://user-images.githubusercontent.com/40768736/222213909-f2e1e43f-b712-4dc5-b6e4-94e484408ff5.png)

We will also show in the My Activity Card that they have a mint that needs to be resumed.



![Image](https://user-images.githubusercontent.com/40768736/222214222-4cef207c-229d-4489-82cf-87805228f121.png)



In case the user's browser closes due to various reasons as soon as they open the dApp in the browser and connect their wallet this is what they see 



[...]

PR #495, opened 2023-04-13
Resume deposit

r-czajkowski:
Ref: #457
Closes: #447 

This PR adds support for resuming a deposit based on the uploaded recovery file. The user can resume a deposit at `/tBTC/mint/continue` page and needs to upload the recovery file. After uploading a file and submitting a form a user is redirected to the correct step of the flow or to the success page(if the deposit is already revealed). The connected wallet must match the address of the depositor in the recovery file because to reveal a deposit the transaction must be called by the same Ethereum address as the one used in the P2(W)SH BTC deposit transaction as a depositor.

### Main changes
- add a new page where user can resume a deposit,
- update the minting page routing to share the same layout in different URL,
- update `Tooltip` component theme- we are going to move theme to a component repo in a follow-up work,
- update the `@threshold-network/components` dependency to use the fixed version of the `FileUploader` component,
- fix some UI bugs and update copy based on the [coda docs](https://coda.io/d/Building-Keep_d-fmEgBNFVH/Sasha-2023-27-03-QA-Template-tBTC-v2_suu17#_lueD-)

<details>
  
obraz
  
</details>


src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const isError = Boolean(error)
+
+  const onFileUpload = async (file: File | null) => {
+    if (!file) {
+      setValues({ depositParameters: null })
+      return
+    }
+
+    try {
+      const result = (await parseJsonFile(file)) as DepositDetails
```
michalsmiarowski, line 93: I think we should also check if uploaded file is a JSON file before using `parseJsonFILE`. When uploading different file the dApp crashes:

image

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  return (
+    <Form>
+      <FileUploader
+        onFileUpload={onFileUpload}
+        headerHelperText="Require"
+        accept="text"
+      />
```
michalsmiarowski, line 110: Non-blocking:

Related to above but we could also hide all the files that are not of a JSON type when we click `Choose File`. But it's not mandatory if we will check type of the file after upload so if it's something hard to implement the I would not stress about it.

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,175 @@
...
+      !values.depositParameters?.depositor ||
+      !values.depositParameters?.depositor?.identifierHex ||
+      !isAddress(values.depositParameters?.depositor?.identifierHex) ||
+      !values.depositParameters?.refundLocktime ||
+      !values.depositParameters?.refundPublicKeyHash ||
+      !values.depositParameters?.blindingFactor ||
+      !values.depositParameters?.walletPublicKeyHash ||
+      !values.depositParameters?.btcRecoveryAddress
+    ) {
+      errors.depositParameters = "Invalid JSON. file."
```
michalsmiarowski, line 153:
```suggestion
      errors.depositParameters = "Invalid .JSON file."
```

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  return (
+    <Form>
+      <FileUploader
+        onFileUpload={onFileUpload}
+        headerHelperText="Require"
+        accept="text"
+      />
+      {isError && error}
```
michalsmiarowski, line 111: I think it would look better if we set the color of the error to a red one

src/pages/tBTC/Bridge/ResumeDeposit.tsx
michalsmiarowski: We could also check if the JSON inside this file is valid.

src/web3/utils/files.ts
```diff
@@ -18,3 +18,13 @@ export const downloadFile = (
   a.dispatchEvent(clickEvt)
   a.remove()
 }
+
+export async function parseJsonFile(file: File) {
```
michalsmiarowski, line 22: Non-blocking:

Maybe:
```suggestion
export async function parseJSONFile(file: File) {
```
?

michalsmiarowski (changes requested):
First round of review done. Probably more to come soon. Overall looks very good!

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: > I think we should also check if uploaded file is a JSON file before using `parseJsonFILE`. When uploading different file the dApp crashes:
> 
> image

I noticed that we save the recovery file as plain text so it's probably impossible to check if this file is in JSON format Probably to achieve that we should save the recovery file as `application/json`. But IMO we are ok and checking that content inside the file is a valid JSON should be enough.

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: See https://github.com/threshold-network/token-dashboard/pull/495#discussion_r1177505856 we save the recovery file as text so tbh I don't know how to hide other files probably we should save the file as `application/json`.

src/web3/utils/files.ts
r-czajkowski: 35a1cc6

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: 35a1cc6

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: > We could also check if the JSON inside this file is valid.

35a1cc6

src/pages/tBTC/Bridge/Minting/MintingTimeline.tsx
```diff
@@ -53,7 +59,7 @@ export const MintingTimelineStep2: FC<MintingTimelineStepProps> = ({
       isActive={isActive}
       isComplete={isComplete}
       stepText="Step 2"
-      helperLabelText="ACTION ON BITCOIN NETWORK"
+      helperLabelText="ACTION ON BITCOIN"
```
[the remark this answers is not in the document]
r-czajkowski: bf08a52

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const { setDepositDataInLocalStorage } = useTBTCDepositDataFromLocalStorage()
+  const threshold = useThreshold()
+
+  useEffect(() => {
+    updateState("mintingStep", MintingStep.InitiateMinting)
+
+    return () => {
+      updateState("mintingStep", MintingStep.ProvideData)
+    }
+  }, [updateState])
```
michalsmiarowski, line 44: I'm not sure why we need that `useEffect` here to be honest. Can you explain it to me why we need it and why we set the initial step here to `InitiateMinting`?

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,187 @@
...
+    setDepositDataInLocalStorage({
+      ethAddress: restDepositParameters?.depositor.identifierHex!,
+      blindingFactor: restDepositParameters?.blindingFactor!,
+      btcRecoveryAddress: btcRecoveryAddress!,
+      walletPublicKeyHash: restDepositParameters?.walletPublicKeyHash!,
+      refundLocktime: restDepositParameters?.refundLocktime!,
+      btcDepositAddress,
+    })
+
+    updateState("mintingStep", MintingStep.ProvideData)
```
michalsmiarowski, line 65: Do we need to update the `mintingStep` to `ProvideData` here if we actually update it in a cleanup function in the `useEffect` above?

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,187 @@
...
+    return getErrorsObj(errors)
+  },
+  handleSubmit: (values, { props }) => {
+    props.onSubmitForm(values)
+  },
+  displayName: "ResumeDepositForm",
+})(ResumeDepositForm)
+
+ResumeDepositPage.route = {
+  path: "continue",
```
michalsmiarowski, line 184: I think it won't be easy to implement displaying unrevealed mints in `my activity` tab and we might not implement it in this PR. In such case It would be a wise idea to put a resume link someweher in the dApp. WYT?

src/theme/Tooltip.ts
```diff
@@ -0,0 +1,29 @@
+import { theme } from "@chakra-ui/react"
+import { SystemStyleFunction, mode, cssVar } from "@chakra-ui/theme-tools"
+
+const $bg = cssVar("tooltip-bg")
+const $arrowBg = cssVar("popper-arrow-bg")
+
+// TODO: move to the components repo.
```
michalsmiarowski, line 7: Let's not forget about that. We can merge it now and fix it in separate PR if you want.

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,187 @@
...
+          ? error.message
+          : "Unexpected error while reading file, try again..."
+      )
+    }
+  }
+
+  return (
+    <Form>
+      <FormControl isInvalid={isError}>
+        <FileUploader onFileUpload={onFileUpload} headerHelperText="Required" />
```
michalsmiarowski, line 120: Let's update the `@threshold-network/components` repo in `yarn.lock` so that we are using the newest version of the `FileUploader`:

1. Delete the `@threshold-network/components` package from `package.json`
2. `yarn install` (this should remove it from `yarn.lock` )
3. Add the `@threshold-network/components` package back with the `development` tag
4. `yarn install`

michalsmiarowski (changes requested):
Left some more comment but overall looks very good and it works wonderful!

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: To display the correct step on the timeline. This should be step 3 according to Figma views.

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: Yeah, I agree but I need Sasha here. We can merge as is and even release to allow users to resume deposits through the link and the ship improvements later. I mentioned unrevealed mints in `my activity` card [here](https://github.com/threshold-network/token-dashboard/issues/457#issuecomment-1525803287).

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: 86c9ea1

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: 44e6a1b

src/pages/tBTC/Bridge/ResumeDeposit.tsx
```diff
@@ -0,0 +1,186 @@
...
+      refundLocktime: restDepositParameters?.refundLocktime!,
+      btcDepositAddress,
+    })
+
+    navigate("/tBTC/mint")
+  }
+
+  return (
+    <>
+      <TbtcMintingCardTitle />
```
michalsmiarowski, line 70: Non-blocking but good to have:

Could we add a back arrow here so that the user can go back to the main minting page from the `tbtc/mint/continue`?

Designs:
image

https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?type=design&node-id=8604-249512&t=TVSv26q2a4z87ey4-0

michalsmiarowski (approved):
Left one non-blocking comment but overall I think we can merge.

src/pages/tBTC/Bridge/ResumeDeposit.tsx
r-czajkowski: f7018bc

michalsmiarowski (approved):

PR #505, opened 2023-04-28
Fix the `RecentDeposits` component

r-czajkowski:
Align content in the list so each item always takes the same size.


michalsmiarowski (approved):
LGTM 🔥

Issue #469, opened 2023-03-01
tBTC Explore - Stats Page inside of the dApp


sashatanase:
Design 
<img width="942" alt="image" src="https://user-images.githubusercontent.com/40768736/222143913-2c709b14-d2f5-4e2e-90e0-c5015372537e.png">
Figma File - https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?node-id=8224%3A246360&t=F2GkIBFvRzvHAH3c-1

sashatanase:
@pdyraga I opened a separate issue for this.

pdyraga:
Unassigning @sashatanase given the design part is over and it is queued for implementation now.

pdyraga:
Can we close this issue now @r-czajkowski ?

r-czajkowski:
> Can we close this issue now @r-czajkowski ?

Nope. We implemented tBTC stats on overview page see https://github.com/threshold-network/token-dashboard/issues/428 and https://github.com/threshold-network/token-dashboard/issues/458. As I remember the separate page for the tBTC explorer was a stretch goal for the last sprint. I'll take this task for the next sprint.

PR #511, opened 2023-05-05
tBTC Explorer page

r-czajkowski:
Closes: #469 

This PR adds the tBTC Explorer page. Here we display:
- the tBTC total supply,
- tBTC TVL in USD,
- total mints,
- total tBTC holders,
- recent deposits.

<details>
  
  
obraz

</details>


src/pages/tBTC/Explorer/index.tsx
michalsmiarowski, line 1: We should fix this page in mobile view. It looks bad when the width is less than 1280 px:

image

src/pages/tBTC/Explorer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+                <Th textAlign={"right"}>
+                  <LabelXs>Timestamp</LabelXs>
+                </Th>
+              </Tr>
+            </Thead>
+            <Tbody>{latestMints.map(renderHistoryRow)}</Tbody>
+          </Table>
+          <BodySm color="gray.500" mt="10">
+            Showing {MINTS_TO_DISPLAY} out of {metrics.totalMints} transactions
+          </BodySm>
```
michalsmiarowski, line 134: Are we planning to add arrows so that we are able to see more all mints instead of last ten?

Designs:
image

src/pages/tBTC/Explorer/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+                </Th>
+                <Th>
+                  <LabelXs>Wallet Address</LabelXs>
+                </Th>
+                <Th>
+                  <LabelXs>TX Hash</LabelXs>
+                </Th>
+                <Th textAlign={"right"}>
+                  <LabelXs>Timestamp</LabelXs>
+                </Th>
```
michalsmiarowski, line 127: Non-blocking:

I believe it would be more user-friendly if we add some color (like gray-ish) on gover of each row. This way the user will know that he is clicking on the exact mint.

Same thing would be useful for `tBTC BRIDGE STATS` in Overview page.

michalsmiarowski (changes requested):
Looking good! Left some comments to look at before the merge

src/pages/tBTC/Explorer/index.tsx
r-czajkowski: Yes, it's on my plate. Probably I'll open a separate PR.

src/pages/tBTC/Explorer/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+              </Tr>
+            </Thead>
+            <Tbody>{latestMints.map(renderHistoryRow)}</Tbody>
+          </Table>
+          <BodySm color="gray.500" mt="10">
```
michalsmiarowski, line 140:
```suggestion
          <BodySm color="gray.500" mt="10" mb="2">
```

I think we could add some margin-bottom here so that it looks better on smaller screens when there is a horizontal scroll in the table.

Current implementation:
image

After adding margin-bottom:
image

michalsmiarowski (changes requested):
Left one small comment to look at before the merge and we are good to go

src/pages/tBTC/Explorer/index.tsx
r-czajkowski: 7654dfb

michalsmiarowski (dismissed):
LGTM 🎋

michalsmiarowski (approved):
LGTM

PR #514, opened 2023-05-12
Legacy Keep Stake- take into account `OwnerRefreshed` event

r-czajkowski:
The Threshold Staking contract allows users to refresh the owner address for legacy KEEP stakes so in that case we have to take into account the `OwnerRefreshed` event while fetching all stakes for a given owner.  Here we add a new method to the `Staking` interface- `findRefreshedKeepStakes` which returns the current and outdated staking providers for a given owner address. The outdated array is necessary while fetching all owner stakes. We need to filter out outdated staking providers eg. the `Staked` event was emitted with the owner address that we are looking for but then the owner was changed(`OwnerRefreshed` event emitted with `oldOwner = owner`) meaning the owner is not the owner of this staking- we should skip that address.

### Main changes
- add the `findRefreshedKeepStakes` method to the `Staking` interface and add unit test,
- update the `getOwnerStakes` implementation- call the `findRefreshedKeepStakes` to take into account new and outdated staking providers,
- update the `useFetchOwnerStakes` hook- use the `threshold-ts` lib to fetch all owner stakes.


src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,86 @@ export class Staking implements IStaking {
...
+
+    // For each staking provider, we need to verify that the current owner of
+    // the stake is indeed the one we are looking for(the `owner` argumment).
+    // It's possible that the owner can be changed multiple times. Instead of
+    // iterating through the `OwnerRefreshed` events and comparing the
+    // `oldOwner` and `newOner` params from that event, we can just check the
+    // current owner for a given staking provider by calling `rolesOf`.
+    const rolesOf = await this._multicall.aggregate(multicalls)
+
+    // Staking provideres are owned by the `owner` address.
```
michalsmiarowski, line 335: I'm not sure but I believe this comment is describing what `stakingProviders` variable is. If so the I would change it to:

```suggestion
    // Staking provideres that are owned by the `owner` address.
```

src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,86 @@ export class Staking implements IStaking {
...
+    const stakingProviders: string[] = rolesOf
+      .filter((rolesOf) => isSameETHAddress(rolesOf.owner, owner))
+      .map((_, index) => possibleStakingProviders[index])
```
michalsmiarowski, line 338: I haven't test it but I'm not sure if this would work for every case 🤔 

Let's say we have two stakes:
```
Stake 1
owner: 0x1
stakingProvider: 0x1

Stake 2
Owner: 0x2
StakingProvider: 0x11
```

Now let's say we have those `OwnerRefreshed(stakingProvider, oldOwner, newOwner)` events (order is from oldest to newest):
```
OwnerRefreshed(0x1, 0x1, 0x3)
OwnerRefreshed(0x11, 0x2, 0x3)
OwnerRefreshed(0x1, 0x3, 0x4)
```
Now we call `findRefreshedKeepStakes` function for the owner `0x3`.

`possibleStakingProviders` would equal `[0x1, 0x11]` (assuming the events `ownerRefreshedEvents` would be fetched from the oldest to newest one).

Now we check `rolesOf` for all the `possibleStakingProviders` to check what are the owners of those two. THe acutal situation is:
```
stakingProvider: 0x1
owner: 0x4

stakingProvider: 0x11
owner: 0x3
```

Now here, in the code that I marked, we defined the variable `stakingPorivers`. First we filter out the `rolesOf` to have only those whose owner is the one that we passed to this function, which in this case is `0x3`. After filter we will only have one element:

```
stakingProvider: 0x11
owner: 0x3
```

And doing the `map` function we now use the only index of the only one element in array, which is `0` and get the same index from `possibleStakingProvider`. As i mentioned before the `possibleStakingProvider` array right now looks like this:
`[0x1, 0x11]`
So it will take the first element of this array which is actually wrong. In this case the only owner of stakingProvider `0x11` is the `0x11` address not `0x1`.

So i believe the line with the `.map` might not work correctly for every scenario. I'm not sure though because I haven't test it yet in the code, only on paper 😅 Can you check if what I'm talking makes sense? Maybe I missed something down the road?

src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,86 @@ export class Staking implements IStaking {
...
+      .map((event) => ({
+        stakingProvider: event.args?.stakingProvider,
+        stakeType: event.args?.stakeType,
+      }))
+      .reverse()
+
+    const { current, outdated } = await this.findRefreshedKeepStakes(owner)
+
+    const stakingProviders = stakes
+      .filter(({ stakingProvider }) => !outdated.includes(stakingProvider))
```
michalsmiarowski, line 285: Do we actually need to have the exclamation mark (`!`) next to the `outdated` word? When I remove it I don't see any errors in my IDE.

src/threshold-ts/staking/__test__/staking.test.ts
```diff
@@ -288,4 +295,114 @@ describe("Staking test", () => {
...
+      const result = await staking.findRefreshedKeepStakes(owner)
+
+      ownerRefreshedEventEmittedExpectation()
+      rolesOfMulticallCalledExpectation(rolesOfMulticallSpyOn, [
+        stakingProvider,
+      ])
+      expect(result).toEqual({ current: [stakingProvider], outdated: [] })
+    })
+
+    test("should return outdated staking providers if the owner was updated multiple times", async () => {
```
michalsmiarowski, line 360: If may assumption in <link to comment> would be correct then maybe we could also write a test for such case

src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,86 @@ export class Staking implements IStaking {
...
+  findRefreshedKeepStakes = async (
+    owner: string
+  ): Promise<OwnerRefreshedResult> => {
+    // Find all events where the `owner` was set as a new owner or old owner of
+    // the stake.
+    const ownerRefreshedEvents = await getContractPastEvents(this._staking, {
+      eventName: "OwnerRefreshed",
+      filterParams: [null, owner, owner],
+      fromBlock: this.STAKING_CONTRACT_DEPLOYMENT_BLOCK,
+    })
```
michalsmiarowski, line 309: Doesn't this return events where both `oldOwner` AND `newOwner` is equal to `owner`? I think the docs (https://docs.ethers.org/v5/concepts/events/#events--filters) indicates that it's an `AND` condition, not `OR`.

michalsmiarowski (changes requested):
Left some comments to look at before the merge

src/threshold-ts/staking/index.ts
r-czajkowski: It's logical NOT/negation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT

src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,86 @@ export class Staking implements IStaking {
...
+      (stakingProvider) => ({
+        address: this._staking.address,
+        interface: this._staking.interface,
+        method: "rolesOf",
+        args: [stakingProvider],
+      })
+    )
+
+    // For each staking provider, we need to verify that the current owner of
+    // the stake is indeed the one we are looking for(the `owner` argumment).
```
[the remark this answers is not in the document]
r-czajkowski: 9539bfa

src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,86 @@ export class Staking implements IStaking {
...
+    // `oldOwner` and `newOner` params from that event, we can just check the
+    // current owner for a given staking provider by calling `rolesOf`.
+    const rolesOf = await this._multicall.aggregate(multicalls)
+
+    // Staking provideres are owned by the `owner` address.
+    const stakingProviders: string[] = rolesOf
+      .filter((rolesOf) => isSameETHAddress(rolesOf.owner, owner))
+      .map((_, index) => possibleStakingProviders[index])
+
+    // Outdated staking providers- the `owner` address is no longer an owner of
```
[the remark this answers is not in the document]
r-czajkowski: 9539bfa

src/threshold-ts/staking/index.ts
r-czajkowski: 9539bfa

src/threshold-ts/staking/index.ts
r-czajkowski: 6ca4ce9

src/threshold-ts/staking/index.ts
r-czajkowski: Unit test updates: ee7a182

src/threshold-ts/staking/index.ts
michalsmiarowski: Ahhh, I'm not sure why but I thought it was [non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) 😅  My bad!

src/threshold-ts/staking/index.ts
```diff
@@ -249,21 +266,109 @@ export class Staking implements IStaking {
...
+      })
+    )
+
+    // For each staking provider, we need to verify that the current owner of
+    // the stake is indeed the one we are looking for (the `owner` argument).
+    // It's possible that the owner can be changed multiple times. Instead of
+    // iterating through the `OwnerRefreshed` events and comparing the
+    // `oldOwner` and `newOwner` params from that event, we can just check the
+    // current owner for a given staking provider by calling `rolesOf`.
+    const rolesOf: { owner: string }[] = await this._multicall.aggregate(
```
michalsmiarowski, line 350: I think we can use [Role Of interface](https://github.com/threshold-network/token-dashboard/blob/main/src/threshold-ts/staking/index.ts#L40) here.

```suggestion
    const rolesOf: RolesOf[] = await this._multicall.aggregate(
```

michalsmiarowski (changes requested):
Left some more comments below.

src/threshold-ts/staking/index.ts
r-czajkowski: 2a2d714

src/threshold-ts/staking/__test__/staking.test.ts
```diff
@@ -288,4 +295,185 @@ describe("Staking test", () => {
...
+
+      const aggregateSpyOn = jest
+        .spyOn(multicall, "aggregate")
+        .mockResolvedValue([])
+
+      const result = await staking.findRefreshedKeepStakes(owner)
+
+      ownerRefreshedEventEmittedExpectation()
+      rolesOfMulticallCalledExpectation(aggregateSpyOn, [])
+      expect(result).toEqual({ current: [], outdated: [] })
```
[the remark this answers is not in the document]
r-czajkowski: 3a1a958

michalsmiarowski (approved):
LGTM 🔥

Issue #482, opened 2023-03-28
Stake details not displayed for legacy stakes


nkuba:
Stake details are not displayed for the legacy stakes. In the console, a `Block range is too wide` error is reported.

![image](https://user-images.githubusercontent.com/10741774/228166411-0ad96bda-c22a-4082-8502-42c6b6d2f7ec.png)

I was able to reproduce the problem with Taho by impersonating account `0x768536f03B64E6903c452514596886305e752589`. We already had a workaround for this issue in V1's `tbtc.js`:
https://github.com/keep-network/tbtc.js/blob/9838c06f21082cc050fab5f1304b95d6fa3db6a7/src/EthereumHelpers.js#L204-L243

pdyraga:
Worth noting, this is not only for legacy stakes. Some Taho users report they can't see their rewards or can't see their T stake.

PR #518, opened 2023-05-17
Partial events pulls mechanism

r-czajkowski:
Closes: #482 

Here we use the partial events pulls mechanism from the tbtc-v2 ts lib, see https://github.com/keep-network/tbtc-v2/pull/604. This mechanism can be useful for some Ethereum providers who limit the maximum number of blocks when querying the contract past events.

### Main changes
- fetch contract past events in batches using the `getEvents` function from `tbtc-v2.ts` lib. See https://github.com/keep-network/tbtc-v2/pull/604,
- bump the `ethers` lib to the same version used by the `tbtc-v2.ts` lib so we can use the same `Contract` type/instance. Eg. the `getEvents` helper function requires the `Contract` instance. The `Contract` type/class was updated in `v5.5.2` compared to the `v5.4.7` so the types were incompatible,
- Add a new env variable- add the new `GET_EVENTS_BLOCK_INTERVAL` env variable- defines a range of blocks using in the partial events pulls mechanism. The main idea behind passing this value as an env variable is that we can easily decrease the range and redeploy the T dapp to prod w/o changing anything in code. The default value is `10_000`- if this variable is omitted,
- bump the `tbtc-v2.ts` package to the latest one with `development` tag to be able to use this mechanism,


src/threshold-ts/utils/contract.ts
```diff
@@ -41,13 +43,22 @@ interface EventFilterOptions {
...
   contract: Contract,
   options: EventFilterOptions
 ): Promise<Array<Event>> => {
   const filter = contract.filters[options.eventName](...options.filterParams)
 
-  return await contract.queryFilter(filter, options.fromBlock, options.toBlock)
+  return await getEvents(
+    contract,
+    filter,
+    Number(options?.fromBlock ?? 0),
```
michalsmiarowski, line 58: Do we actually need the qeustion mark here next to `options`? I think we don't since `options` object is always defined. So the question mark is not needed here (see one line below where we also use `options.toBlock` without a question mark and there is no error because it will probably just use undefined value).

The question mark would only be required here if `options` argumenf for `getContractPastEvents` was optional.

michalsmiarowski, line 58: I think it's unnecessary to put all of it in `Number` constructor because if `fromBlock` is undefined then we will be left with `Number(0)` 🤔 

Maybe `Number(options.fromBlock) ?? 0`?

Based on my testing the `Number(undefined)` should give a NaN value which in this case will also be falsy and will pass the `0` to the `getEvents` value, but I would appreciate if you could test it in the code too.

michalsmiarowski (changes requested):
Overall looking good 🔥 Left some small comments to look at before the merge.

Also let's resolve the conlflicts.

r-czajkowski:
I think we can close this PR because it doesn't solve the problem. The Taho provider uses the https://www.ankr.com/rpc/eth/ as an RPC URL under the hood. The max block range in ankr is `3000`. In that case, fetching events in batches takes forever. For example the `Staked` event - the contract was deployed in `14113768` block. The current block is `18192095`. So that's 1360 requests to find the `Staked` event for a given owner address...

I took a look at the injected provider by Taho from `window` object and looks like it is not possible to override the RPC url to use eg. Alchemy. We need smarter providers like Alchemy/Infura because they work differently - from the Alchemy docs:
> You can make `eth_getLogs` requests on any block range with a cap of 10K logs in the response OR a 2K block range with no cap on logs in the response and 150MB limit on the response size

As far as I know, there are other dapps struggling with this issue. I believe this must be addressed on the Taho side

Issue #501, opened 2023-04-27
Two active subnav tabs in minting dapp


erdogan:
When I click on "How it works", both tabs are active/underlined, see sshot. Should be only the "How it works" tab. 

![image](https://user-images.githubusercontent.com/14485/234971132-8f357a83-3d73-48d1-b9a8-b4cf966ab78f.png)

PR #524, opened 2023-05-26
Fix two active navigation pills

michalsmiarowski:
Closes: #501 

There was a UI bug where it displayed two navigation pills as active instead of one. For example when clicking on `How it works` link in tbtc both `Bridge` and `How it works` links were active.

We've fixed that by adding a new method - `addActiveStatusToPills` - which adds `isActive` property to each of the link. If there are two or more links that are active (based on `useMatch` hook) then we only keep the active status for the last one.

Additionally I've also fixed the sub-navigation for `Staking` page because the `Staking` pill was not active when being on authorization page.

<details>
  <summary>Screenshot of the fix</summary>
 
image

image

</details>

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -204,12 +204,10 @@ const Routing = () => {
   )
 }
 
-const renderPageComponent = (
-  PageComponent: PageComponent,
-  /** @see PageComponent type */
-  parentPathBase: string = ""
-) => {
+const renderPageComponent = (PageComponent: PageComponent) => {
   if (!PageComponent.route.isPageEnabled) return null
+  const { parentPathBase: parentPathBaseFromRoute } = PageComponent.route
+  const parentPathBase = parentPathBaseFromRoute || ""
   const updatedParentPathBase = PageComponent.route.path
     ? `${parentPathBase}/${PageComponent.route.path}`
     : parentPathBase
@@ -231,9 +229,10 @@ const renderPageComponent = (
           />
         }
       >
-        {PageComponent.route.pages?.map((page) =>
-          renderPageComponent(page, updatedParentPathBase)
-        )}
+        {PageComponent.route.pages?.map((page) => {
+          page.route.parentPathBase = updatedParentPathBase
+          return renderPageComponent(page)
+        })}
       </Route>
     </Fragment>
   )
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -197,7 +197,7 @@ const Routing = () => {
     <Routes>
       <Route path="*" element={<Layout />}>
         <Route index element={<Navigate to="overview" />} />
-        {pages.map((page) => renderPageComponent(page))}
+        {pages.map(renderPageComponent)}
         <Route path="*" element={<Navigate to="overview" />} />
       </Route>
     </Routes>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -197,15 +197,22 @@ const Routing = () => {
     <Routes>
       <Route path="*" element={<Layout />}>
         <Route index element={<Navigate to="overview" />} />
-        {pages.map(renderPageComponent)}
+        {pages.map((page) => renderPageComponent(page))}
         <Route path="*" element={<Navigate to="overview" />} />
       </Route>
     </Routes>
   )
 }
 
-const renderPageComponent = (PageComponent: PageComponent) => {
+const renderPageComponent = (
+  PageComponent: PageComponent,
+  /** @see PageComponent type */
+  parentPathBase: string = ""
+) => {
   if (!PageComponent.route.isPageEnabled) return null
+  const updatedParentPathBase = PageComponent.route.path
+    ? `${parentPathBase}/${PageComponent.route.path}`
+    : parentPathBase
 
   return (
     <Fragment key={PageComponent.route.path}>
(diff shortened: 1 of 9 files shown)


src/components/SubNavigationPills/index.tsx
```diff
@@ -81,6 +84,38 @@ const NavPill: FC<RouteProps> = ({ path, pathOverride, title }) => {
...
+ * @return {RouteProps[]} Array of links with active status added to each of them. Only one
+ * link can have an active status.
+ */
+const addActiveStatusToPills = (pills: RouteProps[]) => {
+  const pillsWithActiveStatus: NavPill[] = []
+  let indexOfLastActivePill
+  for (let i = 0; i < pills.length; i++) {
+    const { path, pathOverride } = pills[i]
+    const resolved = useResolvedPath(pathOverride || path)
+    const match = useMatch({ path: resolved.pathname, end: true })
```
r-czajkowski, line 103: We should not call hooks in loops see https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level. We can use [`match-path`](https://reactrouter.com/en/main/utils/match-path) and [`resolvePath`](https://reactrouter.com/en/main/utils/resolve-path).

src/components/SubNavigationPills/index.tsx
```diff
@@ -11,12 +11,17 @@ import { useMatch, useResolvedPath } from "react-router-dom"
...
+interface SubNavigationPillsProps {
   links: RouteProps[]
 }
 
-const SubNavigationPills: FC<Props> = ({ links }) => {
+interface NavPill extends RouteProps {
+  isActive: boolean
+}
+
+const SubNavigationPills: FC<SubNavigationPillsProps> = ({ links }) => {
```
r-czajkowski, line 22: Can we also fix sub-navigation in staking pages?

obraz

r-czajkowski (changes requested):

r-czajkowski:
Looks like it doesn't work on Staking How it works page.

obraz

r-czajkowski (changes requested):
Looks like it doesn't work on Staking How it works page.

obraz

michalsmiarowski:
> Looks like it doesn't work on Staking How it works page.

@r-czajkowski It should be fixed now in 780e85913fdbf4f6d85ae3056bdcf901d15422d9

src/App.tsx
```diff
@@ -197,15 +197,22 @@ const Routing = () => {
...
       </Route>
     </Routes>
   )
 }
 
-const renderPageComponent = (PageComponent: PageComponent) => {
+const renderPageComponent = (
+  PageComponent: PageComponent,
+  /** @see PageComponent type */
+  parentPathBase: string = ""
```
r-czajkowski, line 210: Shouldn't we use `PageComponent["parentPathBase"]`? IMO it's more clear that it's the same type as in the page component.

src/App.tsx
```diff
@@ -197,15 +197,22 @@ const Routing = () => {
     <Routes>
       <Route path="*" element={<Layout />}>
         <Route index element={<Navigate to="overview" />} />
-        {pages.map(renderPageComponent)}
+        {pages.map((page) => renderPageComponent(page))}
```
r-czajkowski, line 200: Unnecessary change - `parentPathBase` is optional so we can still use `pages.map(renderPageComponent)`

src/components/SubNavigationPills/index.tsx
```diff
@@ -7,16 +7,33 @@ import {
...
-import { useMatch, useResolvedPath } from "react-router-dom"
+import { matchPath, resolvePath } from "react-router-dom"
 import { RouteProps } from "../../types"
 import Link from "../Link"
 
-interface Props {
+interface SubNavigationPillsProps {
   links: RouteProps[]
+  /** @see PageComponent type */
+  parentPathBase: string
```
r-czajkowski, line 17: Same as above.

src/types/page.ts
```diff
@@ -10,4 +10,11 @@ export type RouteProps = {
   isPageEnabled: boolean
 }
 
-export type PageComponent = FC<RouteProps> & { route: RouteProps }
+export type PageComponent = FC<
+  RouteProps & {
+    // Paths combined from all Route parents of the current Route
+    parentPathBase: string
```
r-czajkowski, line 16: Why is it outside of `RouteProps`? It's related to the routing so I think we can keep it in the `route` object 🤔

michalsmiarowski: I see your point but I believe it should not be in `RouteProps`. Yes, it is related to routing but the path itself has to be created in `renderPageComponent` function which uses `PageComponent` object (containing `route` property of a type `RouteProps`) to render the pages. Placing it in `RouteProps` might be misleading when defining a `route` props for a given page because we can't define it at the point of `route` property initialization. We can only create it later when we gather all routes together so we can reproduce the tree nature of the routes step by step (which is inside `renderPageComponent`). What do you think? 🤔

src/App.tsx
michalsmiarowski: After I've moved `parentPathBase` to `RouteProps` we don't have to pass the `parentPathBase` as a separate parameter now. See https://github.com/threshold-network/token-dashboard/pull/524#discussion_r1295540700

src/components/SubNavigationPills/index.tsx
michalsmiarowski: See https://github.com/threshold-network/token-dashboard/pull/524#discussion_r1295623207

src/App.tsx
michalsmiarowski: 12b40c9800bcb33be8541fe38bc88d9a427890fa

r-czajkowski (approved):

PR #531, opened 2023-06-06
Don't run postinstall script separately from `yarn install`

michalinacienciala:
Background:
Previously we've been running postinstall script in a separate step from the `yarn install` command. We've been doing that as a workaround for a problem we had with the postinstall script in the `@threshold-network/solidity-contracts` package (the script in that package often randomly failed). Running `yarn upgrade` with `--ignore-scripts` flag prevented the execution of postinstall script in the main project and its dependencies. And running `yarn run postinstall` after did execute the postinstall script in the main project.

The change:
Recently we have refactored the `@threshold-network/solidity-contracts` project and got rid of the problematic script. We can go back to executing `yarn install` without the `--ignore-scripts` flag.
We still need to execute `yarn run postinstall` in couple of places though - we need to do that after `yarn upgrade` commands, as `yarn upgrade` does not execute the postinstall scripts in the main project (and we have a script in the `threshold-network/token-dashboard` project that we need to execute).

Ref:
https://github.com/threshold-network/solidity-contracts/issues/142
https://github.com/threshold-network/solidity-contracts/pull/143
https://github.com/keep-network/tbtc-v2/pull/629

### DIFF
--- a/.github/workflows/dashboard-ci.yml
+++ b/.github/workflows/dashboard-ci.yml
@@ -42,10 +42,7 @@ jobs:
         run: git config --global url."https://".insteadOf git://
 
       - name: Install dependencies
-        run: yarn install --ignore-scripts --frozen-lockfile
-
-      - name: Run token-dashboard post-install script
-        run: yarn run postinstall
+        run: yarn install --frozen-lockfile
 
       - name: Check formatting
         run: yarn format
@@ -80,7 +77,7 @@ jobs:
             @keep-network/tbtc-v2 \
             @keep-network/tbtc-v2.ts \
             @keep-network/ecdsa \
-            @keep-network/random-beacon --ignore-scripts
+            @keep-network/random-beacon
 
       - name: Run postinstall script
         # `yarn upgrade` doesn't trigger the `postinstall` script.
--- a/.github/workflows/dashboard-mainnet.yml
+++ b/.github/workflows/dashboard-mainnet.yml
@@ -30,10 +30,7 @@ jobs:
         run: git config --global url."https://".insteadOf git://
 
       - name: Install dependencies
-        run: yarn install --ignore-scripts --frozen-lockfile
-
-      - name: Run token-dashboard post-install script
-        run: yarn run postinstall
+        run: yarn install --frozen-lockfile
 
       # FIXME: It's work in progress, the contracts are not yet published.
       # - name: Resolve latest mainnet contracts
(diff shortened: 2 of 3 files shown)


dimpar:
Why are we removing `--ignore-scripts` here? The `postinstall` script task https://github.com/threshold-network/token-dashboard/blob/main/package.json#L79 doesn't seem to be related in any way to the changes from https://github.com/threshold-network/solidity-contracts/pull/143 where we removed the `prepare-dependencies.sh` script. It looks to me that here the flag `--ignore-scripts` is set because of the `chakra` issues which are something different.
Having said that, I'm not sure if we should touch anything here.

michalinacienciala:
> Why are we removing --ignore-scripts here? The postinstall script task https://github.com/threshold-network/token-dashboard/blob/main/package.json#L79 doesn't seem to be related in any way to the changes from https://github.com/threshold-network/solidity-contracts/pull/143 where we removed the prepare-dependencies.sh script. It looks to me that here the flag --ignore-scripts is set because of the chakra issues which are something different.
Having said that, I'm not sure if we should touch anything here.

When `yarn install` is executed without `--ignore-scripts`, it by default executes the commands assigned to the `postinstall` script. It also executes all the postinstall scripts of the installed dependencies.
When a `yarn install` is run with the `--ignore-scripts` flag, neither the scripts from the main project nor from its dependencies are installed.
The `yarn run postinstall` command executes postinstall script of just the main project, not of it's dependencies.

As we have a dependency to `@threshold-network/solidity-contracts` set in `token-dashboard`, it means that we execute the postinstall script of that dependency when running `yarn install`. That was problematic when the installed `@threshold-network/solidity-contracts` had a postinstall script `prepare-dependencies.sh` that was sometimes randomly failing. That's why we decided to run install with the `--ignore-scripts`. But this would cause the `postinstall` script of the `token-dashboard` project be not executed, so we added a separate `yarn run postinstall` step. See https://github.com/threshold-network/token-dashboard/commit/3b06a289bb46ac8565b6ba8dd595e1b2bd711ba7.
Now that we removed the faulty script from the dependency and updated `yarn.lock` to use the lates `@threshold-network/solidity-contracts` package , we can go back to running installation using just `yarn install --frozen-lockfile` command.

And one more explanation, regarding the change removing `--ignore-scripts` from the `yarn upgrade...` command, but not removing the `yarn run postinstall` step afterwards: the `yarn upgrade <package>` command runs only the postinstall scripts of the upgraded packages, it does not run the postinstall script of the main project. So we are removing `--ignore-scripts` flag (as we no longer worry about executing `prepare-dependencies.sh`), but we still need to use `yarn run postinstall` to run the chakra-related script in the main project.

dimpar:
Ok, thanks for the explanation. It makes sense.

dimpar (approved):

PR #523, opened 2023-05-25
Redemption details page

r-czajkowski:
Depends on: #521

This PR adds the Redemption details page where users can track the unminting progress. After initializing the unminting process the dapp redirects users to the redemption details page `tBTC/unmint/redemption/<redemption_key>`. Here we only implement UI and this PR doesn't cover contract integration- we are going to address it in a separate PR.

<details>
  <summary>Screenshots</summary>
  
  
obraz
 
 obraz
</details>

PR #532, opened 2023-06-07
Redemption- select wallet

r-czajkowski:
~Depends on: https://github.com/keep-network/tbtc-v2/pull/630~
~Depends on: #523~

Check if there is a wallet in network that has enough BTC to handle a redemption request. If such a wallet exists we trigger the submit function and pass the wallet data to this function. Otherwise, the function from the threshold-ts lib throws an error and we set an error message for the `amount` field.


src/threshold-ts/tbtc/index.ts
```diff
@@ -629,4 +652,40 @@ export class TBTC implements ITBTC {
...
+    )
+  }
+
+  /**
+   * Returns the amount of tBTC to be minted/unminted, the remainder, and the
+   * balance to be transferred for the given mint/unmint. Note that if the
+   * `amount` is not divisible by SATOSHI_MULTIPLIER, the remainder is left on
+   * the caller's account when minting or unminting.
+   * @param {BigNumberish} amount Amount of tBTC to be converted.
+   * @return {AmountToSatoshiResult} The object that represnts convertible
```
michalsmiarowski, line 675: Typo
```suggestion
   * @return {AmountToSatoshiResult} The object that represents convertible
```

src/threshold-ts/tbtc/index.ts
```diff
@@ -629,4 +652,40 @@ export class TBTC implements ITBTC {
...
+   * amount, remainder and amount in statoshi.
```
michalsmiarowski, line 676: Typo
```suggestion
   * amount, remainder and amount in satoshi.
```

src/pages/tBTC/Bridge/Unmint.tsx
```diff
@@ -210,11 +233,27 @@ const UnmintForm = withFormik<UnmitnFormProps, UnmintFormValues>({
...
+  ) => {
+    try {
+      setSubmitting(true)
+
+      const wallet = await props.findRedemptionWallet(values.amount)
+      setFieldValue("wallet", wallet, false)
+
+      props.onSubmitForm({ ...values, wallet })
+    } catch (error) {
+      setFieldError("amount", (error as Error).message)
```
michalsmiarowski, line 253: Can we `setFieldError` to `wallet` instead of `amount` here? We could then get `errors` object in `UnmintFormBase`:

`const { isSubmitting, errors } = useFormikContext<UnmintFormValues>()`

and then print the `errors.wallet`  above the `Unmint` button.

This might be helpful because when I tried to unmint it on mocked bitcoin client I've got `Invalid Ethereum address` error that appeared below `amount` form. I was confused what is going on. I think errors that will happen in `findRedemptionWallet` should be printed just above the `Unmint` button.

michalsmiarowski (changes requested):
Left one more comment here:

src/pages/tBTC/Bridge/Unmint.tsx
r-czajkowski: 40e4397

src/threshold-ts/tbtc/index.ts
r-czajkowski: 0e3556b

src/threshold-ts/tbtc/index.ts
r-czajkowski: 0e3556b

src/threshold-ts/tbtc/index.ts
```diff
@@ -198,6 +219,8 @@ export interface ITBTC {
   ): string
 
   findAllRevealedDeposits(depositor: string): Promise<RevealedDepositEvent[]>
+
+  findWalletForRedemption(amount: BigNumberish): Promise<RedemptionWalletData>
```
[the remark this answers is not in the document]
r-czajkowski: 0e3556b

src/pages/tBTC/Bridge/Unmint.tsx
```diff
@@ -210,11 +250,32 @@ const UnmintForm = withFormik<UnmitnFormProps, UnmintFormValues>({
...
-  handleSubmit: (values, { props }) => {
-    props.onSubmitForm(values)
+  handleSubmit: async (
+    values,
+    { props, setFieldValue, setFieldError, setSubmitting }
+  ) => {
+    try {
+      setSubmitting(true)
+
+      throw new Error("eoeoeooe")
```
michalsmiarowski, line 265: This was probably put there by mistake.

michalsmiarowski (changes requested):
Left one small comment

src/pages/tBTC/Bridge/Unmint.tsx
michalsmiarowski: Resolved in 547ee4022915329430466e1b202e32cff14383a8

michalsmiarowski (approved):
LGTM 🔥

PR #556, opened 2023-06-26
Fix issue with Link Overlay in tBTC Explorer

michalsmiarowski:
We encountered an issue with the <LinkOverlay> component from chakra-ui while using it in the History table on the tBTC Explorer page. This component was intended to facilitate navigation for each row, directing the user to the corresponding etherscan link. However, we noticed an abnormal behavior specifically on real mobile devices, even though it worked correctly in the mobile view on Chrome.

On mobile, the overlay for the last item appeared to cover the entire page. As a result, any other links on the tBTC Explorer page, such as "View on Dune Analytics" or "Start Minting," unintentionally redirected the user to etherscan.

The cause of this was that the `LinkBox` component uses `position: relative` to contain the absolute positioned `LinkOverlay`.  Unfortunately `<Tr />` component, which we passed to the `as property, does not allow `position: relative`.

As a hacky workaround we are applying `transform: scale(1)` to the tr element.  See https://github.com/chakra-ui/chakra-ui/issues/3832 for more info.

To test it out run `yarn start` as usual and then go to the link from `On Your Network:` on your mobile device. I've tested it out using Chrome nad Safari on iPhone and it seems to work as expected now.

### DIFF
--- a/src/pages/tBTC/Explorer/index.tsx
+++ b/src/pages/tBTC/Explorer/index.tsx
@@ -157,6 +157,7 @@ const HistoryRow: FC<RecentDeposit> = ({ txHash, address, amount, date }) => {
       key={`latest-mints-${txHash}`}
       _odd={{ backgroundColor: "gray.50" }}
       sx={{ td: { borderBottom: "none" } }}
+      transform="scale(1)"
     >
       <Td paddingLeft={12}>
         <BodyXs>


r-czajkowski (changes requested):
Looks like it fixes an issue with links but now I can't scroll the table horizontally on mobile on FF and Safari.

michalsmiarowski:
> https://preview.dashboard.test.threshold.network/fix-issue-with-link-overlay-in-tbtc-explorer/index.html

Hmmm, I've tested on Chrome and Safarit and it works ok.

Chrome version: 114.0.5735.124 (You can check your by going into Chrome -> click three dots (`...`) in right-bottom corner -> Settings (in the top bar) -> Google Chrom)

Safari version: Probably 16.3 (Not sure how to check it, but I've for iOS 16.3.1 and it based on my really quick research if has Safari 16.3)

Here are videos showing that it works for me:

https://github.com/threshold-network/token-dashboard/assets/40306834/29c7395a-f6b8-4806-b9ad-91473b1e3c46


https://github.com/threshold-network/token-dashboard/assets/40306834/1d9673e4-cec5-4bad-9c87-41936218235c

Do you mind checking it again @r-czajkowski? And also let me know what versions of the browser did you use

michalsmiarowski:
@r-czajkowski Checked on firefox and it also works for me 🤔 Firefox version 115.1 (32781). Can you update firefox and safari and check again using preview link?

https://github.com/threshold-network/token-dashboard/assets/40306834/f5951dd9-7539-4afe-a9a5-e4670b92045e

r-czajkowski (approved):

PR #561, opened 2023-07-04
Fetch redemption details

r-czajkowski:
This PR adds support for fetching the redemption details data from the chain. Based on that data we render different states of the unminting process on the Redemption Details page.

### Add new methods to the `TBTC` interface from `threshold-ts` lib
Add methods that fetch the redemption request and redemption-related events. Here we also noticed a bug in the `ethers.js` lib- the `ethers.js` lib encodes the `bytesX` param in the wrong way. It uses the left-padded rule but based on the Solidity docs it should be a sequence of bytes in X padded with trailing zero-bytes to a length of 32 bytes(right-padded). See https://docs.soliditylang.org/en/v0.8.17/abi-spec.html#formal-specification-of-the-encoding
Consider this wallet public key hash
`0x03B74D6893AD46DFDD01B9E0E3B3385F4FCE2D1E`:
- `ethers.js` returns
  `0x00000000000000000000000003b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e`
- should be:
  `0x03b74d6893ad46dfdd01b9e0e3b3385f4fce2d1e000000000000000000000000`
  
In that case, in methods that fetch the past events by indexed param which has `bytesX` type(`RedemptionsCompleted`, `RedemptionRequested`, `RedemptionTimedOut`) we build the filter topics manually.

### Hook that fetches the redemption details
This hook fetches the redemption request details based on the:
- **redemption requested tx hash**-  We also need to find an event by
  transaction hash because it's possible that there can be multiple
  `RedemptionRequest` events with the same redemption key but created at
  different times eg:
    - redemption X requested,
    - redemption X was handled successfully and the redemption X was
      removed from `pendingRedemptions` map,
    - the same wallet is still in `live` state and can handle the
      redemption request with the same `walletPubKeyHash` and
      `redeemerOutputScript` pair,
    - now 2 `RedemptionRequested` events exist with the same redemption
      key(the same `walletPubKeyHash` and `redeemerOutputScript` pair).

      In that case, we must know exactly which redemption request we
      want to fetch.
- **wallet public key hash**- we need to find `RedemptionRequested` event by
  wallet public key hash to get all necessary data and make sure that
  the request actually happened. It's also used to build the redemption key,
- **redeemer**-  We need `redeemer` address as well to reduce the number of
  records- any user can request redemption for the same wallet.
- **redeemer output script**- we need this param to build the redemption key
  and find the Bitcoin transfer for this redeemer output script.
  
### Hook that fetches the block details such as `timestamp`
We need timestamps for a given block umber to calculate the total time it took to complete a redemption or how long a redemption request takes.

TODO(probably address in separate PRs):
- [ ] take into account fees
- [x] convert the `scriptPubKey` to Bitcoin address to display it on the success page see #https://github.com/keep-network/tbtc-v2/pull/646. Will update in a separate PR once we merge https://github.com/keep-network/tbtc-v2/pull/646.
- [x] pass all required data via url to fetch the redemption details #566 

<details>
  
  Pending Redemption:
  obraz

Completed Redemption:
obraz
</details>


src/threshold-ts/tbtc/index.ts
michalsmiarowski, line 1: Non-blocking/to discuss:

The `TBTC` class is slowly getting bigger and bigger. I believe we could split it up somehow like in `tbtc-v2.ts` lib so that we have parts that are responsible for deposits, parts that are responsible for redemptions, parts that are responsible for sweeping (in the future), generic parts etc. We shouldn't of course do that in this PR but i believe we could discuss it and address it so that it will be easier for us to implement new things to that class and find the old ones 😅 

I found some resources that might be helpful:
- https://stackoverflow.com/a/67837778 - the guy proposed to use `mixins`. I've never used them but some people in the comments say that it created some circular import references when they tried to implement this so I believe this solution might be complicated
- https://stackoverflow.com/questions/23876782/how-do-i-split-a-typescript-class-into-multiple-files - here I found some simpler solutions.

Curious what you think. If you don't think such refactor is needed then let me know. Like I said - this convo is only for discuss this. It should not be implemented in this PR.

src/threshold-ts/tbtc/index.ts
```diff
@@ -629,4 +711,212 @@ export class TBTC implements ITBTC {
...
+      blockNumber: number
+      transactionHash: string
+    }
+  ): RedemptionRequestedEvent => {
+    return {
+      amount: event.args?.requestedAmount.toString(),
+      walletPublicKeyHash: event.args?.walletPubKeyHash.toString(),
+      redeemerOutputScript: event.args?.redeemerOutputScript.toString(),
+      redeemer: event.args?.redeemer.toString(),
+      treasuryFee: event.args?.redeemer.toString(),
```
michalsmiarowski, line 815: Why do we assign `event.ergs?.redeemer` to the `reasuryFee` here? Shouldn't be `event.args?.treasuryFee.toString()`?

src/threshold-ts/tbtc/index.ts
```diff
@@ -629,4 +711,212 @@ export class TBTC implements ITBTC {
...
+      fromBlock,
+      toBlock
+    )
+
+    return logs.map((log) => ({
+      walletPublicKeyHash: log.args?.walletPubKeyHash.toString(),
+      redeemerOutputScript: log.args?.redeemerOutputScript.toString(),
+      blockNumber: log.blockNumber,
+      txHash: log.transactionHash,
+    }))
```
michalsmiarowski, line 891: I believe this is also a workaround until the fix in `ethers.js` is ready so I would add a `TODO` here too.

src/threshold-ts/tbtc/index.ts
```diff
@@ -629,4 +711,212 @@ export class TBTC implements ITBTC {
...
+    )
+
+    return logs.map((log) => ({
+      walletPublicKeyHash: log.args?.walletPubKeyHash.toString(),
+      redemptionBitcoinTxHash: Hex.from(log.args?.redemptionTxHash.toString())
+        .reverse()
+        .toString(),
+      blockNumber: log.blockNumber,
+      txHash: log.transactionHash,
+    }))
```
michalsmiarowski, line 920: Same as above

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+          requestedAt === redemptionRequestedEventTimestamp
+        ) {
+          setRedemptionData({
+            amount: redemptionRequest.amount,
+            redemptionRequestedTxHash: redemptionRequest.txHash,
+            redemptionCompletedTxHash: undefined,
+            requestedAt: requestedAt,
+            redemptionTimedOutTxHash: timedOutTxHash,
+            isTimedOut,
+          })
```
michalsmiarowski, line 135: Shouldn't we also set `btcAddress` here?

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+
+const toPrefixedRawRedeemerOutputScript = (scriptPubKey: string) => {
+  const rawRedeemerOutputScript = Buffer.from(scriptPubKey.toString(), "hex")
+
+  // Prefix the output script bytes buffer with 0x and its own length.
+  return `0x${Buffer.concat([
+    Buffer.from([rawRedeemerOutputScript.length]),
+    rawRedeemerOutputScript,
+  ]).toString("hex")}`
+}
```
michalsmiarowski, line 225: I know we use this function only in that hook, but it seems too specified to be placed here (and also it's too hard to find it IMO). Let's assume that in the future someone (and that someone is not me or you 😛 ) will need to prefix the output script with `0x` and its own length in some other place in our dApp. I doubt that he will found it here. Mybe we could move it to our `TBTC` class?

michalsmiarowski (changes requested):
Left some comments to look at before the merge. 

Overall looks great 💪 Love how earch step in the `useFetchRedemptionDetails` hook is commencted ❤️

src/pages/tBTC/Bridge/UnmintDetails.tsx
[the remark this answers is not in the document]
r-czajkowski: `Unmint` and `UnmintDetails` are page components so they are at the top. We probably should move `BridgeLayout`, `MintUnmintNav`, `TBTCBalanceCard` components to `src/pages/Bridge/components` dir.

src/threshold-ts/tbtc/index.ts
r-czajkowski: Agree, we need to take a closer look at this.

src/threshold-ts/tbtc/index.ts
```diff
@@ -629,4 +711,212 @@ export class TBTC implements ITBTC {
...
+  ): Promise<RedemptionRequestedEvent[]> => {
+    const { walletPublicKeyHash, redeemer, fromBlock, toBlock } = filter
+
+    // TODO: Use this sinppet to fetch events once we provide a fix in the `ethers.js` lib.
+    // const events = await getContractPastEvents(this.bridgeContract, {
+    //   eventName: "RedemptionRequested",
+    //   filterParams: [walletPublicKeyHash, null, redeemer],
+    //   fromBlock,
+    //   toBlock,
+    // })
```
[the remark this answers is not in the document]
r-czajkowski: I'm going to open a PR in `ethers.js` repo next week. I have fix for that just need to add tests and open a PR with an explanation of the problem.

src/hooks/tbtc/useFetchRedemptionDetails.ts
r-czajkowski: As I can see on Figma we need the Bitcoin address only on the success page so I think it's not mandatory. But up to you I can add, just let me know.

src/hooks/tbtc/useFetchRedemptionDetails.ts
r-czajkowski: febbedc

src/threshold-ts/tbtc/index.ts
r-czajkowski: 1cac4b9

src/threshold-ts/tbtc/index.ts
r-czajkowski: 1cac4b9

src/threshold-ts/tbtc/index.ts
r-czajkowski: 275749d

src/pages/tBTC/Bridge/UnmintDetails.tsx
```diff
@@ -93,104 +130,113 @@ export const UnmintDetails: PageComponent = () => {
...
+                ml="auto"
+                tokenDecimals={8}
+                withHigherPrecision
+              />
+            </BridgeProcessCardSubTitle>
+            <Timeline>
+              <Badge
+                variant="subtle"
+                size="sm"
+                bg={useColorModeValue("white", "brand.800")}
```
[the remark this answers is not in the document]
r-czajkowski: 990ea35

src/pages/tBTC/Bridge/UnmintDetails.tsx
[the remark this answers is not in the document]
r-czajkowski: 8d50c6d984e0e1f6e6a852facf3b81a34893bc08

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+    const fetch = async () => {
+      setIsFetching(true)
+      try {
+        const redemptionKey = threshold.tbtc.buildRedemptionKey(
+          walletPublicKeyHash,
+          redeemerOutputScript
+        )
+
+        // We need to find `RedemptionRequested` event by wallet public key hash
+        // and `reedemer` address to get all necessary data and make sure that
```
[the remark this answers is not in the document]
r-czajkowski: a7e1384

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+        // the request actually happened. We need `redeemer` address as well to
+        // reduce the number of records- any user can request redemption for the
+        // same wallet.
```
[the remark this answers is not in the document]
r-czajkowski: a7e1384

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+        const redemptionRequest = (
+          await threshold.tbtc.getRedemptionRequestedEvents({
+            walletPublicKeyHash,
+            redeemer,
+          })
+        ).find(
+          (event) =>
+            // It's not possible that the redemption request with the same
+            // redemption key can be created in the same transaction- it means
```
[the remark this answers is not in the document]
r-czajkowski: a7e1384

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+              (event) => event.redeemerOutputScript === redeemerOutputScript
+            )?.txHash
+          : undefined
+
+        if (
+          (isTimedOut || isPending) &&
+          // We need to make sure this is the same redemption request. Let's
+          // consider this case:
+          // - redemption X requested,
+          // - redemption X was handled sucesfully and the redemption X was
```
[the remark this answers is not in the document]
r-czajkowski: a7e1384

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+            redemptionRequestedTxHash: redemptionRequest.txHash,
+            redemptionCompletedTxHash: undefined,
+            requestedAt: requestedAt,
+            redemptionTimedOutTxHash: timedOutTxHash,
+            isTimedOut,
+          })
+          return
+        }
+
+        // If we are here it menas that the redemption request was handled
```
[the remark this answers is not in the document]
r-czajkowski: a7e1384

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,225 @@
...
+        // successfully and we need to find all `RedemptionCompleted` events
+        // that happend after `redemptionRequest` block and filter by
```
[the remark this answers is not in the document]
r-czajkowski: a7e1384

src/hooks/tbtc/useFetchRedemptionDetails.ts
michalsmiarowski: No, not needed. I was juts afraid of a situation where we could have some old `btcAddress` saved in state bu since we set the data as a whole object, and replace the whole `RedemptionDetails` then it looks like it will neeve be such situation and this `btcAddress` will be undefined in this case.

src/hooks/tbtc/useFetchRedemptionDetails.ts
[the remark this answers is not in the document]
r-czajkowski: 9776847

src/threshold-ts/tbtc/index.ts
```diff
@@ -198,6 +251,35 @@ export interface ITBTC {
...
+
+  getRedemptionRequest(redemptionKey: string): Promise<RedemptionRequest>
+
+  getRedemptionTimedOutEvents(
+    filter: RedemptionTimedOutEventFilter
+  ): Promise<RedemptionTimedOutEvent[]>
+
+  getRedemptionsCompletedEvents(
+    filter: RedemptionsCompletedEventFilter
+  ): Promise<RedemptionsCompletedEvent[]>
```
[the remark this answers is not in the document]
r-czajkowski: ed3e3c4

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -0,0 +1,216 @@
...
+              redemptionRequestedTxHash: redemptionRequest.txHash,
+              redemptionCompletedTxHash: {
+                chain: txHash,
+                bitcoin: redemptionBitcoinTxHash,
+              },
+              requestedAt: redemptionRequestedEventTimestamp,
+              completedAt: redemptionCompletedTimestamp,
+              isTimedOut: false,
+              // TODO: convert the `scriptPubKey` to address.
+              btcAddress: "2Mzs2YNphdHmBoE7SE77cGB57JBXveNGtae",
```
michalsmiarowski, line 184: Let's not forget about this TODO

michalsmiarowski (changes requested):
Left one comment to point out the `TODO` but overall looking good. I'm ready to merge it

src/hooks/tbtc/useFetchRedemptionDetails.ts
r-czajkowski: We need https://github.com/keep-network/tbtc-v2/pull/646 first. Once we merge this PR I'll update on the dapp side.

michalsmiarowski (approved):
LGTM 🔥

src/hooks/tbtc/useFetchRedemptionDetails.ts
r-czajkowski: Resolved in #568

PR #566, opened 2023-07-10
Get redemption params from url to fetch details

r-czajkowski:
Depends on: #561 

Get the redemption params from url to fetch the redemption details on the Redemption details page. Also add validation in the `useFetchRedemptionDetails` hook to make sure all params are in correct format.


src/components/Modal/tBTC/InitiateUnminting.tsx
```diff
@@ -41,6 +41,8 @@ const InitiateUnmintingBase: FC<InitiateUnmintingProps> = ({
   // TODO: implement submit function
   const initiateUnminting = () => {
     // TODO: It's a temporary solution to be able to go through the whole flow.
+    // TODO: Update in
+    // https://github.com/threshold-network/token-dashboard/pull/537
```
michalsmiarowski, line 45: Let's remove that (although it will probably be removed when you pull the changes from `main` since https://github.com/threshold-network/token-dashboard/pull/537 is already merged.

src/hooks/tbtc/useFetchRedemptionDetails.ts
```diff
@@ -17,11 +19,13 @@ interface RedemptionDetails {
...
+
 export const useFetchRedemptionDetails = (
-  redemptionRequestedTxHash: string,
-  walletPublicKeyHash: string,
-  redeemerOutputScript: string,
-  redeemer: string
+  redemptionRequestedTxHash: FetchRedemptionDetailsParamType,
+  walletPublicKeyHash: FetchRedemptionDetailsParamType,
+  redeemerOutputScript: FetchRedemptionDetailsParamType,
+  redeemer: FetchRedemptionDetailsParamType
```
michalsmiarowski, line 28: Can't we just keep the types as string and make all the propeties optional?

```suggestion
  redemptionRequestedTxHash?: string,
  walletPublicKeyHash?: string,
  redeemerOutputScript?: string,
  redeemer?: string
```

r-czajkowski: I decided to add `null` and create a custom type because `searchParams.get` from [here](https://github.com/threshold-network/token-dashboard/pull/566/files#diff-c0362ebe68a4f572dd5a124a7d56f3abc9086c667f2513883bcbe783181596beR74) returns `string | null` but optional param is meaning "type X or **undefined**" so I got a ts error w/o `null`.

src/components/Modal/tBTC/InitiateUnminting.tsx
r-czajkowski: I'll remove once we merge #561 and then merge the main branch into `get-redemption-params-from-url`.

src/utils/tBTC.ts
```diff
@@ -85,7 +86,7 @@ export const buildRedemptionDetailsLink = (
 ): To => {
   const queryParams = new URLSearchParams()
   queryParams.set("redeemer", redeemer)
-  queryParams.set("walletPublicKey", walletPublicKey)
+  queryParams.set("walletPublicKeyHash", `0x${computeHash160(walletPublicKey)}`)
   queryParams.set("redeemerOutputScript", redeemerOutputScript)
```
michalsmiarowski, line 90: When doing a redemption I've got `Error: Redemption not found...`. I think there is a mistake of how we create `redeemerOutputScript` here:
https://github.com/threshold-network/token-dashboard/blob/5fd3f64eeb555fde9791939a47ef419f055000b2/src/components/Modal/tBTC/InitiateUnminting.tsx#L62

The `redeemerOutputScript` itself is correct but we only add `0x` prefix at the beginning and I velieve we should also prefix it with a byte length. For example for address `2NBDAoFXUvdFZhvaQ5DxZjKCRHdMBnjRBTq` the `redeemerOutputScript` from this call is `0xa914c50ec4e54c362b7bd64ce136e967dd58fe907ef787`.

The problem is that here:
https://github.com/threshold-network/token-dashboard/blob/5fd3f64eeb555fde9791939a47ef419f055000b2/src/hooks/tbtc/useFetchRedemptionDetails.ts#L101-L104
the redemption keys comparison fails, because `event.redeemerOutputScript` is also prefixed with the byte length so the actual values of it is `0x17a914c50ec4e54c362b7bd64ce136e967dd58fe907ef787` (notice the `17` after the `0x`).

michalsmiarowski: Similarly this piece of code also doesn't work properly:
https://github.com/threshold-network/token-dashboard/blob/5fd3f64eeb555fde9791939a47ef419f055000b2/src/hooks/tbtc/useFetchRedemptionDetails.ts#L116-L122

because we pass the `redeemerOutputScript` that is not prefixed with length. After using `redemptionRequest.redeemerOutputScript` (which is prefixed with `0x` and length) it correclty finds the redemption request.

michalsmiarowski (changes requested):
There is an error that I found with `redeemerOutputScript`. Please see the comments below:

src/utils/tBTC.ts
r-czajkowski: 81875f9

michalsmiarowski (approved):
LGTM 🔥

PR #568, opened 2023-07-12
Convert script pub key to Bitcoin address

r-czajkowski:
Depends on: #566 

Convert the `scriptPubKey` placed on the output of a Bitcoin
transaction to a Bitcoin address. We want to display the Bitcoin
address in UI instead of the output script.


michalsmiarowski (approved):
LGTM 🔥

PR #572, opened 2023-07-13
Render empty state on the Unmint page

r-czajkowski:
Render an empty state on the Unmint page. Here we also extract empty state component to a separate file since we want to render the same empty state for unmint and mint page.

<details>
  
obraz
  
</details>

### DIFF
--- a/src/pages/tBTC/Bridge/Unmint.tsx
+++ b/src/pages/tBTC/Bridge/Unmint.tsx
@@ -288,7 +288,7 @@ UnmintFormPage.route = {
   isPageEnabled: true,
 }
 
-export const UnmintPage: PageComponent = ({}) => {
+export const UnmintPageLayout: PageComponent = ({}) => {
   const { active } = useWeb3React()
 
   return (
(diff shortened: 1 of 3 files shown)


src/pages/tBTC/Bridge/Unmint.tsx
```diff
@@ -286,7 +289,22 @@ UnmintFormPage.route = {
...
-  return <Outlet />
+  const { active } = useWeb3React()
+
+  return (
+    <BridgeLayout>
+      <BridgeLayoutMainSection>
+        {active ? (
+          <Outlet />
+        ) : (
+          <BridgeProcessEmptyState title="Ready to unmit tBTC?" />
```
michalsmiarowski, line 300: After connecting the wallet while being on redemption page i see something like this:

image

I believe this is not expected behaviour

michalsmiarowski (changes requested):
Found bug in UI, see comment below

src/pages/tBTC/Bridge/Unmint.tsx
r-czajkowski: f35c091

michalsmiarowski (approved):
LGTM 🖖

PR #598, opened 2023-08-21
Fix WalletConnection alert

michalsmiarowski:
There was a bug that caused the walletconnection alert to not show up the second time if the error message stays the same.

This commit fixes that.

image

To see if it works you can:
1. Click "Connect  Wallet"
2. Choose "WalletConnect"
3. Close the WalletConnect modal with `X` button
4. The "Connection request reset. Please try again" alert should appear in the top right corner.
5. Close the alert
6. Repeat steps 1-3
7. The "Connection request reset. Please try again" alert should appear again

### DIFF
--- a/src/components/Navbar/WalletConnectionAlert.tsx
+++ b/src/components/Navbar/WalletConnectionAlert.tsx
@@ -15,7 +15,7 @@ const WalletConnectionAlert: FC<{
   chainId?: number
 }> = ({ account, chainId }) => {
   const [hideAlert, setHideAlert] = useState(false)
-  const { error } = useWeb3React()
+  const { error, deactivate } = useWeb3React()
   const [alertDescription, setAlertDescription] = useState("")
 
   const errorMessage = error?.message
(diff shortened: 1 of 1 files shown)


src/components/Navbar/WalletConnectionAlert.tsx
```diff
@@ -46,6 +46,7 @@ const WalletConnectionAlert: FC<{
   const resetAlert = () => {
     setHideAlert(true)
     setAlertDescription("")
+    if (error) error.message = ""
```
r-czajkowski, line 49: We should never change or update the state directly. `error` is a state variable see https://github.com/Uniswap/web3-react/blob/v6/packages/core/src/manager.ts#L117. We should probably use `setError` eg:
```ts
const { error, setError } = useWeb3React()
...
setError(null)
```

or we should call `deactivate` fn from `useWeb3React` hook to clear the state (see https://github.com/Uniswap/web3-react/blob/v6/packages/core/src/manager.ts#L168 and https://github.com/Uniswap/web3-react/blob/v6/packages/core/src/manager.ts#L90)

r-czajkowski (changes requested):

src/components/Navbar/WalletConnectionAlert.tsx
michalsmiarowski: Great catch! Fixed in 779f5ae0daa5fb4548469288df986558db7524a8

r-czajkowski (approved):

PR #600, opened 2023-08-22
Add a possibility to reveal multiple transactions

michalsmiarowski:
Add a possibility to reveal multiple transactions send to the same BTC deposit address. In such case the user should use `/tBTC/mint/continue` url to initiate the minting of each subsequent transaction after he revealed the first one.

### DIFF
--- a/src/store/tbtc/effects.ts
+++ b/src/store/tbtc/effects.ts
@@ -78,20 +78,31 @@ export const findUtxoEffect = async (
           continue
         }
 
-        // UTXOs returned from `findAllUnspentTransactionOutputs` are in
-        // reversed order so we have to get the last element of the `utxos`
-        // array to get the oldest utxo related to this deposit address.
-        const utxo = utxos.pop()!
+        let utxo = utxos[0]
+        let areAllDepositRevealed = true
 
-        // Check if deposit is revealed.
-        const deposit = await forkApi.pause(
-          listenerApi.extra.threshold.tbtc.getRevealedDeposit(utxo)
-        )
+        // We have to find the first utxo that is revealed. The UTXOs returned
+        // from `findAllUnspentTransactionOutputs` are in reversed order so
+        // that's why we start our search from the last elements of the `utxos`.
+        // If all deposits are revealed then we just use the first utxo (which
+        // should be the most recent transaction)
+        for (let i = utxos.length - 1; i >= 0; i--) {
+          // Check if deposit is revealed.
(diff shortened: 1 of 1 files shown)


src/store/tbtc/effects.ts
```diff
@@ -78,20 +78,31 @@ export const findUtxoEffect = async (
...
-          listenerApi.extra.threshold.tbtc.getRevealedDeposit(utxo)
-        )
+        let utxo = utxos[0]
+        let areAllDepositRevealed = true
+
+        // We have to find the first utxo that is revealed. The UTXOs returned
+        // from `findAllUnspentTransactionOutputs` are in reversed order so
+        // that's why we start our search from the last elements of the `utxos`.
+        // If all deposits are revealed then we just use the first utxo (which
+        // should be the most recent transaction)
```
r-czajkowski, line 88:
```suggestion
        // should be the most recent transaction).
```

src/store/tbtc/effects.ts
```diff
@@ -78,20 +78,31 @@ export const findUtxoEffect = async (
...
-        const utxo = utxos.pop()!
-
-        // Check if deposit is revealed.
-        const deposit = await forkApi.pause(
-          listenerApi.extra.threshold.tbtc.getRevealedDeposit(utxo)
-        )
+        let utxo = utxos[0]
+        let areAllDepositRevealed = true
+
+        // We have to find the first utxo that is revealed. The UTXOs returned
```
r-czajkowski, line 84:
```suggestion
        // We have to find the first utxo that is not revealed. The UTXOs returned
```

Here we are looking for the first not revealed utxo. If we find it we break the loop and redirect to step 3. So I think it should be `the first utxo that is not revealed.` 🤔

r-czajkowski (changes requested):
Left two minor comments. Overall LGTM.

src/store/tbtc/effects.ts
michalsmiarowski: Fixed in 30d657fc7225f883070330a10f4697567c5a9be9. I changed the comment a little bit so it describes better what's going on there (at least I think it's better, but if you don't agree then let me know in the review 😛 )

src/store/tbtc/effects.ts
michalsmiarowski: Fixed in 30d657fc7225f883070330a10f4697567c5a9be9

r-czajkowski (approved):

PR #602, opened 2023-09-01
chore: Add .env to .gitignore and add example env

theref:
`.env` should not be versioned controlled. Instead have an example env file


beaurancourt (approved):

therealsphinx (approved):
Approved

.github/workflows/reusable-build-and-publish.yml
```diff
@@ -106,7 +106,7 @@ jobs:
 
       - name: Run token-dashboard post-install script
         shell: bash
-        run: yarn run postinstall
+        run: yarn run postinstall ; cp .env.example .env
```
beaurancourt, line 109: What is this for?

nkuba (approved):

theref:
fixed by #625

PR #613, opened 2023-09-15
Coverage Pools balance

michalsmiarowski:
Closes: https://github.com/threshold-network/website/issues/88

Implements the Coverage Pools balance in the dashboard.

Note: In the project ([on release branch](https://github.com/threshold-network/token-dashboard/blob/releases/mainnet/v1.10.0/package.json#L14)) we are using `v1.0.0` version of `@keep-network/coverage-pools` which has contracts related to Keep. The T Coverage Pools are in `v2.0.0` version. Since we still use `v1.0.0` version in our dApp to display it in the TVL I've decided to isntall `v2.0.0` version next to the first one anc call it `@threshold-network/coverage-pools`.

On the `main` branch they both have `development` tag but on the release one we should set the correct version accordingly.


michalsmiarowski:
@r-czajkowski I've also created another PR related to this one with some small code cleaning - #614

src/web3/hooks/useKeepAssetPoolContract.ts
```diff
@@ -0,0 +1,6 @@
+import { useContract } from "./useContract"
+import AssetPool from "@threshold-network/coverage-pools/artifacts/AssetPool.json"
+
+export const useKeepAssetPoolContract = () => {
+  return useContract(AssetPool.address, AssetPool.abi)
+}
```
r-czajkowski, line 6: Hmmm looks like we should import `AssetPool` artifact from `"@keep-network/coverage-pools/artifacts/AssetPool.json"` but in the `useAssetPoolContract` hook from `"@threshold-network/coverage-pools/artifacts/AssetPool.json"` 🤔

src/components/StatHighlightCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+  )
+}
+
+export const StatHighlightTitle: FC<{ title: string } & TextProps> = ({
+  title,
+  children,
+  ...restProps
+}) => {
+  return (
+    <HStack>
```
r-czajkowski, line 28: IMO it's unnecessary and always forces the title to be inside a flex container. We can use `HStack` in the parent component if needed.

src/components/StatHighlightCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+    </H1>
+  )
+}
+
+export const StatHighlightLink: FC<{ href: string & TextProps }> = ({
+  href,
+  ...restProps
+}) => {
+  return (
+    <H1 mt="10" mb="9" textAlign="center" {...restProps}>
```
r-czajkowski, line 70: Shouldn't we add `as={Link}` to render the link as `a` tag?

r-czajkowski (changes requested):

src/components/StatHighlightCard.tsx
r-czajkowski: It also seems unnecessary- we do not use it anywhere.

r-czajkowski (changes requested):

src/components/StatHighlightCard.tsx
michalsmiarowski: I've pushed this component by mistake. I've removed it in 203a931ed0a9e97c7d74060620519546d1430b4e

src/components/StatHighlightCard.tsx
michalsmiarowski: Hmm I'm not sure about that, because the `<HStack>` is only in `StatHighlightTitle`. Any children here should be related to the title and title in this card is usually one-horizonal-liner thus this `<HStack>` here.

If the user would need to add something below/above title he could just create that outside of the `<Stat Highlight Title> component. What do you think about that approach? @r-czajkowski

src/web3/hooks/useKeepAssetPoolContract.ts
michalsmiarowski: 772aaf55ff25f6e59b0973472e6eef41e05d92fb

src/pages/tBTC/CoveragePools/CoveragePoolsTVLCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+    </>
+  )
+  return (
+    <StatHighlightCard {...restProps}>
+      <StatHighlightTitle title={"coverage pool tvl"}>
+        <StatHighlightTitleTooltip label={tooltipLabel} />
+      </StatHighlightTitle>
+      <StatHighlightValue value={formattedTVL} />
+      <Center>
+        <LearnMoreLink m={"0 auto"}>
```
r-czajkowski, line 48: We can use `mx="auto"` - I believe this will improve readability.

src/hooks/useFetchTvl.ts
r-czajkowski, line 1: Let's update unit tests for this hook.

r-czajkowski (changes requested):

r-czajkowski (approved):

r-czajkowski (approved):

PR #624, opened 2023-09-22
Add docker file, docker-compose, and env changes

theref:
This is an attempt to simplify the dev process for newcomers. To test, follow the instructions in the README regarding updating `package.json` and `.env`, then: `docker-compose run --build`. First build will take a while, but after that should be very quick.

Example logs can be seen here https://app.warp.dev/block/qnf8YCpPJIuxvSQRwnRJ1i


beaurancourt:
Hey @theref, thanks for opening this up!

Can you include in the PR description some context (what the change is and why you're making it), testing procedures, and a screenshot of docker running beautifully in the console? 

It doesn't have to be too in-depth, but it's nice for our future selves to have the record :D

README.md
```diff
@@ -114,3 +114,50 @@ The following procedure allows to deploy T token dashboard to production:
...
+    "@keep-network/keep-ecdsa": "goerli",
+    "@keep-network/random-beacon": "goerli",
+    "@keep-network/tbtc": "goerli",
+    "@keep-network/tbtc-v2": "goerli",
+    "@keep-network/tbtc-v2.ts": "development",
+    "@threshold-network/components": "development",
+    "@threshold-network/solidity-contracts": "goerli",
+```
+
+Update `.env` to contain:
```
manumonti, line 135: What about having a new .env file called `.env.localdev` or something like that? The instructions could be something like "cp `.env.localdev` `.env`" and fill the `<your API key here>` placeholders.

theref: `.env.example` already exists for this, but the values are slightly wrong which is why i put them here

README.md
```diff
@@ -114,3 +114,50 @@ The following procedure allows to deploy T token dashboard to production:
...
+REACT_APP_ELECTRUM_PORT=8443
+REACT_APP_MOCK_BITCOIN_CLIENT=false
+
+REACT_APP_WALLET_CONNECT_PROJECT_ID=$WALLET_CONNECT_PROJECT_ID
+```
+
+Then build the docker container and run the dashboard:
+
+```bash
+docker-compose run --build
```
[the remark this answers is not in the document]
theref: wrong command, have updated

manumonti (dismissed):
LGTM 🙌

derekpierre:
@theref , looks good.

Couple questions:
1. Does it make sense to tuck the docker-compose.yml and Dockerfile.dev files into a `docker` folder or some other named folder instead of having them directly in the base directory? If so, may need to update paths used in config files.
2. Is it helpful to sneak `docker-compose run --build` or `docker-compose run -f ./docker/docker-compose.yml run build` (if you move the files from 1. above) into a yarn script command in the `package.json`, so then the user runs `yarn docker-run` or something like that? This is meant to be phrased as a question, not a suggestion - I'm less familiar with yarn best practices.

theref:
This PR is no longer ready as the build is no longer successful. There seems to have been a lot of changes in upstream packages that are causing issues here. I am attempting to narrow them down. As I make discoveries I will continue to post them here.

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    container_name: threshold-dashboard
+    working_dir: /app
+    environment:
+      - PYTHON=/usr/bin/python3
+      - NODE_OPTIONS=--max_old_space_size=3072
+    ports:
+      - "3000:3000"
+    volumes:
+      - .:/app # Bind mount the current directory to /app in the container
+      - /app/node_modules # This will prevent node_modules from being overwritten by the local volume
```
piotr-roslaniec, line 15: Adding `node_modules` to `.dockerignore` instead would also have an additional benefit of a smaller build context. (Even thought we don't `COPY` it, `docker` still has to index it before starting a build process).

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: "3.8"
+
+services:
+  threshold-dashboard:
+    image: node:14-buster-slim
+    container_name: threshold-dashboard
+    working_dir: /app
+    environment:
+      - PYTHON=/usr/bin/python3
+      - NODE_OPTIONS=--max_old_space_size=3072
```
piotr-roslaniec, line 10: We've already set these up in `Dockerfile.dev`. Any reason why we set these in two different places?

Dockerfile.dev
```diff
@@ -0,0 +1,22 @@
...
+
+# Install dependencies
+RUN apt-get update && apt-get install -y python3 make g++ git openssh-client ca-certificates && \
+    git config --global url."https://".insteadOf git:// && \
+    rm -rf /var/lib/apt/lists/* && \
+    apt-get clean
+
+# Set the environment variables
+ENV PYTHON=/usr/bin/python3
+ENV NODE_OPTIONS=--max_old_space_size=3072
```
piotr-roslaniec, line 15: Could we document the motivation for using this Node.js flag?

Dockerfile.dev
```diff
@@ -0,0 +1,22 @@
+# Use the specified image
+FROM node:14-buster-slim
```
piotr-roslaniec, line 2: Node.js 14 has already reached the end-of-life, and [Node.js 16 is going to reach EOL shortly](https://github.com/nodejs/Release#release-schedule).

I suggest using `node:16-buster-slim` since it looks from the contents of `package.json`
```
"@types/node": "^16.9.1",
```

```suggestion
FROM node:16-buster-slim
```

We can address transition from 16 to 18 (the next LTS version) in another PR.

piotr-roslaniec:
You may want to consider using multistage builds to slightly speed up the building process. Not sure if this is helpful here, just dropping a hint:

```
# The first stage is for installing dependencies
FROM node:14-buster-slim  AS build-stage

WORKDIR /app

RUN apt-get update && apt-get install -y python3 make g++ git openssh-client ca-certificates && \
    git config --global url."https://".insteadOf git:// && \
    rm -rf /var/lib/apt/lists/* && \
    apt-get clean

ENV PYTHON=/usr/bin/python3
ENV NODE_OPTIONS=--max_old_space_size=3072

COPY package*.json yarn.lock ./

RUN yarn install

# Start a new stag
FROM node:14-buster-slim

WORKDIR /app

# Copy needed files from the previous stage
# This works like a cache, effectively
COPY --from=build-stage /app ./

# Expose port 3000
EXPOSE 3000
```

theref:
replaced by #625

PR #625, opened 2023-09-22
Add Taco application to dashboard

### LINKED ISSUE #678: TACo specific Operator Bonding
Currently mapping an operator is tightly wrapped around the idea of TBTC + Random Beacon, and so currently, when you set the operator, you set it for all 3 apps (TACo, TBTC, RB)

image

A new card needs adding in `src/components/Modal/MapOperatorToStakingProviderModal/index.tsx` but there are many more changes required underneath too

theref:
Builds on top of https://github.com/threshold-network/token-dashboard/pull/624

Fixes #678 #680 #675 #673


therealsphinx (approved):

manumonti (changes requested):
This is a great work 🙌

What is the plan with the TODOs created in commit #7eb5030? Are you planning to address them in other PR?

src/pages/Staking/HowItWorks/StakingApplications/index.tsx
```diff
@@ -16,6 +16,8 @@ import randomBeaconAppIllustrationLight from "../../../../static/images/randomBe
 import randomBeaconAppIllustrationDark from "../../../../static/images/randomBeaconAppIllustrationDark.png"
 import preAppIllustrationLight from "../../../../static/images/preAppIllustrationLight.png"
 import preAppIllustrationDark from "../../../../static/images/preAppIllustrationDark.png"
+// import tacoAppIllustrationLight from "../../../../static/images/tacoAppIllustrationLight.png"
```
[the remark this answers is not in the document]
theref: no it's a reminder that we might end up with new images

src/enums/externalHref.ts
```diff
@@ -3,8 +3,8 @@ export enum ExternalHref {
   thresholdDiscord = "https://discord.gg/WXK9PC6SRF",
   metamaskHomePage = "https://metamask.io/",
   stakingContractLeanMore = "https://github.com/threshold-network/solidity-contracts/issues/53",
-  preNodeSetup = "https://docs.nucypher.com/en/latest/pre_application/running_a_node.html",
-  preStakingProvidersList = "https://docs.nucypher.com/en/latest/pre_application/node_providers.html",
+  tacoNodeSetup = "https://docs.nucypher.com/en/latest/pre_application/running_a_node.html", // TODO THEREF
+  tacoStakingProvidersList = "https://docs.nucypher.com/en/latest/pre_application/node_providers.html", // TODO THEREF
```
derekpierre, line 7: Can we point to docs.threshold.network instead? That information should be available now.

src/pages/Staking/HowItWorks/StakingApplications/index.tsx
```diff
@@ -144,6 +150,28 @@ const StakingApplications: PageComponent = () => {
...
+              <ButtonLink to="/staking" isFullWidth>
+                Authorize TACO
+              </ButtonLink>
+              <ButtonLink
+                isExternal
+                href={ExternalHref.tacoNodeDocs}
+                isFullWidth
+                variant="outline"
+              >
+                TACO Node Docs
```
derekpierre, line 169:
```suggestion
                TACo Node Docs
```

src/types/staking.ts
```diff
@@ -27,6 +27,7 @@ export interface UseStakingState {
 }
 
 export interface PreConfig {
+  // TODO THEREF
```
derekpierre, line 30: Do we need to address this before merging, or is there a follow-up?

.env.production
```diff
@@ -19,3 +19,4 @@ REACT_APP_ELECTRUM_PORT=$ELECTRUM_PORT
 REACT_APP_MOCK_BITCOIN_CLIENT=false
 
 REACT_APP_WALLET_CONNECT_PROJECT_ID=$WALLET_CONNECT_PROJECT_ID
+REACT_APP_TACO_NETWORK=mainnet
```
derekpierre, line 22: Can we rename these to use "DOMAIN" instead of "NETWORK" i.e. `REACT_APP_TACO_DOMAIN`?

src/components/StakingApplicationOperationIcon/index.tsx
```diff
@@ -17,6 +17,10 @@ const iconMap: Record<StakingAppName, Record<Operation, string>> = {
     increase: randomBeaconIncrease,
     decrease: randomBeaconDecrease,
   },
+  taco: {
+    increase: tbtcIncrease,
+    decrease: tbtcDecrease,
```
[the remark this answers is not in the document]
theref: yeah there are a few images that will need updating, Manu commented on another above.

src/enums/externalHref.ts
derekpierre: These are probably sufficient and already live:
- setup: https://docs.threshold.network/staking-and-running-a-node/running-a-node/self-managed/taco-node-setup
- providers: https://docs.threshold.network/staking-and-running-a-node/running-a-node/staking-providers

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  getStakingAppLabelFromAppAddress,
+  getStakingAppLabelFromAppName,
+  getStakingAppNameFromAppAddress,
+} from "../getStakingAppLabel"
+import { StakingAppName } from "../../store/staking-applications"
+import { getThresholdLib } from "../getThresholdLib"
+
+const MOCK_ADDRESSES: Record<StakingAppName, string> = {
+  tbtc: getThresholdLib().multiAppStaking.ecdsa.address,
+  randomBeacon: getThresholdLib().multiAppStaking.randomBeacon.address,
```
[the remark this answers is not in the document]
derekpierre: @theref , What's the resolution here?

.env.example
```diff
@@ -18,3 +18,5 @@ REACT_APP_ELECTRUM_PORT=$ELECTRUM_PORT
 REACT_APP_MOCK_BITCOIN_CLIENT=true
 
 REACT_APP_WALLET_CONNECT_PROJECT_ID=$WALLET_CONNECT_PROJECT_ID
+
+REACT_APP_TACO_DOMAIN=dashboard
```
derekpierre, line 22: 🚀

.env.production
```diff
@@ -19,3 +19,4 @@ REACT_APP_ELECTRUM_PORT=$ELECTRUM_PORT
 REACT_APP_MOCK_BITCOIN_CLIENT=false
 
 REACT_APP_WALLET_CONNECT_PROJECT_ID=$WALLET_CONNECT_PROJECT_ID
+REACT_APP_TACO_DOMAIN=mainnet
```
derekpierre, line 22: 🚀

src/utils/__tests__/getStakingAppLabel.test.ts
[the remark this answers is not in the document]
derekpierre: nvm. I see it now 😅

theref:
ok, current bugs:
TACo app authorization not showing correctly

Screenshot 2023-10-18 at 10 23 53
Screenshot 2023-10-18 at 10 25 26
Screenshot 2023-10-18 at 10 25 44

Staking provider is showing an authorized stake of 40K on the contract, but that's not displaying properly in a couple of places

theref:
On the production app, we have this box:
Screenshot 2023-10-18 at 10 33 28
But this isn't showing on my dev version and I can't find anywhere to go through the Operator mapping process

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    working_dir: /app
+    environment:
+      - PYTHON=/usr/bin/python3
+      - NODE_OPTIONS=--max_old_space_size=3072
+    ports:
+      - "3000:3000"
+    volumes:
+      - .:/app # Bind mount the current directory to /app in the container
+      - /app/node_modules # This will prevent node_modules from being overwritten by the local volume
+    command: bash -c "yarn format:fix && yarn start"
```
cygnusv, line 15: why the `yarn format:fix` part?

theref: otherwise it refuses to start if the code isn't correctly formatted

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -36,32 +36,43 @@ import ModalCloseButton from "../ModalCloseButton"
...
       openModal(ModalType.MapOperatorToStakingProviderConfirmation, {
         operator,
-        isOperatorMappedOnlyInTbtc,
-        isOperatorMappedOnlyInRandomBeacon,
+        appName,
+        mappedOperatorTbtc,
+        mappedOperatorRandomBeacon,
+        mappedOperatorTaco,
       })
     }
```
michalsmiarowski, line 77: I think opening modal based on `appName` might be little bit problematic if we want to use the same confirmation modal for all the apps (and I assume we do). The `appName` indicates that we would need to open separate modal for each app 🤔 

Instead of that I would pass two operators, one for tbtc + random beacon and a second one for taco. Based on that, and on the rest parameters, we could display the proper information in the confirmation modal. For example if user input only operator for tbtc + random beacon we will pass the value here, and the operator value for taco as undefined. Remember that there are cases where users can have tbtc operator mapped without random beacon. Such cases can be covered here too.

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -36,32 +36,43 @@ import ModalCloseButton from "../ModalCloseButton"
...
 }
 
 const MapOperatorToStakingProviderModal: FC<
   BaseModalProps & MapOperatorToStakingProviderModalProps
 > = () => {
   const { account } = useWeb3React()
-  const formRef =
+  const formRefTbtc =
+    useRef<FormikProps<MapOperatorToStakingProviderFormValues>>(null)
+  const formRefTaco =
```
michalsmiarowski, line 48: `formRefTbtc` should have a name that also covers random beacon, because it might get confused. I would propose to come up with some better name for that 🤔 Either use both app names in such cases or maybe just mention them as `keepApps` or `thresholdApps` so:
1. `formRefTbtcRandomBeacon
2. `formRefKeepApps`
3. `formRefThresholdApps` (but this might be confusing too, because Taco is also a threshold app).

src/pages/Staking/index.tsx
```diff
@@ -75,13 +75,14 @@ const StakingPage: PageComponent = (props) => {
...
+          {(address &&
             isStakingProvider &&
             isOperatorMappingInitialFetchDone &&
             (isAddressZero(mappedOperators.tbtc) ||
-              isAddressZero(mappedOperators.randomBeacon)) && (
+              isAddressZero(mappedOperators.randomBeacon))) ||
+            (isAddressZero(mappedOperators.taco) && (
               <OperatorAddressMappingCard stakingProvider={address} />
-            )}
+            ))}
```
michalsmiarowski, line 85: I think the brackets here are in wrong places, because now the `OperatorAddressMappingCard` is displayed when the user did not connect the wallet, where it should be the opposite 😬 :

image

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -89,33 +96,58 @@ const MapOperatorToStakingProviderForm = withFormik<
...
         checkIfOperatorIsMappedToAnotherStakingProvider,
         mappedOperatorTbtc,
         mappedOperatorRandomBeacon
       )
     }
+    if (!errors.operator && mappedOperatorTaco !== undefined) {
+      let validationMsg: string | undefined = ""
+      try {
+        const isOperatorMappedToAnotherStakingProvider =
+          await checkIfOperatorIsMappedToAnotherStakingProvider(values.operator)
```
michalsmiarowski, line 129: I think we should refactor `checkIfOperatorIsMappedToAnotherStakingProvider` function so that we can provide either an app name parameter or app type flag (where type is either `Keep` (random beacon + tbtc) or `nu` (taco)) so that we can check if the operator is mapped to another staking provider for current app.

Right now, when we call this function here, it fires for taco form validation and inside we check if the staking provider is mapped for tbtc + random beacon, where this should not matter (because keep and nu apps are independent of each other) so we actually should check it for taco.

michalsmiarowski (commented):
Reviewed only the code related to https://github.com/threshold-network/token-dashboard/issues/678. Here are my thoughts:

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: > Remember that there are cases where users can have tbtc operator mapped without random beacon.

Interesting, I didn't know that. I thought they had to be mapped to the same thing.

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -89,33 +96,58 @@ const MapOperatorToStakingProviderForm = withFormik<
...
+      !errors.operator &&
+      mappedOperatorTbtc !== undefined &&
+      mappedOperatorRandomBeacon !== undefined
+    ) {
       errors.operator = await validateInputtedOperatorAddress(
         values.operator,
         checkIfOperatorIsMappedToAnotherStakingProvider,
         mappedOperatorTbtc,
         mappedOperatorRandomBeacon
       )
```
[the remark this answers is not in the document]
theref: ah ok! this makes sense :) whilst they're not _supposed_ to be mapped to different things, due to user error they might be

src/pages/Staking/index.tsx
theref: hmmm, that's not what the behaviour I've seen when testing, but i'll take a close look

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -36,61 +36,125 @@ import ModalCloseButton from "../ModalCloseButton"
...
+          operator: operatorTaco,
+        },
+      ])
+    }
+    openModal(ModalType.MapOperatorToStakingProviderConfirmation, {
+      applications: selectedApps.map((_) => ({
+        appName: _.appName,
+        operator: _.operator,
+      })),
+    })
```
michalsmiarowski, line 146: This won't work properly, because changes made by setState (in this case `setSelectedApps`) are not reflected in the current submit function (in `selectedApps` variable). The reason is that the state is not expected to be changed in the same render. The state values are used by functions based on their current closures, and state updates will reflect in the next re-render by which the existing closures are not affected, but new ones are create.

Here is a [stack overflow thread](https://stackoverflow.com/questions/54069253/the-usestate-set-method-is-not-reflecting-a-change-immediately) that I found about the exact issue you have, it explains a lot.

I would propose to select the apps based on the operator value (check if it's empty or not), and then just store it in an array inside the submit function. Something like this:

```
const submitMapping = async () => {
    const operatorTbtc = formRefTbtc.current?.values?.operator
    const operatorRandomBeacon = formRefRandomBeacon.current?.values?.operator
    const operatorTaco = formRefTaco.current?.values?.operator

    const applications = []
    if (operatorTbtc) {
      await formRefTbtc.current?.validateForm()
      formRefTbtc.current?.setTouched({ operator: true }, false)
      applications.push({
        appName: "tbtc",
        operator: operatorTbtc,
      })
    }
    if (operatorRandomBeacon) {
      await formRefRandomBeacon.current?.validateForm()
      formRefRandomBeacon.current?.setTouched({ operator: true }, false)
      applications.push({
        appName: "randomBeacon",
        operator: operatorRandomBeacon,
      })
    }
    if (operatorTaco) {
      await formRefTaco.current?.validateForm()
      formRefTaco.current?.setTouched({ operator: true }, false)
      applications.push({
        appName: "taco",
        operator: operatorTaco,
      })
    }
    openModal(ModalType.MapOperatorToStakingProviderConfirmation, {
      applications: applications,
    })
  }
  ```
  
  Note tha validation for each form should also be done only if the operator value is set.

File: src/components/Modal/MapOperatorToStakingProviderModal/index.tsx [hunk 17]
```diff
@@ -36,61 +36,125 @@ import ModalCloseButton from "../ModalCloseButton"
...
   BaseModalProps & MapOperatorToStakingProviderModalProps
 > = () => {
   const { account } = useWeb3React()
-  const formRef =
+  const formRefTbtc =
     useRef<FormikProps<MapOperatorToStakingProviderFormValues>>(null)
+  const formRefRandomBeacon =
+    useRef<FormikProps<MapOperatorToStakingProviderFormValues>>(null)
+  const formRefTaco =
+    useRef<FormikProps<MapOperatorToStakingProviderFormValues>>(null)
```
michalsmiarowski, line 51: One more thing - I don't think we should separate tbtc and random beacon. Instead of 3 forms I would keep it as two forms - one for tbtc + random beacon (as it was before) and second one for taco. This is what I meant here: https://github.com/threshold-network/token-dashboard/pull/625#discussion_r1419550685 and here: https://github.com/threshold-network/token-dashboard/pull/625#discussion_r1419548789.

The reason is that the tbtc and random beacon should always (from what I know) have the same operator address, so keeping it in one form makes more sense.

michalsmiarowski (changes requested):
Left some comments regarding the `MapOperatorToStaking` form

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -28,43 +34,52 @@ const MapOperatorToStakingProviderFormBase: FC<
...
+            !isAddressZero(mappedOperatorRandomBeacon) &&
+            !isSameETHAddress(operator, mappedOperatorRandomBeacon)
+          ) {
+            validationMsg =
+              "The operator address doesn't match the one used in random beacon app"
+          }
+        }
+        break
+      case "taco":
+        console.log("taco")
```
cygnusv, line 79: There's a few console.log statements around. I don't think it's much of an issue, but maybe some of them can be cleaned up

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -108,6 +162,7 @@ const MapOperatorToStakingProviderModal: FC<
         </InfoBox>
         <BodyLg mt={"10"}>
           Choose an application to map the Operator Address:
+          {/*  JAMES TODO JAMES THEREF do we need another input for tbtc/rb */}
```
cygnusv, line 165: Is this still pending?

src/components/Modal/TACoCommitmentModal/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+          status="warning"
+          withIcon={false}
+          alignItems="center"
+          p={"8px 10px"}
+        >
+          <BodyXs>
+            Once you choose a lock-up duration and hit Authorize below, your
+            tokens will be irreversibly locked until the unlock date at the
+            earliest. This cannot be undone. You may authorize more tokens to
+            the same unlock horizon before the bonus deadline (12/30/23) , but
```
cygnusv, line 116:
```suggestion
            the same unlock horizon before the bonus deadline (December 30th, 2023), but
```

src/components/Modal/TACoCommitmentModal/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+            you cannot decrease the amount. You must also manually initiate
+            deauthorization 6 months before the unlock date stated if you wish
+            to withdraw tokens on said date. For more rules and information,
+            head to [link to documentation].
```
cygnusv, line 120: Missing documentation link (cc @arjunhassard )

src/components/StakingTimeline/index.tsx
```diff
@@ -89,9 +89,9 @@ export const LegacyStakesDepositSteps: FC = () => {
           size="sm"
         >
           <BodyMd>
-            For each stake, there are three applications available. PRE does not
-            require authorization. To authorize tBTC and Random Beacon, go to
-            the <Link to="/staking">Staking Page</Link> and select “Configure
+            For each stake, there are three applications available. To authorize
+            tBTC, TACo, and Random Beacon, go to the{" "}
```
cygnusv, line 93:
```suggestion
            tBTC, Random Beacon, and TACo, go to the{" "}
```

src/hooks/staking-applications/useRegisterMultipleOperatorsTransaction.ts
```diff
@@ -36,15 +43,17 @@ export const useRegisterMultipleOperatorsTransaction = () => {
...
+          throw new Error("All apps already have mapped operator!")
 
         if (isOperatorMappedOnlyInRandomBeacon)
           throw new Error("Random beacon app already has mapped operator!")
 
         if (isOperatorMappedOnlyInTbtc)
           throw new Error("Tbtc app already have mapped operator!")
 
+        if (isOperatorMappedOnlyInTaco)
+          throw new Error("TACo app already have mapped operator!")
```
cygnusv, line 56:
```suggestion
          throw new Error("TACo app already has mapped operator!")
```

src/pages/Staking/StakeDetailsPage/index.tsx
```diff
@@ -139,8 +154,8 @@ const StakeDetailsPage: FC = () => {
             isAddress
             address={stake.beneficiary}
           />
-          <StakeDetailRow label="PRE Node Status">
-            <NodeStatusLabel isAuthorized />
+          <StakeDetailRow label="TACo Node Status">
+            <NodeStatusLabel isAuthorized={tacoApp.isAuthorized} />
```
cygnusv, line 158: This should be changed to either represent "Authorization status" rather than "Node status", or check different events (e.g., operator bonded). However, given that it seems the same happens to tBTC/RB below, so we may discuss this for a later, unified change.

src/threshold-ts/applications/index.ts
```diff
@@ -314,7 +325,14 @@ export class Application implements IApplication {
...
     if (operator && !isAddressZero(operator)) {
-      isOperatorInPool = await this._application.isOperatorInPool(operator)
+      try {
+        isOperatorInPool = await this._application.isOperatorInPool(operator)
+      } catch (error) {
+        console.warn(
+          "isOperatorInPool method does not exist (eg on TACo app)",
+          error
+        )
+      }
```
cygnusv, line 335: This should be tackled as part of a later refactor that doesn't assume methods that are only relevant to TBTC (see https://github.com/threshold-network/token-dashboard/issues/627).

cygnusv (commented):
Looking good! Made a first pass, will make a second pass tomorrow

src/hooks/staking-applications/useRegisterMultipleOperatorsTransaction.ts
```diff
@@ -36,15 +43,17 @@ export const useRegisterMultipleOperatorsTransaction = () => {
...
-        if (isOperatorMappedInBothApps)
-          throw new Error("Both apps already have mapped operator!")
+        if (isOperatorMappedInAllApps)
+          throw new Error("All apps already have mapped operator!")
 
         if (isOperatorMappedOnlyInRandomBeacon)
           throw new Error("Random beacon app already has mapped operator!")
 
         if (isOperatorMappedOnlyInTbtc)
           throw new Error("Tbtc app already have mapped operator!")
```
derekpierre, line 53:
```suggestion
          throw new Error("Tbtc app already has mapped operator!")
```

src/pages/Staking/HowItWorks/StakingApplications/index.tsx
```diff
@@ -109,7 +115,8 @@ const StakingApplications: PageComponent = () => {
                 <Stack>
                   <BodyMd>Change your authorized amount at any time. </BodyMd>
                   <BodyXs>
-                    There is a deauthorization cooldown period of 45 days.
+                    There is a deauthorization cooldown period of 45 days for
+                    tBTC and Random Beacon apps and 180 days for TACo.
```
derekpierre, line 119: Should we be consistent with what we have in other places i.e. 6 months. Technically it is 182 days I believe.
```suggestion
                    tBTC and Random Beacon apps and 6 months for TACo.
```

src/pages/Staking/HowItWorks/StakingApplications/index.tsx
```diff
@@ -122,6 +129,28 @@ const StakingApplications: PageComponent = () => {
         />
       </Stack>
       <Stack spacing={6}>
+        <ApplicationDetailsCard
+          preTitle="TACo APP"
+          title="TACo is an access control layer that enables decentralized e2e encrypted data sharing in Web3 applications."
+          description="TACo is a plug-in access control layer and cryptographic primitive, operationally distributed across a cohort of Threshold nodes. Private data are encrypted by a data owner along with embedded access conditions. The data remain encrypted until it reaches the device of a qualifying recipient."
```
derekpierre, line 135:
```suggestion
          description="TACo is a plug-in access control layer and cryptographic primitive, operationally distributed across a cohort of Threshold nodes. Private data is encrypted by a data owner along with embedded access conditions. The data remains encrypted until it reaches the device of a condition-satisfying recipient."
```

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -9,48 +9,49 @@ import { getThresholdLib } from "../getThresholdLib"
...
     expect(resultRandomBeaconLabel).toBe(mockLabels[randomBeaconName])
+    expect(resultTacoLabel).toBe(mockLabels[tacoName])
   })
 
   it("returns correct app label if app name is given", () => {
     const resultTbtcLabel = getStakingAppLabelFromAppName(tbtcName)
-    const resultRandomBeaconLabel =
-      getStakingAppLabelFromAppName(randomBeaconName)
+    const resultRbLabel = getStakingAppLabelFromAppName(randomBeaconName)
+    const resultTacoLabel = getStakingAppLabelFromAppName(randomBeaconName)
```
derekpierre, line 41:
```suggestion
    const resultTacoLabel = getStakingAppLabelFromAppName(tacoName)
```

package.json
```diff
@@ -23,6 +23,7 @@
     "@ledgerhq/connect-kit-loader": "^1.1.2",
     "@reduxjs/toolkit": "^1.6.1",
     "@rehooks/local-storage": "^2.4.4",
+    "@nucypher/nucypher-contracts": "0.13.0",
```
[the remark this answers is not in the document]
cygnusv: It's not necessary, since all contracts and relevant ABIs are in 0.13.0

michalsmiarowski (changes requested):
Left some comments to look at before the merge. Still not finished with the review, might come back with more comments once I test it more.

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -117,29 +174,35 @@ const MapOperatorToStakingProviderModal: FC<
...
-                : ""
-            }
-            onSubmitForm={onSubmit}
+            innerRef={formRefTbtc}
+            formId="map-operator-to-staking-provider-form-tbtc"
             checkIfOperatorIsMappedToAnotherStakingProvider={
               checkIfOperatorIsMappedToAnotherStakingProvider
             }
+            appName={"tbtc"}
             mappedOperatorTbtc={mappedOperatorTbtc}
```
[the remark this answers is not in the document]
michalsmiarowski: Also I think we should make sure with @lukasz-zimnoch if the operator mapped for tbtc and random-beacon apps should be the same or they can be completely different. We implemented it in such way on the front end back in the day, but based on my testing I think it's possible to use different operators on the contract side 🤔 I did a transaction with my wallet and it didn't throw any error. Maybe that was a design decision to do that in the dApp?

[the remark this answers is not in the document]
michalsmiarowski: Confirmed with @lukasz-zimnoch that operator mapped for tbtc and random-beacon apps should be the same so we should keep the validation on the front end.

package.json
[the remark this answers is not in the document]
theref: marking resolved

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -104,6 +116,7 @@ const MapOperatorToStakingProviderForm = withFormik<
     if (!errors.operator) {
       errors.operator = await validateInputtedOperatorAddress(
         values.operator,
+        values.appName,
         checkIfOperatorIsMappedToAnotherStakingProvider,
         mappedOperatorTbtc,
         mappedOperatorRandomBeacon
```
michalsmiarowski, line 122: 5. Improve validation here:

```
  if (values.operator) {
    errors.operator = validateETHAddress(values.operator)
    if (!errors.operator) {
      errors.operator = await validateInputtedOperatorAddress(
        values.operator,
        checkIfOperatorIsMappedToAnotherStakingProvider,
        mappedOperatorTbtc,
        mappedOperatorRandomBeacon
      )
    }
  }
```

Note that we've added `if (values.operator)` condition at the top, before `validateETHAddress` call. This is neede so that we don't run validation if the input is empyt (so that we can submit either RandomBeaconTbtc or Taco).

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -36,52 +37,104 @@ import ModalCloseButton from "../ModalCloseButton"
...
+            operator
+          )
+        return (
+          !isAddressZero(stakingProviderMapped) &&
+          !isSameETHAddress(stakingProviderMapped, account!)
+        )
+      default:
+        throw new Error(`Unsupported app name: ${appName}`)
     }
   }
```
michalsmiarowski, line 97: 6. Split this into three cases based on the `appName`:

```
  const checkIfOperatorIsMappedToAnotherStakingProvider: (
    operator: string,
    appName: string
  ) => Promise<boolean> = async (operator: string, appName: string) => {
    let stakingProviderMapped
    switch (appName) {
      case "tbtc":
        const stakingProviderMappedEcdsa =
          await threshold.multiAppStaking.ecdsa.operatorToStakingProvider(
            operator
          )
        return (
          !isAddressZero(stakingProviderMappedEcdsa) &&
          !isSameETHAddress(stakingProviderMappedEcdsa, account!)
        )
      case "randomBeacon":
        const stakingProviderMappedRandomBeacon =
          await threshold.multiAppStaking.randomBeacon.operatorToStakingProvider(
            operator
          )
        return (
          !isAddressZero(stakingProviderMappedRandomBeacon) &&
          !isSameETHAddress(stakingProviderMappedRandomBeacon, account!)
        )
      case "taco":
        stakingProviderMapped =
          await threshold.multiAppStaking.taco.operatorToStakingProvider(
            operator
          )
        return (
          !isAddressZero(stakingProviderMapped) &&
          !isSameETHAddress(stakingProviderMapped, account!)
        )
      default:
        throw new Error(`Unsupported app name: ${appName}`)
    }
  }
```

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -36,52 +37,104 @@ import ModalCloseButton from "../ModalCloseButton"
...
+    }
+    const isValid =
+      applications.length > 0 &&
+      applications.every((app) => app.formRef.current?.isValid)
+    if (isValid) {
+      openModal(ModalType.MapOperatorToStakingProviderConfirmation, {
+        applications: applications,
+      })
+    }
   }
```
michalsmiarowski, line 138: 7. The `applications` type here should be the same as the `applications` type in `MapOperatorToStakingProviderConfirmation`. Besides that we shouldn't push two applications (`randomBeacon` and `tbtc`) in all cases because it will throw an errror if user has only tbtc app validated. I propose this:

```
const submitMapping = async () => {
    const operatorTbtcRb = formRefTbtcRb.current?.values?.operator
    const operatorTaco = formRefTaco.current?.values?.operator

    const applications: ApplicationForOperatorMapping[] = []
    let isValid = true
    if (operatorTbtcRb) {
      await formRefTbtcRb.current?.validateForm()
      formRefTbtcRb.current?.setTouched({ operator: true }, false)
      if (mappedOperatorTbtc && isAddressZero(mappedOperatorTbtc)) {
        applications.push({
          appName: "tbtc",
          operator: operatorTbtcRb,
          stakingProvider: account!,
        })
      }
      if (
        mappedOperatorRandomBeacon &&
        isAddressZero(mappedOperatorRandomBeacon)
      ) {
        applications.push({
          appName: "randomBeacon",
          operator: operatorTbtcRb,
          stakingProvider: account!,
        })
      }
      isValid = isValid && formRefTbtcRb.current?.isValid
    }
    if (operatorTaco) {
      await formRefTaco.current?.validateForm()
      formRefTaco.current?.setTouched({ operator: true }, false)
      applications.push({
        appName: "taco",
        operator: operatorTaco,
        stakingProvider: account!,
      })
      isValid = isValid && formRefTaco.current?.isValid
    }
    if (applications.length > 0 && isValid) {
      openModal(ModalType.MapOperatorToStakingProviderConfirmation, {
        applications: applications,
      })
    }
  }
```

I've also introduced new interface here: `ApplicationForOperatorMapping` - see next point

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
```diff
@@ -52,16 +53,13 @@ const OperatorMappingConfirmation: FC<
...
 const MapOperatorToStakingProviderConfirmationModal: FC<
   BaseModalProps & {
-    operator: string
-    isOperatorMappedOnlyInTbtc: boolean
-    isOperatorMappedOnlyInRandomBeacon: boolean
+    applications: {
+      appName: string
+      operator: string
+      stakingProvider: string
+    }[]
```
michalsmiarowski, line 60: 8. Based on the point aboe - extract it to a separate type and export it:

```
export interface ApplicationForOperatorMapping {
  appName: string
  operator: string
  stakingProvider: string
}

const MapOperatorToStakingProviderConfirmationModal: FC<
  BaseModalProps & {
    applications: ApplicationForOperatorMapping[]
  }
  ```

File: src/components/Modal/MapOperatorToStakingProviderModal/index.tsx [hunk 36]
```diff
@@ -109,40 +164,37 @@ const MapOperatorToStakingProviderModal: FC<
...
+            <MapOperatorToStakingProviderForm
+              innerRef={formRefTbtcRb}
+              formId="map-operator-to-staking-provider-form-tbtc"
+              checkIfOperatorIsMappedToAnotherStakingProvider={
+                checkIfOperatorIsMappedToAnotherStakingProvider
+              }
+              appName={"tbtcRb"}
+              mappedOperatorRandomBeacon={mappedOperatorRandomBeacon}
+              mappedOperatorTbtc={mappedOperatorTbtc}
+            />
```
michalsmiarowski, line 195: 9. Change this form for tbtcRb:

```
              <MapOperatorToStakingProviderForm
                innerRef={formRefTbtcRb}
                formId="map-operator-to-staking-provider-form-tbtc"
                initialAddress={
                  isOperatorMappedOnlyInRandomBeacon
                    ? mappedOperatorRandomBeacon
                    : isOperatorMappedOnlyInTbtc
                    ? mappedOperatorTbtc
                    : ""
                }
                checkIfOperatorIsMappedToAnotherStakingProvider={async (
                  operator: string
                ) => {
                  const isOperatorAlreadyMappedTbtc =
                    await checkIfOperatorIsMappedToAnotherStakingProvider(
                      operator,
                      "tbtc"
                    )
                  const isOperatorAlreadyMappedRandomBeacon =
                    await checkIfOperatorIsMappedToAnotherStakingProvider(
                      operator,
                      "tbtc"
                    )
                  return (
                    isOperatorAlreadyMappedTbtc &&
                    isOperatorAlreadyMappedRandomBeacon
                  )
                }}
                mappedOperatorRandomBeacon={mappedOperatorRandomBeacon}
                mappedOperatorTbtc={mappedOperatorTbtc}
              />
```

Here we add `initialAddress` and we pass a callback funtion to `checkIfOperatorIsMappedToAnotherStakingProvider` where we call `checkIfOperatorIsMappedToAnotherStakingProvider` with a proper app arguments.

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -155,12 +207,36 @@ const MapOperatorToStakingProviderModal: FC<
...
+            <MapOperatorToStakingProviderForm
+              innerRef={formRefTaco}
+              formId="map-operator-to-staking-provider-form-taco"
+              checkIfOperatorIsMappedToAnotherStakingProvider={
+                checkIfOperatorIsMappedToAnotherStakingProvider
+              }
+              appName={"taco"}
+              mappedOperatorTaco={mappedOperatorTaco}
+            />
```
michalsmiarowski, line 231: 10. (final step) Refactor this form:

```
            <MapOperatorToStakingProviderForm
              innerRef={formRefTaco}
              formId="map-operator-to-staking-provider-form-taco"
              checkIfOperatorIsMappedToAnotherStakingProvider={async (
                operator: string
              ) => {
                return await checkIfOperatorIsMappedToAnotherStakingProvider(
                  operator,
                  "taco"
                )
              }}
              mappedOperatorTaco={mappedOperatorTaco}
            />
```

SImilarly to above we will call `checkIfOperatorIsMappedToAnotherStakingProvider` but this time with `taco` as an app name.

And that should be it. I belive that this is not an ideal solution though. Ideally we would want to further separate those two forms (RandomBeaconTbtc and Taco) and just extract similar logic to a shared component, but considering how much time we have then I think this would be enough for now.

michalsmiarowski (changes requested):
Left some important comments that should be addressed before the merge.

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -109,40 +164,37 @@ const MapOperatorToStakingProviderModal: FC<
...
-            }
-            onSubmitForm={onSubmit}
-            checkIfOperatorIsMappedToAnotherStakingProvider={
-              checkIfOperatorIsMappedToAnotherStakingProvider
-            }
-            mappedOperatorTbtc={mappedOperatorTbtc}
-            mappedOperatorRandomBeacon={mappedOperatorRandomBeacon}
-          />
-        </Box>
+        {isEmptyOrZeroAddress(mappedOperatorTbtc) && (
```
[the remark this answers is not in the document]
michalsmiarowski: Actually this is wrong condition. The prettier removed the brackets for me 😬  Should be:

```suggestion
        {(isEmptyOrZeroAddress(mappedOperatorTbtc) ||
          isEmptyOrZeroAddress(mappedOperatorRandomBeacon)) && (
```

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -109,40 +164,37 @@ const MapOperatorToStakingProviderModal: FC<
...
+            borderRadius={"12px"}
+            mt={"5"}
+            mb={"5"}
+          >
+            <LabelSm>
+              {isEmptyOrZeroAddress(mappedOperatorTbtc)
+                ? "tBTC"
+                : isEmptyOrZeroAddress(mappedOperatorRandomBeacon)
+                ? "Random Beacon"
+                : "tBTC & Random Beacon"}
```
michalsmiarowski, line 181: Let's rever it to previous implementation, because it shows wrong values right now:

```suggestion
          {isOperatorMappedOnlyInRandomBeacon ? (
            <LabelSm>tBTC app</LabelSm>
          ) : isOperatorMappedOnlyInTbtc ? (
            <LabelSm>random beacon app</LabelSm>
          ) : (
            <LabelSm>tBTC + Random Beacon apps (requires 2txs)</LabelSm>
          )}
```

michalsmiarowski (changes requested):
two more comments below

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -109,40 +164,37 @@ const MapOperatorToStakingProviderModal: FC<
...
+              }
+              appName={"tbtcRb"}
+              mappedOperatorRandomBeacon={mappedOperatorRandomBeacon}
+              mappedOperatorTbtc={mappedOperatorTbtc}
+            />
+          </Box>
+        )}
         <AlertBox
           status="warning"
           withIcon={false}
```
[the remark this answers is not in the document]
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/17b3668529cd43c75f4e64d8641ea8541332fdf9

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -28,43 +34,51 @@ const MapOperatorToStakingProviderFormBase: FC<
...
+          }
+        }
+        break
+      case "taco":
+        break
+      default:
+        throw new Error(`Unsupported app name: ${appName}`)
     }
   } catch (error) {
     console.error("`MapOperatorToStakingProviderForm` validation error.", error)
```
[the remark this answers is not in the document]
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/7897c853021d0280902d0b9444653ae843ff6924

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -75,23 +89,21 @@ const validateInputtedOperatorAddress = async (
...
+  mappedOperatorRandomBeacon?: string
+  mappedOperatorTaco?: string
   innerRef: Ref<FormikProps<MapOperatorToStakingProviderFormValues>>
   checkIfOperatorIsMappedToAnotherStakingProvider: (
-    operator: string
+    operator: string,
+    appName: string
   ) => Promise<boolean>
-  onSubmitForm: (values: MapOperatorToStakingProviderFormValues) => void
 } & ComponentProps
```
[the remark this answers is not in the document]
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -75,23 +89,21 @@ const validateInputtedOperatorAddress = async (
...
 
 const MapOperatorToStakingProviderForm = withFormik<
   MapOperatorToStakingProviderFormProps,
   MapOperatorToStakingProviderFormValues
 >({
-  mapPropsToValues: ({ initialAddress }) => ({
-    operator: initialAddress,
-  }),
```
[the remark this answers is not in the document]
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
```diff
@@ -1,20 +1,26 @@
...
 
-export interface MapOperatorToStakingProviderFormValues {
+export type MapOperatorToStakingProviderFormValues = {
   operator: string
+  appName: string
+  mappedOperatorTbtc?: string
+  mappedOperatorRandomBeacon?: string
+  mappedOperatorTaco?: string
+  innerRef: Ref<FormikProps<MapOperatorToStakingProviderFormValues>>
 }
```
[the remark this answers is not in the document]
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/MapOperatorToStakingProviderForm.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/50fd0c87c787d079509eb083d894e51b74419681

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -109,58 +177,108 @@ const MapOperatorToStakingProviderModal: FC<
...
+                      )
+                    const isOperatorAlreadyMappedRandomBeacon =
+                      await checkIfOperatorIsMappedToAnotherStakingProvider(
+                        operator,
+                        "tbtc"
+                      )
+                    return (
+                      isOperatorAlreadyMappedTbtc &&
+                      isOperatorAlreadyMappedRandomBeacon
+                    )
```
michalsmiarowski, line 227: I think I've done a mistake here: https://github.com/threshold-network/token-dashboard/pull/625#discussion_r1429891983

We should use OR (`||`) operator here instead of AND {`&&`}. The condition should work the same as here:
https://github.com/threshold-network/token-dashboard/blob/17afd96148b72593c20d2c487bd79047f874a927/src/components/Modal/MapOperatorToStakingProviderModal/index.tsx#L79-L84
and we use the OR operator there.

```suggestion
                    return (
                      isOperatorAlreadyMappedTbtc ||
                      isOperatorAlreadyMappedRandomBeacon
                    )
```

michalsmiarowski (changes requested):

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
```diff
@@ -71,57 +66,77 @@ const MapOperatorToStakingProviderConfirmationModal: FC<
...
+      switch (app.appName) {
+        case "tbtc":
+          transaction = await registerOperatorTbtc(app.operator)
+          break
+        case "randomBeacon":
+          transaction = await registerOperatorRandomBeacon(app.operator)
+          break
+        case "taco":
+          transaction = await registerOperatorTaco(account, app.operator)
+          break
```
michalsmiarowski, line 120: There is not information in the dApp that you have to `Commit to Taco` first before mapping the operator. WIthout the commitment it throws an error:

`Not owner or provider`

image

michalsmiarowski (changes requested):
Left some comments regarding TACo commitment

src/components/Modal/TACoCommitmentModal/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+                <Td>2%</Td>
+              </Tr>
+              <Tr>
+                <Td>
+                  <Radio value="47174400" />
+                </Td>
+                <Td>24 months (6 month minimum + 18 month extension)</Td>
+                <Td>{getEndDate(47174400)}</Td>
+                <Td>3%</Td>
+              </Tr>
```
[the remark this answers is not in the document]
theref: that is expected on development due to the version of the contract that is deployed

michalsmiarowski (changes requested):

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
michalsmiarowski: The type is more related to `MapOperatorToStakingProviderConfirmationModal` so it would be better if it was moved there, just above the component.

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
michalsmiarowski: Actually, after further testing, it looks it's not commit that is required but authorization. The stake should be authorized for Taco first with the minimum amount of T tokens (which is 40k). Only then the mapping is allowed.

Here is where the error is thrown in the contract: https://github.com/nucypher/nucypher-contracts/blob/2f7b241aa216ff85ebdea3b796285ef35ae57768/contracts/contracts/TACoApplication.sol#L276
It checks if `stakingProviderInfo[_stakingProvider].authorized` value greater than 0 [here](https://github.com/nucypher/nucypher-contracts/blob/2f7b241aa216ff85ebdea3b796285ef35ae57768/contracts/contracts/TACoApplication.sol#L653-L655). I believe that the `authorized` amount is set during the autorization, authorizationIncrease and deauthorization.

Maybe we should reflect that the authorization is needed to map the operator in the UI? 🤔

src/components/Modal/TACoCommitmentModal/index.tsx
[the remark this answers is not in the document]
michalsmiarowski: I see. In this case I'm resolving the issue.

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/d1c18620ceb3a34e1ae7ab4a5ed176470eace2a6

theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/ed0535912603406365af2a8e3f4457b5fce39d72

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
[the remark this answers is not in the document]
michalsmiarowski: Addresed in d1c18620ceb3a34e1ae7ab4a5ed176470eace2a6

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
theref: if i make the Button only appear if TACo has already been authorized, would that suffice?

theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/d174a384aeb25eca1a47cf7d347f8774fa9b69c7

cygnusv (dismissed):
LGTM, this is a massive piece of work @theref and essential for taco release.

src/components/Modal/TACoCommitmentModal/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <ModalFooter>
+        <Button onClick={closeModal} variant="outline" mr={2}>
+          Cancel
+        </Button>
+        <Button
+          onClick={() => submitCommitment(stakingProvider, value)}
+          type="submit"
+        >
+          Commit
+        </Button>
```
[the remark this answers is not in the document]
cygnusv: Can we compromise here?

src/store/account/selectors.ts
```diff
@@ -8,21 +8,29 @@ export const selectAccountState = (state: RootState) => state.account
...
+        isOperatorMappedOnlyInRandomBeacon ||
+        isOperatorMappedOnlyInTbtc ||
+        isOperatorMappedOnlyInTaco,
+      isOperatorMappedInAllApps:
+        !isAddressZero(randomBeacon) &&
+        !isAddressZero(tbtc) &&
+        !isAddressZero(taco),
     }
   }
 )
```
michalsmiarowski, line 36: We have to refactor it a little bit. `isOperatorMappedOnlyInTbtc` and `isOperatorMappedOnlyInRandomBeacon` helper variables were created with the thought of bundled rewards from tbtc + random beacon. We shouldn't add taco app here because it will then displays the `tbtc + random beacon` label when we have tbtc and taco apps mapped. My proposal would be to:

1. Refactor `selectMappedOperators` selector:

```
export const selectMappedOperators = createSelector(
  [selectAccountState],
  (accountState: AccountState) => {
    const { randomBeacon, tbtc, taco } = accountState.operatorMapping.data
    const isOperatorMappedOnlyInTbtcForBundledRewards =
      !isAddressZero(tbtc) && isAddressZero(randomBeacon)
    const isOperatorMappedOnlyInRandomBeaconForBundledRewards =
      isAddressZero(tbtc) && !isAddressZero(randomBeacon)

    return {
      mappedOperatorTbtc: tbtc,
      mappedOperatorRandomBeacon: randomBeacon,
      mappedOperatorTaco: taco,
      isOperatorMappedOnlyInTbtcForBundledRewards,
      isOperatorMappedOnlyInRandomBeaconForBundledRewards,
      isOperatorMappedInAllApps:
        !isAddressZero(randomBeacon) &&
        !isAddressZero(tbtc) &&
        !isAddressZero(taco),
    }
  }
)
```

We will remove `isOperatorMappedOnlyInTaco` and `isOneOfTheAppsNotMapped` properties completely adn rename those two:

`isOperatorMappedOnlyInTbtc` -> `isOperatorMappedOnlyInTbtcForBundledRewards`
`isOperatorMappedOnlyInRandomBeacon` -> `isOperatorMappedOnlyInRandomBeaconForBundledRewards`

src/pages/Staking/OperatorAddressMappingCard.tsx
```diff
@@ -33,12 +33,14 @@ const OperatorAddressMappingCard: FC<{ stakingProvider: string }> = ({
   const {
     mappedOperatorTbtc,
     mappedOperatorRandomBeacon,
+    mappedOperatorTaco,
     isOneOfTheAppsNotMapped,
   } = useAppSelector(selectMappedOperators)
```
michalsmiarowski, line 38: Destructurize `isOperatorMappedOnlyInRandomBeaconForBundledRewards` and `isOperatorMappedOnlyInTbtcForBundledRewards` here:

```
  const {
    mappedOperatorTbtc,
    mappedOperatorRandomBeacon,
    mappedOperatorTaco,
    isOperatorMappedOnlyInRandomBeaconForBundledRewards,
    isOperatorMappedOnlyInTbtcForBundledRewards,
  } = useAppSelector(selectMappedOperators)
```

Then create a helper variable `isOneOfTheBundledAppsNotMapped`:

```
  const isOneOfTheBundledAppsNotMapped =
    isOperatorMappedOnlyInRandomBeaconForBundledRewards ||
    isOperatorMappedOnlyInTbtcForBundledRewards
```

and change all the occurences of  `isOneOfTheAppsNotMapped` to (->)`isOneOfTheBundledAppsNotMapped`.

src/hooks/staking-applications/useRegisterMultipleOperatorsTransaction.ts
```diff
@@ -36,15 +43,17 @@ export const useRegisterMultipleOperatorsTransaction = () => {
...
 
         if (isOperatorMappedOnlyInRandomBeacon)
           throw new Error("Random beacon app already has mapped operator!")
 
         if (isOperatorMappedOnlyInTbtc)
-          throw new Error("Tbtc app already have mapped operator!")
+          throw new Error("Tbtc app already has mapped operator!")
 
+        if (isOperatorMappedOnlyInTaco)
+          throw new Error("TACo app already has mapped operator!")
```
michalsmiarowski, line 56: 3. Change those conditions to:

```
        if (!isEmptyOrZeroAddress(mappedOperatorRandomBeacon))
          throw new Error("Random beacon app already has mapped operator!")

        if (!isEmptyOrZeroAddress(mappedOperatorTbtc))
          throw new Error("Tbtc app already has mapped operator!")

        if (!isEmptyOrZeroAddress(mappedOperatorTaco))
          throw new Error("TACo app already has mapped operator!")
```

and then remove those 3 variables (`isOperatorMappedOnlyInRandomBeacon`, `isOperatorMappedOnlyInTbtc` and `isOperatorMappedOnlyInTaco`) from `useAppSelector` which is few lines above.

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -36,52 +37,117 @@ import ModalCloseButton from "../ModalCloseButton"
...
   const threshold = useThreshold()
 
   const {
     mappedOperatorTbtc,
     mappedOperatorRandomBeacon,
+    mappedOperatorTaco,
     isOperatorMappedOnlyInRandomBeacon,
     isOperatorMappedOnlyInTbtc,
+    isOperatorMappedOnlyInTaco,
   } = useAppSelector(selectMappedOperators)
```
michalsmiarowski, line 68: 4. Refactor this `useAppSelector` destructuring:

```
const {
  const {
    mappedOperatorTbtc,
    mappedOperatorRandomBeacon,
    mappedOperatorTaco,
    isOperatorMappedOnlyInTbtcForBundledRewards,
    isOperatorMappedOnlyInRandomBeaconForBundledRewards,
    isOperatorMappedInAllApps,
  } = useAppSelector(selectMappedOperators)
```

Then, in this file, change all occurences of:
1. `isOperatorMappedOnlyInRandomBeacon` to `isOperatorMappedOnlyInTbtcForBundledRewards`
2. `isOperatorMappedOnlyInTbtc` to `isOperatorMappedOnlyInRandomBeaconForBundledRewards`

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
```diff
@@ -90,7 +156,9 @@ const MapOperatorToStakingProviderModal: FC<
       <ModalCloseButton />
       <ModalBody>
         <InfoBox variant="modal">
-          {isOperatorMappedOnlyInRandomBeacon || isOperatorMappedOnlyInTbtc ? (
+          {isOperatorMappedOnlyInRandomBeacon ||
+          isOperatorMappedOnlyInTbtc ||
+          isOperatorMappedOnlyInTaco ? (
             <H5>
               We noticed you've only mapped 1 application's Operator Address.
             </H5>
```
michalsmiarowski, line 164: 5. Last thing is to change this condition to:

```
          {isOperatorMappedInAllApps ? (
            <H5>
              We noticed you've not mapped Operator Address to some
              applications.
            </H5>
```

We should display a complete new subtitle here, because we've got 3 applictions now. I proposed the one above, which is:
"We noticed you've not mapped Operator Address to some  applications."
which should be displayed when not all of the apps have mapped operators.

michalsmiarowski (changes requested):

src/store/account/selectors.ts
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/3915d2d88cc07e25c2fe36bc0f6f3f76614dae9a

src/pages/Staking/OperatorAddressMappingCard.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/3915d2d88cc07e25c2fe36bc0f6f3f76614dae9a

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/3915d2d88cc07e25c2fe36bc0f6f3f76614dae9a

src/components/Modal/MapOperatorToStakingProviderModal/index.tsx
theref: https://github.com/threshold-network/token-dashboard/pull/625/commits/3915d2d88cc07e25c2fe36bc0f6f3f76614dae9a

michalsmiarowski (changes requested):
Two more comments

src/components/Modal/TACoCommitmentModal/index.tsx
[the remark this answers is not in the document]
michalsmiarowski: Was addressed in b1dc9f8b2c2e4113234da8877058425ee69feaf4

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
michalsmiarowski: Not sure if this change is needed? @theref @cygnusv 

Do you want to have the commit button available only when taco app is authorised?

michalsmiarowski (dismissed):
LGTM

cygnusv (dismissed):

src/components/Modal/MapOperatorToStakingProviderConfirmationModal/index.tsx
cygnusv: Yes, only after authorized

cygnusv (approved):
Approving again after the rebase due to the need of signed commits. As seen here, https://github.com/threshold-network/token-dashboard/compare/524f0b8b2b07f1f57343a061820954fb9579c525..a188adf344fced8ef2c17cce9c594ac6052e79d2, no actual changes happened

michalsmiarowski (approved):
Adding approve again after rebase to make all commits verified 👍

Issue #573, opened 2023-07-13
Hide the info bubble once apps are authorized


erdogan:
Hide / lose this bubble that is still visible after 100% authorization of apps

<img width="1103" alt="image" src="https://github.com/threshold-network/token-dashboard/assets/14485/83dd7448-ec20-4112-8410-39d7d3accc8e">

```[tasklist]
### Tasks
- [ ] https://github.com/threshold-network/token-dashboard/pull/628
```

PR #628, opened 2023-09-29
Add conditional rendering for authorization alert

kpyszkowski:
Closes: https://github.com/threshold-network/token-dashboard/issues/573

Aggregated authorization apps statuses into an `isFullyAuthorized` boolean indicating if the authorization process is fulfilled and used this boolean to conditionally render the `AlertBox` component.

### DIFF
--- a/src/pages/Staking/AuthorizeStakingApps/index.tsx
+++ b/src/pages/Staking/AuthorizeStakingApps/index.tsx
@@ -115,8 +115,9 @@ const AuthorizeStakingAppsPage: FC = () => {
   }
 
   const isFullyAuthorized = Object.values(appsAuthData).every(
-    ({ status }) =>
-      status === "authorized" || status === "authorization-not-required"
+    ({ status, percentage }) =>
+      (status === "authorized" && percentage === 100) ||
+      status === "authorization-not-required"
   )
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/pages/Staking/AuthorizeStakingApps/index.tsx
```diff
@@ -114,6 +114,11 @@ const AuthorizeStakingAppsPage: FC = () => {
     },
   }
 
+  const isFullyAuthorized = Object.values(appsAuthData).every(
+    ({ status }) =>
+      status === "authorized" || status === "authorization-not-required"
+  )
```
michalsmiarowski, line 120: We need to modify this condition. A stake only requires 40k of its tokens authorized to be considered 'authorized.' If we have a stake of 100k T and authorize only 40k T, it will still be labeled 'authorized,' but in reality, it's not fully authorized.

Since it is mentioned in [the issue](https://github.com/threshold-network/token-dashboard/issues/573) that the info bubble should be hidden only after the app is 100% authorized, we should refine this condition accordingly:

```suggestion
  const isFullyAuthorized = Object.values(appsAuthData).every(
    ({ status, percentage }) =>
      (status === "authorized" && percentage === 100) ||
      status === "authorization-not-required"
  )
```

michalsmiarowski (changes requested):
Overall looking good 🔥 I've just proposed some slight improvement to `isFullyAuthorized` condition.

src/pages/Staking/AuthorizeStakingApps/index.tsx
kpyszkowski: c8a8577cf041c04bb411544ba715c03977d275e0

michalsmiarowski (approved):
LGTM

Issue #589, opened 2023-08-07
Modify TVL label in dashboard with more descriptive terms


erdogan:
- [x] Change **TVL** to **TOTAL SUPPLY** to be consistent with the Dune Dashboard labels
- [x] Change the **TOTAL VALUE LOCKED** to  **TOTAL TVL (BTC + STAKING)** (which should be around $124M)

![image](https://github.com/threshold-network/token-dashboard/assets/14485/a00eb7e3-a43f-4df0-926d-5547fa298270)

```[tasklist]
### Tasks
- [ ] https://github.com/threshold-network/token-dashboard/pull/629
```

michalsmiarowski:
@erdogan I need more details about this:

> Change the TOTAL VALUE LOCKED to TOTAL TVL (BTC + STAKING) (which should be around $124M)

What do we actually want to include in our calculations for `TOTAL TVL (BTC + STAKING)`? Additionally, could you please specify the source of the `$124M` value? Is this figure available on the Dune dashboard?"

Right now we include tokens from staking, tbtc, coverage pools and ecdsa. You can see it here: https://github.com/threshold-network/token-dashboard/blob/main/src/hooks/useFetchTvl.ts#L178-L185

PR #629, opened 2023-09-29
Modify TVL labels

kpyszkowski:
Ref: https://github.com/threshold-network/token-dashboard/issues/589

Modified overview page sections labels:
- _TVL_ ➡️ _TOTAL SUPPLY_
- _TOTAL VALUE LOCKED_ ➡️ _TOTAL TVL (BTC + STAKING)_

### DIFF
--- a/src/components/tBTC/Stats.tsx
+++ b/src/components/tBTC/Stats.tsx
@@ -15,7 +15,9 @@ export type TvlProps = {
 export const Tvl: FC<TvlProps> = ({ tvl, tvlInUSD }) => {
   return (
     <>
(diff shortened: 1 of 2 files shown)


michalsmiarowski:
@kpyszkowski 

Can we change how we reference the issue in this PR? `Closes: <task_number>` will close the issue automatically when we merge the PR and I think we don't want to do that yet, because there is still an open discussion in that task about `TOTAL TVL` value difference.

I think for now we can change `Closes: <task_number>` to just `Ref: <task_number>`

michalsmiarowski (approved):
LGTM

Issue #547, opened 2023-06-15
Bring "How it works" page to forefront


erdogan:
```[tasklist]
### Tasks
- [ ] https://github.com/threshold-network/token-dashboard/pull/630
```

erdogan:
Let the users check the How it works page before the warning modal.

sashatanase:
Proposed Solution

Show the modal only if the users hit Bridge in the secondary nav; otherwise, let them check the How it Works page and tBTC explorer.

Added How it works as a secondary button in the warning modal 

![image](https://github.com/threshold-network/token-dashboard/assets/40768736/df601a52-32d9-40e7-8ae1-49e190805d49)

Figma - https://www.figma.com/file/Tx1lZc7us4SqCeAgbt1hBV/Threshold-Dapp?type=design&node-id=9390%3A173582&mode=design&t=nnmPtmyh3yigcEoB-1

PR #630, opened 2023-10-02
New tBTC app modal refinements

kpyszkowski:
Closes: https://github.com/threshold-network/token-dashboard/issues/547

Added "How it works" button for `NewTBTCApp` modal component. Reordered sub navigation links. Changed default page for tBTC section.


src/components/SubNavigationPills/index.tsx
```diff
@@ -26,9 +27,36 @@ interface NavPill extends RouteProps {
...
+          ...order.map(
+            (pathInOrder) =>
+              links.find(
+                ({ path: linkPath }) => pathInOrder === linkPath
+              ) as RouteProps
+          ),
+          ...links.filter(({ path }) => !order.includes(path)),
+        ].filter((item) => !!item)
+      : links
+  const linksWithTitle = reorderedLinks.filter((link) => !!link.title)
```
michalsmiarowski, line 59: This algorithm for ordering pages in unnecessary. To change the order of the sub-navigation in tBTC page you just have to change the order of `pages` property here:

https://github.com/threshold-network/token-dashboard/blob/58e4dc839bb9d02765e1ccce837e200eb74b5b51/src/pages/tBTC/index.tsx#L15

src/components/SubNavigationPills/index.tsx
```diff
@@ -26,9 +27,36 @@ interface NavPill extends RouteProps {
...
+    order.length > 0 &&
+    order.every((pathInOrder) => {
+      const linksPaths = links.map(({ path }) => path)
+      return linksPaths.includes(pathInOrder)
+    })
+  if (!isOrderPropValid) {
+    throw new Error(
+      `${SubNavigationPills.name}: Invalid order keys provided. It should match with route paths.`
+    )
+  }
```
michalsmiarowski, line 44: Similar to above this code is also unnecessary

src/components/SubNavigationPills/index.tsx
```diff
@@ -13,6 +13,7 @@ import Link from "../Link"
 
 interface SubNavigationPillsProps {
   links: RouteProps[]
+  order?: string[]
```
michalsmiarowski, line 16: Same with the order property here. It also breaks the `Upgrade` page, because the `order` property was not passed there:

image

michalsmiarowski (changes requested):
Left some comments to look at down below 👇

michalsmiarowski:
We should also make "How it works" subpage a primary page when you click "tBTC" on the left navigation. Right now when you click it it automatically goes to "Bridge" subpage.

src/components/SubNavigationPills/index.tsx
kpyszkowski: done

src/components/SubNavigationPills/index.tsx
kpyszkowski: reverted

src/components/SubNavigationPills/index.tsx
kpyszkowski: resolved by reverting changes

src/components/SubNavigationPills/index.tsx
```diff
@@ -1,12 +1,12 @@
...
 import {
-  Box,
   BodyMd,
+  Box,
   Divider,
   HStack,
   Stack,
   useColorModeValue,
 } from "@threshold-network/components"
+import { FC } from "react"
```
michalsmiarowski, line 9: Hmmm, I think those changes are unnecessary. They just reorder the imports

src/components/Modal/tBTC/NewTBTCApp.tsx
```diff
@@ -1,28 +1,30 @@
...
-import { TakeNoteList } from "../../tBTC"
-import withBaseModal from "../withBaseModal"
-import tbtcAppBannerIllustration from "../../../static/images/tBTCAppBannerWithGrid.svg"
+import { FC } from "react"
+import { useNavigate } from "react-router-dom"
 import { useTBTCTerms } from "../../../hooks/useTBTCTerms"
+import tbtcAppBannerIllustration from "../../../static/images/tBTCAppBannerWithGrid.svg"
 import { BaseModalProps } from "../../../types"
+import { TakeNoteList } from "../../tBTC"
+import withBaseModal from "../withBaseModal"
```
michalsmiarowski, line 19: Similar to above there are also unnecessary changes here

michalsmiarowski (changes requested):
Works great now 🔥 Left some comments to look at before the merge though

michalsmiarowski:
@kpyszkowski Let's also merge the newest changes from `main` branch to fix the build

src/components/SubNavigationPills/index.tsx
kpyszkowski: Seems like Prettier reorders imports alphabetically for me. I don't use any other code formatters. Perhaps my local config overlaps project-based Prettier rules. In my opinion such refinement is more like improvement rather than something i should revert. Nevertheless, for now I reverted it to keep the consistency but I highly encourage to refine imports globally in the future.

src/components/Modal/tBTC/NewTBTCApp.tsx
kpyszkowski: As above, ref: https://github.com/threshold-network/token-dashboard/pull/630#discussion_r1369785795

src/components/Modal/tBTC/NewTBTCApp.tsx
```diff
@@ -43,7 +45,16 @@ const NewTBTCAppBase: FC<BaseModalProps> = ({ closeModal }) => {
...
+      <ModalFooter gap="4">
+        <Button
+          variant="outline"
+          onClick={() => {
+            navigate("/tBTC/how-it-works")
+            closeModal()
+          }}
+        >
+          How it Works
+        </Button>
```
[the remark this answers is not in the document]
kpyszkowski: Applied, ref commit: b80ca38b868350c605df2389210a7708d0dc60ff

src/pages/tBTC/HowItWorks/index.tsx
```diff
@@ -27,7 +27,7 @@ const HowItWorksPage: PageComponent = (props) => {
 HowItWorksPage.route = {
   path: "how-it-works",
   index: false,
-  title: "How it Works",
+  title: "How it works",
```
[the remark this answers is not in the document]
kpyszkowski: Applied, ref commit: d8058d3e711ff7b4831245044ad2610382889593

kpyszkowski:
@michalsmiarowski The `main` branch has been merged into feature branch, ref commit: d17d9b3883c3b2f79ac6e183c664bad33ac507bb

src/components/SubNavigationPills/index.tsx
michalsmiarowski: > In my opinion such refinement is more like improvement rather than something i should revert

I agree, I think we could format it like this, but I like keeping the things clear and separate from each toher rather than doing everythin o half of it in one PR 😅 If we want to introduce such thing, then let's address it in a separate PR and change it accross the whole app.

src/components/Modal/tBTC/NewTBTCApp.tsx
```diff
@@ -11,18 +11,19 @@ import {
...
   Image,
 } from "@threshold-network/components"
-import { TakeNoteList } from "../../tBTC"
-import withBaseModal from "../withBaseModal"
-import tbtcAppBannerIllustration from "../../../static/images/tBTCAppBannerWithGrid.svg"
 import { useTBTCTerms } from "../../../hooks/useTBTCTerms"
+import tbtcAppBannerIllustration from "../../../static/images/tBTCAppBannerWithGrid.svg"
 import { BaseModalProps } from "../../../types"
+import ButtonLink from "../../ButtonLink"
+import { TakeNoteList } from "../../tBTC"
```
michalsmiarowski, line 18: Let's remove last point from `TakeNoteList` component as it in designs.

Designs:
image

Implementation:
image

michalsmiarowski (changes requested):
Left one small issue to fix before the merge

michalsmiarowski (approved):
LGTM 🔥

Issue #626, opened 2023-09-22
Dashboard listens for wrong AuthorizationIncrease events


cygnusv:
The consensus we had with @pdyraga and @vzotova was that, for authorization purposes, we'd look for events only in the Staking contract, since applications may have different internal semantics. See https://github.com/threshold-network/solidity-contracts/issues/88#issuecomment-1111059312

However, during the work for #603, we've detected that the dashboard listens for events on the application:
https://github.com/threshold-network/token-dashboard/blob/bc59c7ce2169ef9c376bb2086d556599970b8a87/src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts#L12-L27

Coincidentally, tBTC and RB app contracts define `AuthorizationIncreased` events with the exact signature than those from the staking contract, so for their particular case, listening for app events work. However, this is problematic for the TACo app contract that doesn't define them in the same way (which shouldn't be a problem)

The solution would be to listen for these events in the staking contract, as originally intended.

michalsmiarowski:
@cygnusv Based on @pdyraga [comment](https://github.com/threshold-network/solidity-contracts/issues/88#issuecomment-1111059312) it looks like we can also change the source of `AuthorizationDecreaseRequested` and `AuthorizationDecreaseApproved` events to `TokenStaking` contract.  The only problem is that in the dashboard for `AuthorizationDecreaseRequested` event we need the `decreasingAt` property, that is only passed from app's contract event, and is not present in `TokenStaking` contract event. We don't need the exact value, but it is needed to check if deauthorization request is active (see https://github.com/threshold-network/token-dashboard/blob/5875b44992cbe0c24439d227c6ca89e0568b11ca/src/store/staking-applications/slice.ts#L208-L219).

Do you know if it's somehow possible to get that infrormation from `TokenStaking` contract? 

For now @kpyszkowski has only changed the `AuthorizationIncrease` event as it was stated in the issue.

cygnusv:
> Do you know if it's somehow possible to get that infrormation from `TokenStaking` contract?

I don't think so (although please @vzotova correct me if I'm wrong), since that information seems specific to the sortition pool, not the staking contract. I see 2 alternatives:
[...]

PR #636, opened 2023-10-03
Update contract for AuthorizationIncreased event

kpyszkowski:
Ref: https://github.com/threshold-network/token-dashboard/issues/626

tBTC and RB app contracts define `AuthorizationIncreased` events with the exact signature than those from the staking contract, so for their particular case, listening for app events work. However, this is problematic for the TACo app contract that doesn't define them in the same way. For the sake of consistency contract-providing hook (`useStakingAppContract` to `useTStakingContract`) has been replaced. Moreover `getStakingAppLabel` utility set has been refactored with unit tests included.


kpyszkowski:
@michalsmiarowski please let me know if we should update contracts for remaining authorization-related events

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
```diff
@@ -2,14 +2,16 @@ import {
...
+  useTStakingContract,
+} from "../../web3/hooks"
 import { useAppDispatch } from "../store"
-import { useStakingAppContract } from "./useStakingAppContract"
 
 export const useSubscribeToAuthorizationIncreasedEvent = (
   appName: StakingAppName
 ) => {
-  const contract = useStakingAppContract(appName)
+  const contract = useTStakingContract()
```
michalsmiarowski, line 14: Since we remove the `appName` here then I think we should also remove `appName` from the arugments of `useSubscribeToAuthorizationIncreasedEvent` hook. The `appName` that is needed for dispatching the `authorizationIncreased` below should be extracted from the event here:

https://github.com/threshold-network/token-dashboard/blob/89da14d7b6faee081498f88ddd721f129721ee89/src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts#L19

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
michalsmiarowski, line 1: We pass different arguments to `TokenStaking` event comparing to the event in the app contract:

Token Staking event:
https://github.com/threshold-network/solidity-contracts/blob/47268908a1aa92ab215b76e5f26730519643b663/contracts/staking/TokenStaking.sol#L130-L134

App event (in this case it's random beacon):
https://github.com/keep-network/keep-core/blob/742e3739a87852e6bb76e1b042909d927150a42e/solidity/random-beacon/contracts/RandomBeacon.sol#L316-L321

Those changes should be reflected here:
https://github.com/threshold-network/token-dashboard/blob/89da14d7b6faee081498f88ddd721f129721ee89/src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts#L19

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
```diff
@@ -2,14 +2,16 @@ import {
...
 } from "../../store/staking-applications"
-import { useSubscribeToContractEvent } from "../../web3/hooks"
+import {
+  useSubscribeToContractEvent,
+  useTStakingContract,
+} from "../../web3/hooks"
 import { useAppDispatch } from "../store"
-import { useStakingAppContract } from "./useStakingAppContract"
 
 export const useSubscribeToAuthorizationIncreasedEvent = (
```
michalsmiarowski, line 11: If you apply the changes that I've mentioned above then you should aldo remove one of the hook call in `App.tsx`:

https://github.com/threshold-network/token-dashboard/blob/89da14d7b6faee081498f88ddd721f129721ee89/src/App.tsx#L71-L72

michalsmiarowski (changes requested):
Left some comments to look at down below 👇 

Please also test it out and see if the data returned from events after you increase the authorization is similar to the data returned from the events after your changes 😉

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
kpyszkowski: done

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
kpyszkowski: i don't understand what do you mean
i suppose there is a difference between `operator` and `application`?

michalsmiarowski: Yes, those are two different things. `operator` is the address, related to the application (random beacon or tbtc), that you can map to your stake's staking provider address. `application` is the address of the application contract (random beacon or tbtc) on the chain. In the actual code we don't use `operator` address anywhere, but we need the `application` address to get the application name and pass it to `authorizationIncreased` action few lines below.

You just have to change `operator` to `application` here and change that address to an app name using `stakingAppAddressToName` function.

kpyszkowski: done 🙇🏼‍♂️

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
```diff
@@ -2,21 +2,35 @@ import {
...
+const getApplicationName = (address: string) => {
+  const { multiAppStaking } = threshold
+  const namesDictionary = Object.fromEntries(
+    Object.entries(multiAppStaking).map(([name, { address }]) => [
+      address,
+      name === "ecdsa" ? "tbtc" : name,
+    ])
+  )
+  return namesDictionary[address]
+}
```
michalsmiarowski, line 21: This works good, but I think we can refactor it and put it inside [getStakingAppLabel.ts](https://github.com/threshold-network/token-dashboard/blob/main/src/utils/getStakingAppLabel.ts).

First of all it would be good to refactor whole `getStakingAppLabel.ts` file as it seems the function names there are a little confusing. We have to distinguish the staking app name from the staking app label. Staking app name is the name of the application that we use in our redux store to get the data ("randomBeacon", "tbtc"). Staking app label is the label for a given application that we use on front-end ("Random Beacon", 'tBTC").

My proposition for refactoring:
- `stakingAppAddressToName` should return app name instead of the label. Let's also rename it to `stakingAppAddressToAppName`
- `getStakingAppNameFromAddress` should also return the app name instead of the label and it should use the name returned from `stakingAppAddressToAppName` function. The `getStakingAppNameFromAddress` will be the function that you will use in `useSubscribeToAuthorizationIncreasedEvent` event. I think we should also rename it to `getStakingAppLabelFromAppAddress`.
- create a new function `getStakingAppLabelFromAppAddress` that will return either staking label from `getStakingAppLabelFromAppAddress` function or just "App" string
- rename `getSakingAppLabel` to `getStakingAppLabelFromAppName` function

In a separate commit you could also write some tests for that and put them in `stc/utils/__tests__`.

michalsmiarowski (changes requested):
Works good, but left some comments down below with the proposition for refactor

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
kpyszkowski: done

src/utils/getStakingAppLabel.ts
```diff
@@ -6,16 +6,22 @@ const stakingAppNameToAppLabel: Record<StakingAppName, string> = {
   randomBeacon: "Random Beacon",
 }
 
-const stakingAppAddressToName: { [key: string]: string } = {
-  [threshold.multiAppStaking.ecdsa.address]: stakingAppNameToAppLabel.tbtc,
-  [threshold.multiAppStaking.randomBeacon.address]:
-    stakingAppNameToAppLabel.randomBeacon,
+const stakingAppAddressToAppName: Record<string, StakingAppName> = {
+  [threshold.multiAppStaking.ecdsa.address]: "tbtc",
+  [threshold.multiAppStaking.randomBeacon.address]: "randomBeacon",
```
cygnusv, line 11: @theref 👀 🌮

src/hooks/staking-applications/useSubscribeToAuthorizationIncreasedEvent.ts
```diff
@@ -1,22 +1,22 @@
...
+} from "../../web3/hooks"
 import { useAppDispatch } from "../store"
-import { useStakingAppContract } from "./useStakingAppContract"
 
-export const useSubscribeToAuthorizationIncreasedEvent = (
-  appName: StakingAppName
-) => {
-  const contract = useStakingAppContract(appName)
+export const useSubscribeToAuthorizationIncreasedEvent = () => {
+  const contract = useTStakingContract()
```
cygnusv, line 10: yay!

cygnusv (commented):
Great seeing this fix. Thanks, friends.

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  it("returns correct app label if app name is given", () => {
+    const resultTbtcLabel = getStakingAppLabelFromAppName(tbtcName)
+    const resultRbLabel = getStakingAppLabelFromAppName(rbName)
+
+    expect(resultTbtcLabel).toBe(MOCK_LABELS[tbtcName])
+    expect(resultRbLabel).toBe(MOCK_LABELS[rbName])
+  })
+
+  it("returns correct app name if address is given", () => {
+    const resultTbtcName = getStakingAppNameFromAppAddress(tbtcAddress)
```
michalsmiarowski, line 41: Let's also write a test to cover the case when you pass an address that is different than `tbtcAddress` and `rbAddress`

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const MOCK_ADDRESSES: Record<StakingAppName, string> = {
+  tbtc: getThresholdLib().multiAppStaking.ecdsa.address,
+  randomBeacon: getThresholdLib().multiAppStaking.randomBeacon.address,
+}
+const MOCK_LABELS: Record<StakingAppName, string> = {
+  tbtc: "tBTC",
+  randomBeacon: "Random Beacon",
+}
+const MOCK_APP_NAMES: StakingAppName[] = ["tbtc", "randomBeacon"]
```
michalsmiarowski, line 17: Let's rename those mock variable to camelCase

michalsmiarowski (changes requested):
Looking good. Added just few more comments here

src/utils/getStakingAppLabel.ts
```diff
@@ -6,16 +6,22 @@ const stakingAppNameToAppLabel: Record<StakingAppName, string> = {
...
+export const getStakingAppNameFromAppAddress = (stakingAppAddress: string) => {
+  return stakingAppAddressToAppName[stakingAppAddress]
 }
 
-export const getSakingAppLabel = (stakingAppName: StakingAppName) => {
-  return stakingAppNameToAppLabel[stakingAppName]
+export const getStakingAppLabelFromAppName = (
+  stakingAppName: StakingAppName
+) => {
+  return stakingAppNameToAppLabel[stakingAppName] || "App"
```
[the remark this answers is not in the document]
kpyszkowski: correct, removed

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+describe("Staking app label utils tests", () => {
+  const [tbtcName, rbName] = MOCK_APP_NAMES
```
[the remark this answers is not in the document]
kpyszkowski: updated 🙇🏼

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  const tbtcAddress = MOCK_ADDRESSES[tbtcName]
+  const rbAddress = MOCK_ADDRESSES[rbName]
```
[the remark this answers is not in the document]
kpyszkowski: updated 🙇🏼

src/utils/__tests__/getStakingAppLabel.test.ts
kpyszkowski: updated 🙇🏼

src/utils/__tests__/getStakingAppLabel.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it("returns correct app label if app address is given", () => {
+    const resultTbtcLabel = getStakingAppLabelFromAppAddress(tbtcAddress)
+    const resultRbLabel = getStakingAppLabelFromAppAddress(rbAddress)
```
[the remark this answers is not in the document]
kpyszkowski: added

src/utils/__tests__/getStakingAppLabel.test.ts
kpyszkowski: since this fallback has been removed, such test seems to be redundant
I moved fallback value to the `getStakingAppLabelFromAppAddress` function level

michalsmiarowski (approved):
LGTM 🔥

PR #639, opened 2023-10-04
Update T coverage-pools dependency

michalinacienciala:
In the `package.json` there are two different versions of `coverage-pools` used. This is done to allow using contracts from two different cov pool versions on the mainnet (needed for computations of the TVL-related fields) - one relates to KEEP token cov pool (<2.0.0) and one to to T token one (>=2.0.0). On the testnet environmet we currently don't have the T token cov pool deployed. The latest `goerli`- tagged version of the cov-pools package (`1.1.0-goerli.0`) contains KEEP-token related contracts. We want to stick to this version (we'll need to update it once we switch to Sepolia). For the `@threshold-network/coverage-pools` ideally we'd want to use a T token related pool, which if deployed for given environment, should be present in the latest `goerli`/`sepolia` package. If the T-related pool is not yet deployed for given environment, KEEP-related cov-pool contracts will be used in all TVL computations (which differs from the mainnet behavior, but is acceptable on the test env).

Ref: https://github.com/threshold-network/token-dashboard/pull/613


.github/workflows/reusable-build-and-publish.yml
```diff
@@ -113,7 +113,6 @@ jobs:
           query: |
             threshold-contracts-version = github.com/threshold-network/solidity-contracts#version
             random-beacon-contracts-version = github.com/keep-network/keep-core/random-beacon#version
-            ecdsa-contracts-version = github.com/keep-network/keep-core/ecdsa#version
             tbtc-v2-contracts-version = github.com/keep-network/tbtc-v2#version
             tbtc-v2-ts-version = github.com/keep-network/tbtc-v2.ts#version
```
michalsmiarowski, line 118: In [SDK 2.0.0 migration PR](https://github.com/threshold-network/token-dashboard/pull/654/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L23-R22) we've also removed `@keep-network/tbtc-v2` package, and updated the `tbtc-v2.ts` to `2.3.0` on all chains (it no longer needs `goerli`/`development` tags).

Is that something we should also address here? Do you think we should remove `tbtc-v2.contracts-version` from here? Not sure about `tbtc-v2-ts-version` though.

michalsmiarowski (changes requested):
Left one comment to look at before the merge

.github/workflows/reusable-build-and-publish.yml
michalinacienciala: Oh, right, I missed that somehow. I've amended the commit, see [ce23262](https://github.com/threshold-network/token-dashboard/pull/639/commits/ce232625d16fc938a11520d0027b756b4e6d7576).

michalsmiarowski (approved):
LGTM 🔥

PR #641, opened 2023-10-04
tBTC only mode

r-czajkowski:
This PR adds support for building the T dapp in tBTC v2 only mode- it means that all non-tBTC-related features will be hidden. Here we add a new feature flag that renders only tBTC features and we also create a PoC for creating a theme for various chains.

### How to add a new chain theme
All chain themes are live under `src/theme/chains/<chain-name>` and should be built based on the best practices defined in [the Chakra docs](https://v1.chakra-ui.com/docs/styled-system/theming/customize-theme#scaling-out-your-project). Thanks to this we can easily override colors, components styles and semantic tokens. The `getChainTheme` function is responsible for returning the correct theme based on the `CHAIN_ID` env variable.

Here we also introduce chakra semantic tokens that allow to define a theme for each chain to customize the UI. Using semantic tokens from chakra we can define tokens within the color theme namespace that can reference existing color values - we have abstracted specific tints of colors into named tokens that we can easily tweak from the theme file and implement anywhere else in our app. As an example, I've added a `mainBackground` semantic token for which we can override a color in chain theme without conditionally updating the `bg` prop inside the component.

### Caveats
This solution should work very well if the UI and flow for other chains will be almost the same and we only want to change "colors" since we can easily extract common parts to semantic tokens and render some different widgets based on the `CHAIN_ID` env variable. The most difficult parts are the different behaviors and different components for a chain. We currently don't know any details about the UI and flow, so it looks like this approach is a good first step to building tBTC app for other chains.

### TODO:
- [ ] How to build links to block explorer for a given chain? We probably should be more abstract and instead of [`ethereum` type](https://github.com/threshold-network/token-dashboard/blob/main/src/components/ViewInBlockExplorer.tsx#L10) we should rename to `evm` (meaning we want to build a link for an EVM compatible chain). Mostly all evm block explorers are built on top of Etherscan.
- [ ] Update Multicall contract addresses for new chains. We will probably need to upgrade to Multicall3 since the version that we currently use is no longer maintained and the Multicall contract does not exist on e.g. Base chain. See [Multicall3](https://github.com/mds1/multicall); [Multicall3 addresses](https://github.com/mds1/multicall#existing-deployments).
- [ ] How to deploy dapps for other chains? We can use [a matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) to automatically create multiple job runs that are based on the combinations of the variables.
- [ ] Implement services defined in `threshold-ts` dir if needed. E.g Staking is only supported on `Ethereum` so we should probably implement the `IStaking` interface and implement all necessary methods to return empty `arrays`,`string` or just  `0`.


lukasz-zimnoch:
This is an exploratory work that was supposed to check one of the possible paths. Closing in favor of: https://github.com/threshold-network/pm/issues/41

PR #642, opened 2023-10-05
Fix `ecpair` lib error with babel plugin

michalsmiarowski:
After updating the `tbtc-v2.ts` lib with new version for `goerli`, we've ecnoutered a problem running dApp:

```
Failed to compile.

./node_modules/ecpair/src/ecpair.js 82:7
Module parse failed: Unexpected token (82:7)
File was processed with these loaders:
 * ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|
|   class ECPair {
>     __D;
|     __Q;
|     compressed;
```

As you an see it is related to the [ecpair](https://github.com/bitcoinjs/ecpair) lib that we've installed in `tbtc-v2.ts` and it has something to do with our `webpack`. As a solution I've installed a babel plugin - `@babel/plugin-transform-class-properties` - and added it to the craco config with the `loose` option.

You can see more information about this plugin in the babel docs: https://babeljs.io/docs/babel-plugin-transform-class-properties.

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -3,6 +3,7 @@ module.exports = {
     plugins: [
       "@babel/plugin-proposal-nullish-coalescing-operator",
       "@babel/plugin-proposal-optional-chaining",
+      ["@babel/plugin-transform-class-properties", { loose: true }],
     ],
   },
 }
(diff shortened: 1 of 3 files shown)


michalsmiarowski:
This issue will be resolve in https://github.com/threshold-network/token-dashboard/pull/652 (see 890b3e039f8445a66c82a994600b3e938da23f13). Closing this PR...

PR #656, opened 2023-11-02
Rewards allocation for Nov 1st 2023

manumonti:
This PR adds a merkle tree computed in https://github.com/threshold-network/merkle-distribution/pull/94 to Threshold Token Dashboard.


michalsmiarowski (approved):
LGTM 🔥

PR #663, opened 2023-11-21
Inadequate network parameter for address validation utility function

kpyszkowski:
There is an error with redemption validation function caused by the network parameter which fallbacks to mainnet on a testnet when is not provided.


michalsmiarowski (approved):
LGTM

Issue #648, opened 2023-10-16
Switch dApp to SDK `v2.0.0`


lukasz-zimnoch:
Depends on: https://github.com/threshold-network/token-dashboard/issues/647 and SDK `typescript/v2.0.0` release.

We need to make the same adjustments as in https://github.com/threshold-network/token-dashboard/issues/647 but for the `typescript/v2.0.0` release. This translates to the following task list:

```[tasklist]
### Tasks
- [x] Change `package.json` to use `@keep-network/tbtc-v2.ts@v2.0.0` (one version for all networks as chain configs are embedded now). We should probably use the caret `^` sign to allow upgrades to future compatible versions
- [x] Make sure `yarn.lock` reflects that change
- [x] Make all necessary code adjustments to make the dApp build successful (initialize the SDK properly for the given environment)
- [x] Make sure the dApp does no longer depend on `@keep-network/tbtc-v2` and `@keep-network/ecdsa` packages directly. All contract interactions should be handled through SDK
- [x] Make necessary changes in CI jobs
- [x] Make basic regression tests on `https://dashboard.test.threshold.network`
- [x] Make basic regression tests on mainnet preview
- [x] Once this task is done. Release the new dApp on mainnet
```

```[tasklist]
### PRs
- [ ] https://github.com/threshold-network/token-dashboard/pull/654
- [ ] https://github.com/threshold-network/token-dashboard/pull/662
- [ ] https://github.com/threshold-network/token-dashboard/pull/663
- [ ] https://github.com/threshold-network/token-dashboard/pull/664
```

PR #664, opened 2023-11-21
Stringify deposit telemetry parameters

lukasz-zimnoch:
Refs: https://github.com/threshold-network/token-dashboard/issues/648

SDK was updated so deposit parameters are now `Hex`. To properly display them in the Sentry telemetry hub, they must be converted to a string first.


src/hooks/tbtc/useDepositTelemetry.ts
```diff
@@ -15,10 +15,22 @@ export const useDepositTelemetry = (network: BitcoinNetwork) => {
...
+        walletPublicKeyHash,
+        refundPublicKeyHash,
+        refundLocktime,
+      } = deposit
+
       captureMessage(
         `Generated deposit [${depositAddress}]`,
         {
-          ...deposit,
+          depositor: depositor.identifierHex,
```
michalsmiarowski, line 29: Are we sure we want to pass `identifierHex` as a depositor here? It looks like we passed a whole depositor object (of a type `Idenfitier`) in the old version of the SDK. It's probably not big of a deal anyway, but just making sure, because the parameters for new deposits will look slighty different (in compared to the actual ones) in sentry because of that.

image

michalsmiarowski (changes requested):
Left one comment to look at before the merge

src/hooks/tbtc/useDepositTelemetry.ts
lukasz-zimnoch: Yeah, we don't need to pass the entire object.

michalsmiarowski (approved):
LGTM

PR #671, opened 2023-11-29
Improve Readme

michalsmiarowski:
Main changes:
- Removes `@keep-network/ecdsa@sepolia` and `@keep-network/tbtc-v2@sepolia` from `## Install Sepolia contracts` command, because we don't use those libs anymore,
- Removes outdated `# Local development` section,
- Adds `Sepolia-dev contracts` subsection


kpyszkowski (approved):
Approved

PR #697, opened 2023-12-13
Correct the BTC amount displayed in initiate mint

michalsmiarowski:
The wei amount, that user send to deposit address, was wrongly converted to BTC amount in InitiateMinting component. The value was converted to token amount as if it was wei instead of satoshi. Because of that it was converted to a really small value.


kpyszkowski (approved):
Approved

PR #651, opened 2023-10-24
tBTC minting redesign - Phase I - part 1/4

kpyszkowski:
Ref: https://github.com/threshold-network/token-dashboard/issues/590

I've accomplished the first part of ongoing changes in Phase I. The current progress status and PR scope are presented in this [objectives checklist](https://github.com/threshold-network/token-dashboard/issues/590#issuecomment-1755270349).
I've also left some comments on Figma to inform and explain about the unapplied changes I will implement in future parts/phases.

PR #698, opened 2023-12-13
Fix transaction modal not disappearing bug

michalsmiarowski:
There was a bug that prevent transaction status modal to be closed after successful deposit reveal. It happened because in #651 we moved the logic from `TbtcMintingConfirmation` to `InitiateMinting` component, but we didn't include `closeModal()` call that was in `onSuccessfulDepositReveal` method. This PR fixes that.

Besides that I've improved the `useEffect` in `InitiateMinting` componen (see 16451b8b9c520a600c0203366153fda9951b1c19).


kpyszkowski (approved):
Approved

PR #705, opened 2023-12-19
Update @ledgerhq/connect-kit-loader dependency

michalsmiarowski:
Update the `@ledgerhq/connect-kit-loader` dependency to `1.1.8` so that it is tethered to version 1.1.8 of the connect-kit.


cygnusv:
Noting that v1.1.8 is supposed to be safe according to Ledger:
https://support.ledger.com/hc/en-us/articles/15480255246365-Malicious-version-of-the-Ledger-Connect-Kit-Live-thread?support=true

Issue #668, opened 2023-11-28
Remove coverage pool references from the dapp


erdogan:
```[tasklist]
### Pull Requests
- [ ] https://github.com/threshold-network/token-dashboard/pull/709
```

PR #709, opened 2023-12-20
Remove all references to coverage pools

michalsmiarowski:
Closes: #668 

Removes all references to coverage pools including adding it's tvl to the total tvl value. Thanks to this we can remove `@keep-network/coverage-pools` and `@threshold-network/coverage-pools` libs.

Note: `useFetchTvl.test.jsx` test is not working after our tbtc-v2.ts 2.3 integration. All non-working tests should be improved in a separate Pull Request.


kpyszkowski (approved):
Looks good to me 🔥

michalinacienciala (approved):
Looks good from the CI point of view. There's one more thing we'll need to do outside of `token-dashboard` repository that's a consequence of this change - remove the `coverage-pools` from the [CI flow config](https://github.com/keep-network/ci/blob/main/config/config.json) (basically revert https://github.com/keep-network/ci/pull/35).  I'll do that. We don't need to wait with merging current PR for that change.

PR #702, opened 2023-12-19
Add No Bridge Fees annoucement banner.

michalsmiarowski:
Ref: #695 

Adds No Bridge Fees to Overview page and TBTC's How it Works page.

## Screenshots
<details>
  <summary>Normal view</summary>
  
image

</details>

<details>
  <summary>Tablet view</summary>
  
image

</details>

<details>
  <summary>Mobile view</summary>
  
image

</details>

Issue #695, opened 2023-12-07
Add a notification in the dashboard for fee holiday


erdogan:
In the dashboard, display a note for the fee holiday (could be an info bar on top or a modal or a callout in the mint container) indicating our fee holiday. Suggestged text (with a holiday spirited festive image if there is space) 

"Ho ho ho! No bridge fees until February 10, 2024!"
```[tasklist]
### Pull Requests
- [ ] https://github.com/threshold-network/token-dashboard/pull/702
- [ ] https://github.com/threshold-network/token-dashboard/pull/711
```

SorinQ:
Hi @michalsmiarowski Please find the design of the banner here:
https://www.figma.com/file/OUdSfHsmgV8EJpWxAzXjl5/Design-Tasks?type=design&node-id=3865-803&mode=design&t=AqhvhadPd2C1JdlB-4

Sample:
<img width="1040" alt="Final" src="https://github.com/threshold-network/token-dashboard/assets/37292909/17d552ea-d1d3-4cbb-ac3c-aaa70e37f375">

PR #711, opened 2023-12-20
No bridge fees improvements

michalsmiarowski:
Ref: #695 
Related PR: #702 

Fixes padding for NoBridgeFees banner for mobile view. For that I had to change how we apply the theme in `AnnoucementBanner` (`__css` prop instead of `sx`) prop so that we can override the styles with style props.

For standard website view the padding remains the same.

Besides that, it adds a badge to tbtc website title
image

For dark mode:
image

TODO:
- [x] Add text informing about the fees to the Bridge page (cc @SorinQ )
- [x] Remove banner from tbtc's How It Works page once the above is applied


SorinQ:
@michalsmiarowski Is there a reason why we changed the font on the banner?

src/static/images/no-bridge-fees-text.svg
michalsmiarowski, line 1: > @michalsmiarowski Is there a reason why we changed the font on the banner?

@SorinQ Creating a github conversation for that. Which font exactly? You mean `No bridge fees until February 10, 2024!`?

SorinQ: we had this version version: https://www.figma.com/file/OUdSfHsmgV8EJpWxAzXjl5/Design-Tasks?type=design&node-id=3865-803&mode=design&t=Uzhz3zbnTlaeUeuH-4

michalsmiarowski: Ah, I see. I though that this is a new version: https://www.figma.com/file/OUdSfHsmgV8EJpWxAzXjl5/Design-Tasks?type=design&node-id=3865-762&mode=design&t=3TKqETEdG4yA8cBk-0, because I did not see those banner last time I've checked 😅 Didn't notice that Engin comments was from last week.

I'll revert those changes then.

Thank you!

SorinQ: np. sorry you need to work on this again :(

michalsmiarowski: No worries, it's jsut a matter to reverting the code to the previous state.

Fixed in 32aeeead8ec6db0638ac4c0d20386b62386cd1f0

src/components/AnnouncementBanner/index.tsx
```diff
@@ -45,7 +45,7 @@ export const AnnouncementBannerContainer: FC<
   })
 
   return isOpen ? (
-    <Card sx={styles.container} {...props}>
+    <Card __css={styles.container} {...props}>
```
kpyszkowski, line 48: The use of `__css` prop breaks the specificity of component's styles. Now the tBTC's `AnnouncementBanner` instance has visual errors: 
image

Also the use of `AnnouncementBanner` component as `NoBridgeFeesBanner`’s container seems to be redundant. In fact it's a regular `Card` component without the border (which can be easily disabled with the `borderWidth` prop.

I suggest to revert changes in this line and update the `NoBridgeFeesBanner` to use `Card` component as a container with the border width set to `0`.

Some readings about the `__css` vs `sx` props I've researched: https://github.com/chakra-ui/chakra-ui/discussions/5182
As far as I understand the `__css` prop is the renamed [Emotion's css prop](https://emotion.sh/docs/css-prop) to avoid API naming conflicts and it's used by the Chakra under the hood. In 99% cases the way to go is Chakra's `sx` prop.

src/pages/Overview/NoBridgeFeesBanner.tsx
```diff
@@ -14,16 +14,16 @@ export const NoBridgeFeesBanner: FC<AnnouncementBannerContainerProps> = ({
   return (
     <AnnouncementBannerContainer
```
kpyszkowski, line 15: As suggested in <ID HERE>:

```suggestion
    <Card
        borderWidth={0}
```

Remember to update props typings - you can use `BoxProps` since the `Card` components is a wrapper for the `Box` component and doesn't export the `CardProps` type itself.

As a side note - perhaps the box shadow should be disabled as well. There is an inconsistency in this area between the implementation and Figma designs.

src/static/images/page-title-badge.png
kpyszkowski, line 1: This image looks bad on the dark mode:
image

The image itself tends to be blurry on some viewports. For such simple shapes I'd recommend to export is as an SVG. Please remember to optimize the file.

@SorinQ please update designs to fix the issue. Now in Figma colors of this element are solid, perhaps you should use alpha channel to make it semi transparent 🤷🏼

kpyszkowski (commented):
Left some comments:

src/components/AnnouncementBanner/index.tsx
michalsmiarowski: Thanks, that's a great suggestion and a great reading you provided 🔥 

This is also why I've changed it in the first place. Basicaly "the problem" was that since `sx` has the highest precedence (which, I assume, means the styles passed there can't be overwritten by anything else) the props desctructured from `{...props}` were not applied if the same ones were passed to `sx` (so in our case - in the `AnnouncementBanner` theme, because we pass the props from that theme to `sx` prop).

However, I was not aware, that when we pass it through `__css` property, then the `Card` theme from `@threshold-network/components` won't be applied 😅 It probably means that `sx` property is theme-aware, while `__css` prop is not.

To further illustrate the issue: we use multiStyles for `AnnoucementBanner`: 

```
const baseStyle: PartsStyleFunction<typeof parts> = (props) => {
  return {
    container: {
      bg: "black",
    },
    (...)
```

Now let's say we are using a container like this:
```
const styles = useMultiStyleConfig("AnnouncementBanner", {
  size,
  variant,
})

return <Card sx={styles.container} bg="red" />
```

This will render a Card with a black background. This is problematic, because if we use `{...props}` in the place of `bg="red"`, it won't override the properties that are defined in the theme. Two quick solutions that came to my mind when implementing it were:

```
const styles = useMultiStyleConfig("AnnouncementBanner", {
  size,
  variant,
})

return <Card sx={{...styles.container, ...props}} />
```

or changing the `sx` to `__css` prop. I've used the second one, but all of it were unnecessary anyway, because in my final implementation I've used `backgroundImage` instead of `bg` which was not overwritten in the theme. 

Neverethless, it was a very good catch! If you think I've messed something up there then feel free to point it up, because I migh have not understand how chakra exactly works here.

I've applied your solution in bc56dbc1035d2e36466ed003b51bee3d6bfdebb4

src/pages/Overview/NoBridgeFeesBanner.tsx
michalsmiarowski: Thanks. Implemented a solution in bc56dbc1035d2e36466ed003b51bee3d6bfdebb4. I've also removed the box shadow.

Ref: https://github.com/threshold-network/token-dashboard/pull/711#discussion_r1437094324

src/static/images/page-title-badge.png
michalsmiarowski: Replaced PNG with SVG in a84accd520bbf3c5db79a5528440dcf3165f344f.

Need @SorinQ here for dark mode design.

michalsmiarowski: As a workaround I've fixed it for dark mode by myself in a9e7b441f6d16106fc0a12cd042619644132657f

image

We can re-design it later when designs are ready.

kpyszkowski (approved):
Looks good to me 🔥

kpyszkowski (approved):

PR #719, opened 2024-01-02
Upgrade dependencies in CI with `--ignore-scripts` flag

michalinacienciala:
Not long ago we've changed the dapp's `@keep-network/tbtc-v2.ts` dependency to `^2.3.0`, which results in the introduction of the `@threshold-network/solidity-contracts@1.2.1` package in the `yarn.lock`. That package contains a `prepare-dependencies.sh` script which sometimes causes problems during install/upgrade of the dapp dependencies. As a workaround we can install/upgrade the dashboard dependencies with `--ignore-scripts` flag. This way we don't run the misbehaving script (we don't need to run it in this project). I already added the `--ignore-scripts` flag in a couple of places in an earlier PR, but I didn't add it to the `yarn upgrade` command, as I thought that postinstall scripts are not executed during `yarn install`. But lately I realized that this does not apply to postinstall scripts in project's upgraded dependencies. Hence we need to add the flag to the upgrade command.


theref (approved):

michalsmiarowski (approved):
LGTM 🔥

Issue #730, opened 2024-02-05
Install Google Tag Manager


lukasz-zimnoch:
This task is about installing Google Tag Manager in the production instance of the Threshold Dashboard, according to the [Set up and install Tag Manager > 2. Install the container](https://support.google.com/tagmanager/answer/6103696) guide. Relevant code snippets are already generated and can be found [here](https://discord.com/channels/893441201628405760/1022857346789163099/1202567592376533084). Google Tag Manager **SHOULD NOT** be enabled for local and testnet instances.
```[tasklist]
### Pull Requests
- [ ] https://github.com/threshold-network/token-dashboard/pull/733
- [ ] https://github.com/threshold-network/website/pull/112
```

lukasz-zimnoch:
Let's also do the same for Threshold website. We should use the same project ID.

PR #733, opened 2024-02-16
Implement Google Tag Manager 🏷️ 

michalsmiarowski:
Ref: #730 

Implements Google Tag Manager into the Threshold Dashboard. 

For this, I've created a separate `useGoogleTagManager` hook that appends the code snippets from GTA into the DOM. For this to work properly we have to provide `GOOGLE_TAG_MANAGER_ID` as secret through `.env` file. I've also created a feature flag for it so that we know for sure it is fired only on production.


lukasz-zimnoch (approved):
LGTM!

Issue #749, opened 2024-04-11
Bug caused by racing conditions in sdk initialization hook


evandrosaturnino:
A bug was found in `useInitializeTbtcSdk` hook that causes a racing condition.

Video that illustrate the issue:

https://github.com/threshold-network/token-dashboard/assets/60587527/19a0594c-6468-4096-816c-c3137ef54baf

In order to reproduce the issue, the app should be refreshed and the wallet connected right after (Before the sdk without signer initialization process concludes), then you will see the "Connect Wallet" change to "Upgrade" on connection, then regressing to "Connect Wallet" after a few seconds.

The observed issue occurs because the SDK initializes at the launch of the decentralized application (dApp), and this initialization process is slower without a signer than with one. Specifically, when an account is connected, initialization with a signer is faster. However, if an account gets connected before the SDK has completed its initial setup, the SDK object reverts to the state it would be in if initialized without a signer.

michalsmiarowski:
Here is a conversation for more context: https://github.com/threshold-network/token-dashboard/pull/736#discussion_r1556100655

PR #751, opened 2024-04-18
Improve tBTC SDK initialization

michalsmiarowski:
Closes: #749 

Here we use lazy loading for tbtc sdk initialization in our threshold-ts lib. Instead of keeping the sdk object in `_sdk` property we actually store the promise itself in `_sdkPromise` property. This promise is set during the construction of the SDK class and is meant to be resolved once the SDK is fully initialized. The promise is used to ensure that the SDK is initialized only once and that all subsequent retrievals of the SDK instance await this promise, thereby ensuring that the SDK is ready for use before any operations are performed.

The promise also facilitates updating the SDK instance dynamically: if a new initialization is triggered (e.g., when a user logs in and provides a new signer), it ensures that the new SDK object replaces any previously pending SDK object, even if the original promise has not yet resolved.

To retrieve the SDK object, use the `getSdk()` method. This method will either return the already resolved SDK object or wait until the promise resolves before returning the SDK.

Additionally I've changed the `getTransactionConfirmations` method implementation - it does not have to use SDK. The `IsSdkInitializingContext` is also not neede anymore so I've removed it completely. 

### Further explanation of the flow (based on https://github.com/threshold-network/token-dashboard/pull/751#discussion_r1575410420):

1. Instead of storing sdk object in a property and add some helping variables (like `isSdkInitalized`) that will tell if the sdk is initialized or not, we now store the promise itself:

```
  /**
   * Holds the promise for the asynchronously initialized SDK instance.
   * This promise is set during the construction of the SDK class and is meant
   * to be resolved once the SDK is fully initialized. The promise is used to
   * ensure that the SDK is initialized only once and that all subsequent
   * retrievals of the SDK instance await this promise, thereby ensuring that
   * the SDK is ready for use before any operations are performed.
   *
   * The promise also facilitates updating the SDK instance dynamically: if a
   * new initialization is triggered (e.g., when a user logs in and provides a
   * new signer), it ensures that the new SDK object replaces any previously
   * pending SDK object, even if the original promise has not yet resolved.
   *
   * To retrieve the SDK object, use the `getSdk()` method. This method will
   * either return the already resolved SDK object or wait until the promise
   * resolves before returning the SDK.
   */

  private _sdkPromise: Promise<SDK | undefined>
```

2. We retrieve the SDK object by using `_getSDK` method which will wait for the promise mentioned above to be resolved:

```
private _getSdk = async (): Promise<SDK> => {
    const sdk = await this._sdkPromise
    if (!sdk) throw new EmptySdkObjectError()

    return sdk
  }
```

So, there are few cases that can happen:

a) User interacts with tbtc contract but the SDK is not initialized yet:

- in such cases we use `_getSDK` method so the code will wait until `_sdkPromise` is resolved and then proceeds with the code

b) User interacts with tbtc contract and SDK is already initialied:

- this means that the promise is already resolved so we will retudn the sdk object immidietaly from `this_.sdkPromise` (no need to wait for it being initialized because it already is, and the promise is already resolved)


src/components/SubmitTxButton.tsx
```diff
@@ -15,14 +14,13 @@ const SubmitTxButton: FC<Props> = ({
...
   const { isActive } = useIsActive()
-  const { isSdkInitializedWithSigner } = useIsTbtcSdkInitializing()
   const connectWallet = useConnectWallet()
 
   const onConnectWalletClick = () => {
     connectWallet()
   }
 
-  if (isActive && isSdkInitializedWithSigner) {
+  if (isActive) {
```
evandrosaturnino, line 23: @michalsmiarowski Please correct me if i'm wrong, but I assume that if the user connect his wallet soon after the app refreshes and tries to interact with a tBTC contract, not giving enough time to initialize the SDK it won't work as expected, possibly returning an error, right?
From what I look the `isActive` validation is being get from the `useWeb3React()` method, which validates if a wallet is actively connected, so AFAIK it doesn't validate whether the tBTC SDK is initialized or not. 

Do we want to add some flow here in order to improve the user experience?

evandrosaturnino (approved):
LGTM. Just a small question on whether it makes sense for us to create some flow to improve the UX for those buttons which interacts with tBTC contract.

src/components/SubmitTxButton.tsx
michalsmiarowski: @evandrosaturnino It should work as expected because if user interacts with tBTC contract, before SDK is initialized, it shouuld still wait for the initializaiton before proceesing with the code further. I'm not sure how this pattern is called, I call it lazy loading promise pattern. Let me explain:

1. Instead of storing sdk object in a property and add some helping variables (like `isSdkInitalized`) that will tell if the sdk is initialized or not, we now store the promise itself:

```
  /**
   * Holds the promise for the asynchronously initialized SDK instance.
   * This promise is set during the construction of the SDK class and is meant
   * to be resolved once the SDK is fully initialized. The promise is used to
   * ensure that the SDK is initialized only once and that all subsequent
   * retrievals of the SDK instance await this promise, thereby ensuring that
   * the SDK is ready for use before any operations are performed.
   *
   * The promise also facilitates updating the SDK instance dynamically: if a
   * new initialization is triggered (e.g., when a user logs in and provides a
   * new signer), it ensures that the new SDK object replaces any previously
   * pending SDK object, even if the original promise has not yet resolved.
   *
   * To retrieve the SDK object, use the `getSdk()` method. This method will
   * either return the already resolved SDK object or wait until the promise
   * resolves before returning the SDK.
   */

  private _sdkPromise: Promise<SDK | undefined>
```

2. We retrieve the SDK object by using `_getSDK` method which will wait for the promise mentioned above to be resolved:

```
private _getSdk = async (): Promise<SDK> => {
    const sdk = await this._sdkPromise
    if (!sdk) throw new EmptySdkObjectError()

    return sdk
  }
```

So, there are few cases that can happen:

a) User interacts with tbtc contract but the SDK is not initialized yet:

- in such cases we use `_getSDK` method so the code will wait until `_sdkPromise` is resolved and then proceeds with the code

b) User interacts with tbtc contract and SDK is already initialied:

- this means that the promise is already resolved so we will retudn the sdk object immidietaly from `this_.sdkPromise` (no need to wait for it being initialized because it already is, and the promise is already resolved)

I hope this explains my implementation clearly. If you have any questions, feel free to ask. I'll add this to PR description.

To summarize - you don't have to add any flow here to improve the user experience (unless there are some bugs in my code that I've missed 😅 ). I've tested it out by adding additional delay in the initialization process (so I prologned the initialization to 15 seconds) and it worked ok.

evandrosaturnino: @michalsmiarowski, thanks for the thorough explanation! I appreciate the clarity on how the promise-based initialization ensures that interactions with the tBTC contract wait for the SDK to be fully ready before proceeding. There seems no need to add any additional flow based on your current implementation. I'll approve the PR as it stands. Great work!

Issue #768, opened 2024-07-16
WalletConnect is not connecting with Safe


evandrosaturnino:
It is not connecting WalletConnect with Safe
![image](https://github.com/user-attachments/assets/1b3013a0-d885-4e5b-b701-b7234989782b)

PR #565, opened 2023-07-07
[Backport] Update wallet connect to v2

michalsmiarowski:
Backport of: #560 

**Note:** There were some conflicts in `yarn.lock` file during cherry-picking. The conflicts appeared in those commits:
- 6239829ebff6e4341bd663bff66f209e07d08f1a (cherry-pick of 1743d87b91c4cb6cbb74cb3a82c528e54fb483e5),
- 506a6c9a30eebcd000b75d1444357fbc57ed0f63 (cherry-pick of 16a73a55fc92c21ce157803f1b174fb2d6b350dc),
- 401f7c37ba0896ff7f17587efc2433d21c67a29b (cherry-pick of 5e39bb608f788d5cb38510f7786c3476f7d336e1),
- 73249079c647b5acbe3b7e9a9972352256c536e4 (cherry-pick of fb0c41f6d5a979b3883489959b27167979c5bf85).

Instead of resolving them manually I've decided to accept the original `yarn.lock` file from `main` branch and then just `yarn install` to properly install the libraries from the changed `package.json`. This was done for each commit separately.

Between releasing the update and doing a backport the new version of `@walletconnect/ethereum-provider` was released. I've decided to use it and modify the code a little bit. You can see the details in 68f0538dcfa301fe777a7a222a383b9717d727b7.

<p align="center">
&#x1f537; WalletConnect V2 implementation &#x1f537;
</p>

Hero Horizontal

This pull request updates our current Wallet Connect implementation to version 2, incorporating the latest features and improvements. It ensures compatibility with the latest standards and provides a better user experience for our wallet integration.

### Changes Made
- Created a new connector - `WalletConnectConnector` - that is compatible with `web3-react` version 6 which we use
- Added new libraries - `@walletconnect/ethereum-provider` and `@walletconnect/modal` - which helped us implement the new version
- Removed everything related to WalletConnect v1 (this included `@web3-react/walletconnect-connector` lib)
- Added new .env variable `WALLET_CONNECT_PROJECT_ID` that will store our project id needed for WalletConnect v2

PR #769, opened 2024-07-19
Fix update handler validation

evandrosaturnino:
## Description

A bug was identified when connecting WalletConnect to Gnosis Safe via the wc pairing API: Shortly after a user connects their Gnosis Safe vault to the app, the connection was disconnected. 
Upon investigation, it was discovered that the connection between WalletConnect and Gnosis Safe was being interrupted due to the chainId format generated after connecting to Gnosis Safe. This triggered an update that disrupted the connection because of the way the update validation was structured in App.tsx.

Depending on the format in which the chainId is being passed by a 3rd party service, it may trigger an update. To handle these edge cases, this PR aims to add a more comprehensive validation method similarly to what has already been done previously in the commit [@3ca72b93ca72b9](https://github.com/threshold-network/token-dashboard/pull/565/commits/3ca72b9f9739ccda2bf365b31ab17f588f3b7f57).

- [X] - [fix update handler validation](https://github.com/threshold-network/token-dashboard/commit/5d0e4556a9c6f631891cc16fa757b025c7219a6b)

## Type of change

- [x] 🐛 Bug fix (Non-breaking Change: Fixes an issue)
- [ ]  🛠️ Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..)
- [ ]  💅 New Feature (Breaking/Non-breaking Change)

## Notice

- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/shapeshift/web/pulls) for the same update/change?

## Issue (if applicable)

closes https://github.com/threshold-network/token-dashboard/issues/768

## Pull Request Type

- [x] New Feature (Breaking/Non-breaking Change)


somaweb3:
LGTM

src/web3/utils/address.ts
```diff
@@ -17,6 +17,15 @@ export const isSameETHAddress = (
...
 
+export const isSameChainId = (
+  chainId1: number | string,
+  chainId2: number
+): boolean => {
+  const chainId1Str = chainId1.toString()
+  const chainId2Str = chainId2.toString()
+  return chainId1Str === chainId2Str || chainId1Str === `0x${chainId2Str}`
+}
+
```
michalsmiarowski, line 28: This probably will work for mainnet but doesn't work for Sepolia. The id returned from gnosis is in hex value so it's not enough to prefix it wit h`0x` (although for mainnet it's probably enough because 1 as hex is 0x1). For Sepolia, the returned value from Gnosis is `0xaa36a7`, and the int value is `11155111`. You can also check those values in https://chainlist.org/?chain=11155111&testnets=true&search=sepolia for example. 

Besides the function name nad it's variable names are counter intuitive IMO. Here we assume that `chainId1` will be the value in hex while `chainId2` is just normal value, without `0x` prefix (that's why we add it in the end). I believe the names should indicate which is which, but tbh it should not matter, and it should properly check if it's the same chain id without the need to worry about shich should i pass first. 

I think that we should first detect if any of this numbers is a hex by checking if it's starts with `0x` (there might be better ways to check it so please check it out).

If not, then we just compare those two as string.

If both of them are, then we can also compare them as strings.

But if only one of them is hex, then we should convert the other one to hex first. `ethers` provides a function for that:

```
ethers.utils.hexlify(Number(value))
 ```

michalsmiarowski (changes requested):

src/web3/utils/address.ts
evandrosaturnino: @michalsmiarowski After further investigation, I found that the `hexlify` method in the `@ethersproject/bytes` library already performs acomprehensive validation on the value type passed as a parameter. It accept and check strings, numbers, and hex values (returning them as they were originally provided). The method also converts the input to a lowercase hex format. Given that chainId can be in string, number, or hex format, I believe the best way to ensure their equivalence is by passing them through the hexlify function: [refactor chain ID validation commit](https://github.com/threshold-network/token-dashboard/pull/769/commits/c2acee58f8b72d1b5f0299581cbf870aad8d4b51).

You can find the implementation of the hexlify function in Ethers v5.4.7 [here](https://github.com/ethers-io/ethers.js/blob/32a6b2a362815eb85ce3f3abad5adf92f2b80e10/packages/bytes/src.ts/index.ts#L193).

Please let me know if you think additional redundancy checks are necessary in this function.

michalsmiarowski: Looking good now 🔥

michalsmiarowski (approved):
LGTM 🔥

PR #815, opened 2025-03-14
Support Extra Data in Deposit Verification Endpoint Routing

lrsaturnino:
## Description
This PR updates the deposit address verification logic in the token dashboard to support deposits that include an `extraData` field (as currently used in Base or Arbitrum direct minting). 
Previously, the verification function always called the standard `/json` endpoint, which doesn’t account for `extraData`. 
With this change, the function conditionally calls the `/json-extradata` endpoint when `extraData` is provided, ensuring that the computed deposit address is correctly verified against the external service.

## Type of change

- [ ]  💅 New Feature (Non-breaking Change)
- [x] 🐛 Bug fix (Non-breaking Change)
- [ ]  🛠️ Chore (Non-breaking Change: Doc updates, pkg upgrades, typos, etc..

## Background
Our monitoring system (via Sentry telemetry) detected mismatches between the deposit address computed by the tBTC SDK and the address returned by our verification API. These discrepancies occur for deposits that include an `extraData` field, since the standard verification endpoint does not factor `extraData` into its address calculation. 

This PR addresses the issue by routing deposits with extraData to the dedicated `/json-extradata` endpoint.

## Changes Introduced
In `src/utils/verifyDepositAddress.ts`, the function now checks if the deposit includes a non‑empty `extraData` field.
If `extraData` exists, the URL is built using the `/json-extradata` endpoint; otherwise, the standard `/json` endpoint is used.
The rest of the verification process (HTTP GET, address comparison, error handling) remains unchanged.

## Impact
For deposits with `extraData`: The correct address is verified using the extraData-aware endpoint, which should prevent false Sentry alerts due to address mismatches.
For standard deposits: The verification process remains unchanged and backward compatible.

## Testing
Verified locally using deposits with and without the `extraData` field.
Confirmed that the appropriate endpoint is called and that the verification outcome (valid/invalid/error) is correctly attached to the telemetry data.

### DIFF
--- a/src/utils/verifyDepositAddress.ts
+++ b/src/utils/verifyDepositAddress.ts
@@ -12,27 +12,33 @@ export const verifyDepositAddress = async (
   depositAddress: string,
   network: BitcoinNetwork
 ): Promise<VerificationOutcome> => {
-  const endpoint = `https://us-central1-keep-prd-210b.cloudfunctions.net/verify-deposit-address`
-
+  const endpointURL = `https://us-central1-keep-prd-210b.cloudfunctions.net/verify-deposit-address`
   const { depositor, blindingFactor, refundPublicKeyHash, refundLocktime } =
     deposit
 
+  const hasExtraData = deposit.extraData !== undefined
+
+  const url = hasExtraData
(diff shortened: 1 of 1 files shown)


evandrosaturnino (approved):
LGTM

PR #842, opened 2025-06-24
chore: add cookie3 script

piotr-roslaniec:


evandrosaturnino (commented):
@piotr-roslaniec The workflow that checks Prettier failed. I believe running `yarn build` and `yarn format:fix `should fix this.

For the token-dashboard, we usually merge to the releases branch (releases/mainnet/v2.3.x) before backporting to main. I sent you the release flow we have for the token-dashboard privately.

PR #848, opened 2025-07-06
fix: update starknet chain name in connection config

evandrosaturnino:
Update starknet chain name in non-evm connection hook


lrsaturnino (approved):
LGTM

